PR #3, opened 2025-11-24
[Jazzy] Fixing the free joint conversion to include mass and inertial properties on root body

saikishor:
forwardport the changes of https://github.com/NASA-JSC-Robotics/mujoco_ros2_simulation/pull/26

Similar to https://github.com/NASA-JSC-Robotics/mujoco_ros2_simulation/pull/41


eholum-nasa (approved):
This is a non-functional script on jazzy, but good to port to keep the logic similar to what it was if we ever need to refer to how we converted URDFs to MJCFs...

PR #8, opened 2025-11-24
[jazzy] Added capability to provide a key frame as starting configuration.

eholum-nasa:
Ports https://github.com/NASA-JSC-Robotics/mujoco_ros2_simulation/pull/34 from humble to jazzy.

Enables persisting the sim state to disk and restarting from that saved state.


eholum-nasa:
Whoops, go overzealous porting branches. https://github.com/ros-controls/mujoco_ros2_simulation/pull/4.

PR #10, opened 2025-11-24
Add taskbar icon

ndunkelb-nasa:
Adding an icon to the taskbar so that it isn't the generic gear icon. I used the ros-controls icon, but could alternatively be Mujoco's M? I am open to suggestions. I am also definitely not well versed in glfw-things so just did the best way I could find quickly. I am using lodepng to load the image, which is already a mujoco dependency.


saikishor (commented):
@ndunkelb-nasa Maybe better [M of MuJoCo](view-source:https://mujoco.readthedocs.io/en/stable/_static/favicons/favicon-192x192.png) instead of ros2_control logo

In the future, we can combine both logos and create a unique one.

What do you think?

image

PR #14, opened 2025-12-01
Ensure the FetchContent mujoco library is available during runtime

### LINKED ISSUE #13: Plugin not loading correctly in ROS 2 Kilted
**Describe the bug**
I was able to build this package in ROS 2 Kilted but get a runtime error / failure to load the mujoco library at runtime.

**To Reproduce**
1. Clone this repo
2. After `rosdep install`, run `colcon build` -- builds fine
3. Run the example: `ros2 launch mujoco_ros2_simulation test_robot.launch.py`

```
[ros2_control_node-2] [ERROR] [1764428272.472141404] [controller_manager]: 
Caught exception of type : N9pluginlib20LibraryLoadExceptionE while loading hardware: 
Failed to load library /home/sebastian/workspace/mujoco_ros_ws/install/mujoco_ros2_simulation/lib/mujoco_ros2_simulation/libmujoco_ros2_simulation.so. 
Make sure that you are calling the PLUGINLIB_EXPORT_CLASS macro in the library code, and that names are consistent between this macro and your XML. 
Error string: Could not load library dlopen error: libmujoco.so.3.3.4: cannot open shared object file: 
No such file or directory, at ./src/shared_library.c:96
```

**Expected behavior**
Should work

**Screenshots**

 - OS: Ubuntu 24.04
 - Version: ROS 2 Kilted

I can confirm that I have this file auto-fetched and installed:
```
mujoco_ros2_simulation/mujoco/mujoco-3.3.4/lib/libmujoco.so.3.3.4
```

And I can `ldd` it just fine:

```
$ ldd mujoco/mujoco-3.3.4/lib/libmujoco.so.3.3.4
        linux-vdso.so.1 (0x000074f907256000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x000074f90724b000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x000074f906cf1000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000074f906a00000)
        /lib64/ld-linux-x86-64.so.2 (0x000074f907258000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x000074f907244000)
```

eholum-nasa:
Resolves #13.

The issue being that `set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)` won't work for the on-the-fly source download because there isn't anything in the install path that cmake can link to, so the linker just fails to find it at runtime. This change ensures the install is available and on the RPATH for the library.


CMakeLists.txt
```diff
@@ -147,6 +153,10 @@ add_library(mujoco_ros2_simulation SHARED
     src/mujoco_cameras.cpp
     src/mujoco_lidar.cpp
 )
+# Ensure MuJoCo library can be found at runtime regardless of how it was installed above
+set_target_properties(mujoco_ros2_simulation PROPERTIES
+  INSTALL_RPATH "\$ORIGIN/../lib;${CMAKE_INSTALL_PREFIX}/lib"
+)
```
saikishor, line 159: Does it make sense to have this when we use `mujoco_vendor`?

eholum-nasa: I don't think it will be necessary but I don't think it will hurt.

That said... maybe once the vendor package is available we should just remove the other paths? Everything else is pulled in from rosdep so maybe that should be too?

saikishor: Sure.

saikishor: Sure.

PR #17, opened 2025-12-03
Consolidate test URDF and launch files

eholum-nasa:
Mostly non-functional change to consolidate the test resources into a single URDF and launch file. Now you can launch one or the other VIA launch arg, as noted in the readme:

```bash
# Brings up the hardware drivers and mujoco interface, along with a single position controller
ros2 launch mujoco_ros2_simulation test_robot.launch.py

# Or optionally include the PID controller as mentioned above
ros2 launch mujoco_ros2_simulation test_robot.launch.py use_pid:=true
```

While I'm adding reformatting some things and adjusted the link masses to 1 kg.

I was going to consolidate the MJCFs but they were generated from our old tool, so won't poke at those just yet... ideally when we update the converter the difference should just be in which input file to include.

Am background working on some basic tests so wanted to use a single URDF for that.

### DIFF
--- a/test/launch/test_robot.launch.py
+++ b/test/launch/test_robot.launch.py
@@ -33,7 +33,9 @@ from launch_ros.substitutions import FindPackageShare
 def generate_launch_description():
 
     # Refer https://github.com/ros-controls/mujoco_ros2_simulation?tab=readme-ov-file#joints
-    use_pid = DeclareLaunchArgument("use_pid", default_value="false", description="If we should use PID control to enable other control modes")
+    use_pid = DeclareLaunchArgument(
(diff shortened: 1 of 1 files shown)


test/launch/test_robot.launch.py
```diff
@@ -29,6 +31,9 @@
 
 
 def generate_launch_description():
+
+    use_pid = DeclareLaunchArgument("use_pid", default_value="false", description="If we should use PID control")
```
saikishor:
```suggestion
    # Refer https://github.com/ros-controls/mujoco_ros2_simulation?tab=readme-ov-file#joints
    use_pid = DeclareLaunchArgument("use_pid", default_value="false", description="If we should use PID control to enable other control modes")
```

PR #18, opened 2025-12-04
[Feature] MJCF generation at runtime

Ortisa:
This PR allows to generate MJCF at runtime and to be able to publish that on a separate topic.

The following features are added into this PR:
 * Generate MJCF and publish it over a topic using argument `--publish_topic`
 * Added `--scene` arg to be able to switch between scene at runtime
 * Support for defining input tags (`raw_inputs` and `processed_inputs`) directly in the URDF
 * Added `-a` / `--asset_dir` argument to specify the directory to reuse already generated objects from a given asset folder, otherwise generate them.
 * Option to also choose whether to save the model with argument `-s`

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,8 @@ Could map to the following hardware interface:
   </tbody>
 </table>
 
+> [!NOTE]
+> The `torque` and `force` command/state interfaces are semantically equivalent to `effort`, and map to the same underlying data in the sim.
 
 Switching actuator/control types on the fly is an [open issue](#13).
 
@@ -222,6 +224,10 @@ In the corresponding ros2_control xacro, this becomes a single sensor:
     <param name="mujoco_type">fts</param>
     <!-- There is no requirement for the mujoco_sensor_name to match the ros2_control sensor name -->
     <param name="mujoco_sensor_name">fts_sensor</param>
+    <!-- Default value of force_mjcf_suffix is '_force' -->
+    <param name="force_mjcf_suffix">_force</param>
+    <!-- Default value of torque_mjcf_suffix is '_torque' -->
+    <param name="torque_mjcf_suffix">_torque</param>
     <state_interface name="force.x"/>
     <state_interface name="force.y"/>
     <state_interface name="force.z"/>
@@ -248,6 +254,12 @@ Which then map to the corresponding ros2_control sensor:
     <param name="mujoco_type">imu</param>
     <!-- There is no requirement for the mujoco_sensor_name to match the ros2_control sensor name -->
     <param name="mujoco_sensor_name">imu_sensor</param>
+    <!-- Default value of orientation_mjcf_suffix is '_quat' -->
+    <param name="orientation_mjcf_suffix">_quat</param>
+    <!-- Default value of angular_velocity_mjcf_suffix is '_gyro' -->
+    <param name="angular_velocity_mjcf_suffix">_gyro</param>
+    <!-- Default value of linear_acceleration_mjcf_suffix is '_accel' -->
+    <param name="linear_acceleration_mjcf_suffix">_accel</param>
     <state_interface name="orientation.x"/>
     <state_interface name="orientation.y"/>
     <state_interface name="orientation.z"/>
--- a/include/mujoco_ros2_simulation/mujoco_system_interface.hpp
+++ b/include/mujoco_ros2_simulation/mujoco_system_interface.hpp
@@ -33,6 +33,7 @@
 #include <rclcpp/rclcpp.hpp>
 #include <rclcpp_lifecycle/node_interfaces/lifecycle_node_interface.hpp>
 #include <rclcpp_lifecycle/state.hpp>
+#include <realtime_tools/realtime_publisher.hpp>
 #include <rosgraph_msgs/msg/clock.hpp>
 
 #include <mujoco/mujoco.h>
@@ -185,6 +186,8 @@ private:
    */
   void publish_clock();
 
+  rclcpp::Logger get_logger() const;
+
   // System information
   std::string model_path_;
 
@@ -199,6 +202,9 @@ private:
   mjvCamera cam_;
   mjvOption opt_;
   mjvPerturb pert_;
+
+  // Logger
+  rclcpp::Logger logger_ = rclcpp::get_logger("MujocoSystemInterface");
 
   // Speed scaling parameter. if set to >0 then we ignore the value set in the simulate app and instead
   // attempt to loop at whatever this is set to. If this is <0, then we use the value from the app.
@@ -218,6 +224,7 @@ private:
 
   // Primary clock publisher for the world
   std::shared_ptr<rclcpp::Publisher<rosgraph_msgs::msg::Clock>> clock_publisher_;
+  realtime_tools::RealtimePublisher<rosgraph_msgs::msg::Clock>::SharedPtr clock_realtime_publisher_;
 
   // Containers for RGB-D cameras
   std::unique_ptr<MujocoCameras> cameras_;
--- a/src/mujoco_system_interface.cpp
+++ b/src/mujoco_system_interface.cpp
@@ -362,7 +362,7 @@ mjModel* loadModelFromTopic(rclcpp::Node::SharedPtr node)
 
   rclcpp::QoS qos_profile(rclcpp::QoSInitialization::from_rmw(rmw_qos_profile_default));
   qos_profile.reliable().transient_local().keep_last(1);
-  RCLCPP_INFO(rclcpp::get_logger("MujocoSystemInterface"), "Trying to get the mujoco model from topic");
+  RCLCPP_INFO(node->get_logger(), "Trying to get the mujoco model from topic");
 
   // Try to get mujoco_model via topic
   auto mujoco_model_sub = node->create_subscription<std_msgs::msg::String>(
@@ -534,25 +534,24 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
   const auto model_path_maybe = get_parameter("mujoco_model");
   if (!model_path_maybe.has_value())
   {
-    RCLCPP_INFO(rclcpp::get_logger("MujocoSystemInterface"), "Parameter 'mujoco_model' not found in URDF.");
+    RCLCPP_INFO(get_logger(), "Parameter 'mujoco_model' not found in URDF.");
     model_path_.clear();
   }
   else
   {
     model_path_ = model_path_maybe.value();
-    RCLCPP_INFO_STREAM(rclcpp::get_logger("MujocoSystemInterface"), "Loading 'mujoco_model' from: " << model_path_);
+    RCLCPP_INFO_STREAM(get_logger(), "Loading 'mujoco_model' from: " << model_path_);
   }
 
   // Pull the initial speed factor from the hardware parameters, if present
   sim_speed_factor_ = std::stod(get_parameter("sim_speed_factor").value_or("-1"));
   if (sim_speed_factor_ > 0)
   {
-    RCLCPP_INFO_STREAM(rclcpp::get_logger("MujocoSystemInterface"),
-                       "Running the simulation at " << sim_speed_factor_ * 100.0 << " percent speed");
+    RCLCPP_INFO_STREAM(get_logger(), "Running the simulation at " << sim_speed_factor_ * 100.0 << " percent speed");
   }
   else
   {
-    RCLCPP_INFO_STREAM(rclcpp::get_logger("MujocoSystemInterface"), "No sim_speed set, using the setting from the UI");
+    RCLCPP_INFO_STREAM(get_logger(), "No sim_speed set, using the setting from the UI");
   }
 
   // Pull the camera publish rate out of the info, if present, otherwise default to 5 hz.
@@ -562,7 +561,7 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
 
   // Check for headless mode
   bool headless = hardware_interface::parse_bool(get_parameter("headless").value_or("false"));
-  RCLCPP_INFO_EXPRESSION(rclcpp::get_logger("MujocoSystemInterface"), headless, "Running in HEADLESS mode.");
+  RCLCPP_INFO_EXPRESSION(get_logger(), headless, "Running in HEADLESS mode.");
 
   // We essentially reconstruct the 'simulate.cc::main()' function here, and
   // launch a Simulate object with all necessary rendering process/options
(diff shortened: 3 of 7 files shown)


scripts/make_mjcf_from_robot_description.py
eholum-nasa, line 1: I can't believe any one else is touching this file!

Out of curiosity what OS version are you using? I have issues because `bpy` isn't available on arm64 Ubuntu, and unavailable on 24.04 (the last time I checked). Does reorienting obj axes work for you guys?

scripts/make_mjcf_from_robot_description.py
```diff
@@ -107,14 +116,53 @@ def resolve_color(visual):
...
+                    if os.path.exists(mesh_file) and os.path.exists(settings_file):
+                        try:
+                            with open(settings_file) as f:
+                                data = json.load(f)
+                                used_threshold = float(data.get(f"{stem}"))
+                        except (FileNotFoundError, PermissionError, json.JSONDecodeError) as e:
+                            print(f"Warning: could not read thresholds for {stem}: {e}")
+                            used_threshold = None
+                        # Use existing decomposed object only if it has the same threshold, otherwise regenerate it.
+                        if used_threshold is not None and math.isclose(
```
eholum-nasa, line 139: Ah clever. Have you guys found this to be pretty reliable in preventing redoing decomposition? If so that's great!

test/launch/test_robot.launch.py
```diff
@@ -56,6 +57,25 @@ def generate_launch_description():
 
     robot_description = {"robot_description": ParameterValue(value=robot_description_content, value_type=str)}
 
+    converter_command = [
```
eholum-nasa, line 65: Once this works you can remove the arg from the ros2_control tags in the URDF and it'll be magic!

scripts/make_mjcf_from_robot_description.py
```diff
@@ -1142,66 +1361,169 @@ def add_urdf_free_joint(urdf):
...
         help="Adds a free joint before the root link of the robot in the urdf before conversion",
     )
+    parser.add_argument(
+        "-a",
+        "--asset_dir",
+        required=False,
+        default=None,
+        help="Optionally pass an existing folder with pre-generated OBJ meshes.",
+    )
+    parser.add_argument("--scene", required=False, default=None, help="Optionally pass an existing xml for the scene")
```
eholum-nasa, line 1439: So is this essentially just `scene.xml`? Which doesn't necessarily _have_ to come with the pre-generated meshes?

scripts/make_mjcf_from_robot_description.py
```diff
@@ -629,27 +704,99 @@ def parse_inputs_xml(filename=None):
...
+    else:
+        raise ValueError(
+            f"Root tag in file must be either 'mujoco_inputs' (standalone XML) or 'robot' (URDF), not '{root.tagName}'"
+        )
 
     return raw_inputs, processed_inputs
 
 
-def fix_free_joint(dom, urdf, joint_name="floating_base_joint"):
+def parse_scene_xml(filename=None):
```
eholum-nasa, line 750: Ah, was a little confused from the argparse param but I think I see what you're going for. Might be worth adding something similar to the `mujoco_inputs` comments just so we know what files would look like here?

eholum-nasa (commented):
I am impressed that you guys were able to follow that tool! It is a mess, but it's a semi-functional mess.

I took a quick skim through to see the changes and I think it's very promising. I need to actually get a machine that I'm able to test this with the jazzy version on... but hopefully that can happen soon. This [old issue](https://github.com/NASA-JSC-Robotics/mujoco_ros2_simulation/issues/22) is a pain.

Just a few quick questions...

scripts/make_mjcf_from_robot_description.py
saikishor: Last night i realised the same thing when testing things on rolling. We are internally using Humble, that's why we could get it to work. 

One ideal change that I could think of is if the obj files are created and passed through the --assets path, then you can reuse them instead of regenerating them. 

I agree we need to find an alternative with bpy, if not it is going to be hard.

scripts/make_mjcf_from_robot_description.py
saikishor: Yess, we save a metadata config in the folders and we only regenerate them if the threshold in the URDF and the folder doesn't match. This could be the only way to avoid it. 

Similarly, in future we could do the same with the STL files and the .obj files using the checksum

test/launch/test_robot.launch.py
saikishor: Yuppp! That's the final plan 😍
So you can generate the MuJoCo model with different scenes and use them on the fly

scripts/make_mjcf_from_robot_description.py
saikishor: Yes, I thought this would be useful to be able to spawn the robot on different scenes like just an empty world or a world with a certain setup to manipulate objects or a uneven walking terrain etc

scripts/make_mjcf_from_robot_description.py
saikishor: You can also add things into the URDF and also as a scene arg

scripts/make_mjcf_from_robot_description.py
saikishor: @eholum-nasa I already have an alternative implementation using the `trimesh` library. I hope there are not much differences. Once this PR is merged, I can test it against my changes and then if things are working I'll open a PR

scripts/make_mjcf_from_robot_description.py
```diff
@@ -1083,6 +1255,52 @@ def get_parent_chain(link, transform=PyKDL.Frame.Identity()):
...
+    try:
+        rclpy.spin(mjcf_node)
+    except KeyboardInterrupt:
+        pass
+    finally:
+        mjcf_node.destroy_node()
+        try:
+            rclpy.shutdown()
+        except Exception:
+            pass
```
eholum-nasa: Could just use `rclpy.try_shutdown()` I think?

scripts/make_mjcf_from_robot_description.py
```diff
@@ -1142,66 +1361,169 @@ def add_urdf_free_joint(urdf):
...
+        urdf_path = tmp.name
 
     request_add_free_joint = parsed_args.add_free_joint
 
     convert_stl_to_obj = parsed_args.convert_stl_to_obj
 
-    # Part inputs data
-    raw_inputs, processed_inputs = parse_inputs_xml(parsed_args.mujoco_inputs)
+    output_filepath = parsed_args.output
+    if not os.path.isabs(parsed_args.output) and parsed_args.publish_topic:
```
eholum-nasa, line 1469: I'm not sure I understand this conditional? Why the check on if the topic is published?

scripts/make_mjcf_from_robot_description.py
```diff
@@ -1142,66 +1361,169 @@ def add_urdf_free_joint(urdf):
...
+    if parsed_args.save_only:
+        # Grab the output directory and ensure it ends with '/'
+        output_filepath = os.path.join(output_filepath, "")
+    elif parsed_args.publish_topic:
+        temp_dir = tempfile.TemporaryDirectory()
+        output_filepath = os.path.join(temp_dir.name, "")
+    else:
+        raise ValueError(
+            "You must specify at least one of the following options: "
+            "--publish_topic or --save-only with the path of the folder."
```
eholum-nasa: This is slightly unclear... and it might be better to make `--save_only` default to `True`? That way it "works" with only the urdf specified as input.

```suggestion
            "--publish_topic or --save_only."
```

scripts/make_mjcf_from_robot_description.py
```diff
@@ -1142,66 +1361,169 @@ def add_urdf_free_joint(urdf):
...
+
     decompose_dict, cameras_dict, modify_element_dict, lidar_dict = get_processed_mujoco_inputs(processed_inputs)
 
-    # Grab the output directory and ensure it ends with '/'
-    output_filepath = os.path.join(parsed_args.output, "")
+    if parsed_args.asset_dir:
+        assets_filepath = parsed_args.asset_dir
+        if not os.path.isabs(parsed_args.asset_dir):
+            assets_filepath = os.path.join(os.getcwd(), parsed_args.asset_dir)
+        if output_filepath + "assets" == assets_filepath:
```
eholum-nasa: I think?

```suggestion
        if output_filepath + "assets" in assets_filepath:
```

scripts/make_mjcf_from_robot_description.py
saikishor: I'll be frank, we saw that there was a minor bug on the assets directory if you use both `--publish-topic` and `--output `tags and this is just a way to fix it. We didn't spend too much time there. I can also make them mutually exclusive if needed

scripts/make_mjcf_from_robot_description.py
saikishor: Done

saikishor:
> FWIW I merged the trimesh changes into this branch just to check it out in jazzy any it worked! There are still functional things to worry about, like mesh colorings that `bpy` would just handle and `trimesh` doesn't seem to support. So a lot of grey in our converted MJCFs... But the changes in this particular PR seem very nice.

@eholum-nasa I don't see the commits on the branch

scripts/make_mjcf_from_robot_description.py
ndunkelb-nasa: Given Erik's comments, do you have an idea of how to pull in that scene if it has meshes as well? I think this way would work for primitive objects, but not custom meshes

saikishor: Sure. I think it can be done. It makes sense it handles meshes too. For instance, it can be a table and an object scene to start grasping

saikishor: That's the main idea of having this argument. It'll be similar to gazebo worlds

ndunkelb-nasa: Yeah, I guess that is fair, if everything is defined as a part of the URDF, you get it for free. Although now that I am thinking about these things - would it make some sense to add in a "scene" as a urdf instead of an xml? At least on our side, I think we are more comfortable handling things in URDF world instead of mjcfs. I guess you could just handle everything in one, but I'm not positive how it would handle "free" joint type things for mobile robots. I'm not sure I have a suggestion here, just curious what your thoughts are.

saikishor: @ndunkelb-nasa Sure. To be frank with you, we added some of the parameters envisioning what we could do in future. We currently don't have a world to test and see which is better. 

I would propose to merge this as it is and then we can do a second iteration on this part. 

What do you think?

saikishor:
@ndunkelb-nasa Apart from that, I would like to hear from you what you think about the features in this PR?. Do you think it is interesting?

ndunkelb-nasa:
> @ndunkelb-nasa this is great. Do you want to merge it into this PR? or shall we do it in a different PR?. Let me know. Depending on it, I can merge it right now

I think I am happy to keep it separate bc there is already a bunch of focused discussion in this one. But I think after this goes in, we should certainly go ahead and do a PR for that as well. I would like someone to check my work as well!

test/launch/test_robot.launch.py
```diff
@@ -61,6 +62,25 @@ def generate_launch_description():
...
```
ndunkelb-nasa, line 65: So what I **thought** I understood from the new code was that I could replace this with something like
```
    converter_process = Node(
        package="mujoco_ros2_simulation",
        executable="make_mjcf_from_robot_description.py",
        output="screen",
        arguments=[
            "--publish_topic",
            "mujoco_robot_description",
            "--scene",
            PathJoinSubstitution(
                [
                    FindPackageShare("mujoco_ros2_simulation"),
                    "test_resources",
                    "scene.xml",
                ]
            ),
        ],
    )
```
and I could remove the `mujoco_model` parameter from the urdf, and it should just work. However, when I do that, my ros2 control node crashes, but I do see the `mujoco_robot_description_topic` with the mujoco info.

What am I missing here?

saikishor: I believe yes, something like that should already get it to work. Please give me a couple of minutes to check

ndunkelb-nasa: I think im getting a failure on [this line](https://github.com/pal-robotics-forks/mujoco_ros2_simulation/blob/0de172d6e160cc999aa97143c452f0f99548eda5/src/mujoco_system_interface.cpp#L399)

saikishor: Unfortunately, I cannot test it rolling with bpy

```
[make_mjcf_from_robot_description-5] Traceback (most recent call last):
[make_mjcf_from_robot_description-5]   File "/home/user/ros2_control_ws/install/mujoco_ros2_simulation/share/mujoco_ros2_simulation/scripts/make_mjcf_from_robot_description.py", line 21, in <module>
[make_mjcf_from_robot_description-5]     import bpy
[make_mjcf_from_robot_description-5] ModuleNotFoundError: No module named 'bpy'
[ERROR] [make_mjcf_from_robot_description-5]: process has died [pid 8272, exit code 1, cmd 'python3 /home/user/ros2_control_ws/install/mujoco_ros2_simulation/share/mujoco_ros2_simulation/scripts/make_mjcf_from_robot_description.py'].
```

Shall we merge your changes into this as well? that should make things easier

ndunkelb-nasa: ah maybe I just don't have the inputs there

saikishor: Yes, your URDF description needs to have the URDF inputs and all

ndunkelb-nasa: alright, I actually figured it out. It was missing the inputs **and** the typical scene like we have there needs to get rid of [this line](https://github.com/pal-robotics-forks/mujoco_ros2_simulation/blob/0de172d6e160cc999aa97143c452f0f99548eda5/resources/scene.xml#L2) so it doesn't try to reference a file that no longer exists

I still think this PR could benefit from a working example of this workflow, because I think this is actually what everyone should be using instead of generating separate mjcf files.

saikishor: Yes, for now the idea is not to have any includes in the scene file. Just to not to overcomplicate things

saikishor:
> I'm not sure I would have a good way at the moment. What would be normal for y'all? Discord?

We are connected via slack channel of [ros-controls](https://join.slack.com/t/roscontrol/shared_invite/zt-3je6bj56t-9SH5I9aT6qdA8Ph991UCbA) or I also use Discord with same username as github

eholum-nasa (approved):

PR #19, opened 2025-12-04
Set simulation window to max window size

saikishor:
This just sets the simulation window to the max operational size of the window


eholum-nasa (approved):
@saikishor I pulled in the latest main with the icon and headless config, I think this is good to merge but if you wanted to give it a quick sanity check, I'll defer to you to push the button!

Issue #20, opened 2025-12-05
Running the simulation in headless mode


pgesel-rai:
I am interested in running this hardware interface in headless mode, for remote environments, such as CI and evaluation suites. Without a window, I believe this interface will fail to start.  

**Describe the solution you'd like**
There are several options to achieve this. The most straightforward is to create a new code path that skips the `mj::Simulate` class entirely. This would require some refactoring to extract common functionality. Another potential option is to implement a `class HeadlessGlfwAdapter : public PlatformUIAdapter`, which mocks the `GlfwAdapter`. From some preliminary testing, I believe option one is better.

PR #21, opened 2025-12-05
Add support for headless mode

saikishor:
Fixes: #20 
@pgesel-rai FYI

In future, we can add some changes, like if we set the headless parameter from the launch file, we can toggle the behaviour

### DIFF
--- a/src/mujoco_system_interface.cpp
+++ b/src/mujoco_system_interface.cpp
@@ -68,6 +68,9 @@ namespace mujoco_ros2_simulation
 namespace mj = ::mujoco;
 namespace mju = ::mujoco::sample_util;
 
+/**
+ * No-op UI adapter to support running the drivers in a headless environment.
+ */
 class HeadlessAdapter : public mj::PlatformUIAdapter
 {
 public:
--- a/test/test_resources/test_robot.urdf
+++ b/test/test_resources/test_robot.urdf
@@ -148,6 +148,8 @@
 
       <!-- Override the slowdown settings in the Simulate App and run at 3x speed -->
       <param name="sim_speed_factor">3.0</param>
+
+      <!-- If set to true, will not launch the Simulate App window -->
       <param name="headless">$(arg headless)</param>
 
       <!-- Publish camera images at 6 hz -->


src/mujoco_system_interface.cpp
```diff
@@ -67,6 +68,105 @@ namespace mujoco_ros2_simulation
 namespace mj = ::mujoco;
 namespace mju = ::mujoco::sample_util;
 
+class HeadlessAdapter : public mj::PlatformUIAdapter
```
eholum-nasa, line 74:
```suggestion
namespace mju = ::mujoco::sample_util;

/**
 * No-op UI adapter to support running the drivers in a headless environment.
 */
class HeadlessAdapter : public mj::PlatformUIAdapter
```

test/test_resources/test_robot.urdf
```diff
@@ -147,6 +148,7 @@
 
       <!-- Override the slowdown settings in the Simulate App and run at 3x speed -->
       <param name="sim_speed_factor">3.0</param>
+      <param name="headless">$(arg headless)</param>
```
eholum-nasa, line 153:
```suggestion

      <!-- If set to true, will not launch the Simulate App window -->
      <param name="headless">$(arg headless)</param>
```

src/mujoco_system_interface.cpp
```diff
@@ -67,6 +68,108 @@ namespace mujoco_ros2_simulation
...
+    return 96.0;
+  }
+  std::pair<int, int> GetFramebufferSize() const override
+  {
+    return { 800, 600 };
+  }
+  std::pair<int, int> GetWindowSize() const override
+  {
+    return { 800, 600 };
+  }
```
pgesel-rai, line 95: It would be nice if these values were extracted from the camera resolutions instead of being hard coded. But I tried setting them to a value smaller than the rendering resolution and it seemed to have no effect. So it is probably okay.

saikishor: Yeah, it has no effect, but we can modify it in the future, if needed

I personally haven't tested things with camera sensors

eholum-nasa: Yeah I think these will just be thrown away? I don't know that anything will even be calling these functions. The cameras get their own [context](https://github.com/ros-controls/mujoco_ros2_simulation/blob/main/src/mujoco_cameras.cpp#L152). That said, what's happening under the hood is with GLFW is kind of a mystery to me so if there's something else going on let me know...

saikishor: Ok. I'll go ahead and merge it for now. If we see issues, we can open new issues

PR #26, opened 2025-12-09
Use realtime publisher for publishing the clock

saikishor:
This changed the clock publisher to the realtime_publisher


src/mujoco_system_interface.cpp
```diff
@@ -1788,7 +1793,7 @@ void MujocoSystemInterface::publish_clock()
 
   rosgraph_msgs::msg::Clock sim_time_msg;
   sim_time_msg.clock = sim_time_ros;
-  clock_publisher_->publish(sim_time_msg);
+  clock_realtime_publisher_->try_publish(sim_time_msg);
```
eholum-nasa, line 1771: Tangential question - since the physics loop is running in a spawned thread in the hardware interface, will it be get any kind of priority from the parent process? Is it worth adding it if a realtime kernel is available like in the ros2 control node?

saikishor: I believe yes. It is inherited.

eholum-nasa (approved):

PR #28, opened 2025-12-10
Reuse the logger with get_logger method

saikishor:
Just consolidates all get_logger calls with a single one that returns the member variable

### DIFF
--- a/include/mujoco_ros2_simulation/mujoco_system_interface.hpp
+++ b/include/mujoco_ros2_simulation/mujoco_system_interface.hpp
@@ -185,6 +185,8 @@ private:
    */
   void publish_clock();
 
+  rclcpp::Logger get_logger() const;
+
   // System information
   std::string model_path_;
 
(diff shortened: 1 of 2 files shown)


include/mujoco_ros2_simulation/mujoco_system_interface.hpp
```diff
@@ -200,6 +202,9 @@ class MujocoSystemInterface : public hardware_interface::SystemInterface
   mjvOption opt_;
   mjvPerturb pert_;
 
+  // Logger
+  rclcpp::Logger logger_ = rclcpp::get_logger("MujocoSystemInterface");
```
eholum-nasa, line 206: Oh I did not realize this got constructed every time! Nice change!

eholum-nasa (approved):
Good learnings for me here... though should we be using node logging in the system interface once the node is available? Doesn't seem like there's a significant difference.

include/mujoco_ros2_simulation/mujoco_system_interface.hpp
saikishor: Yup!

saikishor:
> Good learnings for me here... though should we be using node logging in the system interface once the node is available? Doesn't seem like there's a significant difference.

As it is a simulation, this is fine

PR #29, opened 2025-12-10
Support torque and force command and state interfaces

saikishor:
This PR add changes that are needed to support torque and force command and state interfaces. Basically, it remaps to the effort part of the interface.

### DIFF
--- a/test/test_resources/test_robot.urdf
+++ b/test/test_resources/test_robot.urdf
@@ -181,6 +181,8 @@
         <param name="initial_value">0.0</param>
       </state_interface>
       <state_interface name="velocity"/>
+
+       <!-- Effort and torque are semantically different but map to the same underlying data. -->
       <state_interface name="effort"/>
       <state_interface name="torque"/>
     </joint>


test/test_resources/test_robot.urdf
```diff
@@ -179,6 +180,7 @@
       </state_interface>
       <state_interface name="velocity"/>
       <state_interface name="effort"/>
+      <state_interface name="torque"/>
```
eholum-nasa, line 187: This is a little weird, right? Should we just remove the `effort` state interface to be more explicit since these are revolute joints?

eholum-nasa (commented):
Nice! Is the difference for say, prismatic vs revolute joints?

The only other comment I have, because of the question on the test robot, is that it's probably worth updating the [README](https://github.com/ros-controls/mujoco_ros2_simulation/blob/main/README.md?plain=1#L190) to show that effort/force/torque interfaces are supported, but that the differences are exclusively semantic... We could also just add a note?

```
**Note:** The `torque` and `force` command/state interfaces are semantically equivalent to `effort`, and map to the same underlying data in the sim.
```

test/test_resources/test_robot.urdf
saikishor: As you prefer, it is just to have it for reference.

eholum-nasa: Oh yeah I think that's fine. Maybe we just double down on emphasizing that the differences are only skin deep? I foresee it being a source of confusion for people that might have different expectation. Added suggested comments!

eholum-nasa, line 187:
```suggestion

       <!-- Effort and torque are semantically different but map to the same underlying data. -->
      <state_interface name="effort"/>
      <state_interface name="torque"/>
```

test/test_resources/test_robot.urdf
```diff
@@ -168,6 +168,7 @@
...
```
eholum-nasa, line 173:
```suggestion

      <!-- Effort and torque are semantically different but map to the same underlying data. -->
      <state_interface name="effort"/>
      <state_interface name="torque"/>
```

PR #32, opened 2025-12-15
Add tests to test the new usecases

saikishor:
Closes https://github.com/ros-controls/mujoco_ros2_simulation/issues/31

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,10 @@ In the corresponding ros2_control xacro, this becomes a single sensor:
     <param name="mujoco_type">fts</param>
     <!-- There is no requirement for the mujoco_sensor_name to match the ros2_control sensor name -->
     <param name="mujoco_sensor_name">fts_sensor</param>
+    <!-- Default value of force_mjcf_suffix is '_force' -->
+    <param name="force_mjcf_suffix">_force</param>
+    <!-- Default value of torque_mjcf_suffix is '_torque' -->
+    <param name="torque_mjcf_suffix">_torque</param>
     <state_interface name="force.x"/>
     <state_interface name="force.y"/>
     <state_interface name="force.z"/>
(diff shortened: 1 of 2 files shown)


test/src/robot_launch_test.py
```diff
@@ -0,0 +1,207 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import time
+import unittest
+
+from ament_index_python.packages import get_package_share_directory
+from controller_manager.test_utils import check_controllers_running, check_if_js_published, check_node_running
```
eholum-nasa, line 22: TIL these exist. Very handy!

test/src/robot_launch_test.py
```diff
@@ -0,0 +1,207 @@
...
+
+        # Create a publisher to send commands to the position controller
+        pub = self.node.create_publisher(Float64MultiArray, "/position_controller/commands", 10)
+
+        # Wait for subscriber to connect
+        end_time = time.time() + 10
+        while time.time() < end_time:
+            if pub.get_subscription_count() > 0:
+                break
+            time.sleep(0.1)
```
eholum-nasa, line 107: I think check that it succeeded before continuing?

```suggestion
        end_time = time.time() + 10
        while True:
            if pub.get_subscription_count() > 0:
                break
            if time.time() > end_time:
                self.fail("Timed out waiting for subscribers")
            time.sleep(0.1)
```

eholum-nasa (approved):
Hero!

image

Thank you for these! I just have the minor nit, but the test would fail anyway so...

test/src/robot_launch_test.py
saikishor: They are maintained by ros-controls and reused by gz_ros2_control too, so it is better this way

test/src/robot_launch_test.py
saikishor: Better the original way right. Not a big fan of `while True`,

It is already checked in the next check right?

```
self.assertGreater(pub.get_subscription_count(), 0, "Controller did not subscribe to commands")
```

eholum-nasa: Oh whoops, yep! Sorry pre-coffee eyes... LGTM!

PR #35, opened 2025-12-18
Use trimesh instead of bpy

saikishor:


ndunkelb-nasa:
Hmm, on that last commit, the pixi test failed, I ran it again and it succeeded. I haven't used pixi at all, so not sure what that might be about

ndunkelb-nasa:
Sounds good. I have verified that it does correct conversions on two of our more complicated robots (just testing in simulate window so far) but just going through and making sure the full pipeline to ROS is working. Should be pretty soon.

PR #36, opened 2025-12-18
Rename package to `mujoco_ros2_control`

saikishor:
This renames the current package to mujoco_ros2_control from Jazzy on.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,7 +39,7 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           file: docker/Dockerfile
-          tags: mujoco_ros2_simulation:${{ github.head_ref || github.ref_name }}
+          tags: mujoco_ros2_control:${{ github.head_ref || github.ref_name }}
           target: mujoco_ros
           build-args: |
             ROS_DISTRO=${{ matrix.ros_distro }}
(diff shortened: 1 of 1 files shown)


docs/TOOLS.md
```diff
@@ -1,6 +1,6 @@
 # Mujoco To ROS Tools
 
-> **NOTE**: TODO: This tool is broken on jazzy and beyond, stick with the `humble-devel` branch until https://github.com/NASA-JSC-Robotics/mujoco_ros2_simulation/issues/22 is resolve.
+> **NOTE**: TODO: This tool is broken on jazzy and beyond, stick with the `humble-devel` branch until https://github.com/NASA-JSC-Robotics/mujoco_ros2_control/issues/22 is resolve.
```
saikishor:
```suggestion
```

Maybe this can be removed now?

eholum-nasa: Agreed!

mujoco_system_interface_plugin.xml
```diff
@@ -1,6 +1,6 @@
-<library path="mujoco_ros2_simulation">
-    <class name="mujoco_ros2_simulation/MujocoSystemInterface"
-        type="mujoco_ros2_simulation::MujocoSystemInterface"
+<library path="mujoco_ros2_control">
+    <class name="mujoco_ros2_control/MujocoSystemInterface"
+        type="mujoco_ros2_control::MujocoSystemInterface"
         base_class_type="hardware_interface::SystemInterface">
```
saikishor, line 4: @eholum-nasa  @ndunkelb-nasa  One more question, should we maintain a copy of the old plugin namings too? so, we don't break your stuff?

eholum-nasa: Appreciate the concern! I was thinking about it and no, better to make a clean cut. All of our stuff is submoduled (and still on humble) so if/when we migrate we'll just have to update this as well.

eholum-nasa (changes requested):
Ah! Container names that didn't get searched in my workspace. Great spot @yassine-cherni!

Issue #22, opened 2025-12-06
Upgrade MuJoCo to v3.4.0


yassine-cherni:
**Is your feature request related to a problem? Please describe.**
The current simulation backend is pinned to MuJoCo version **3.3.4**.
MuJoCo has recently released version [3.4.0,](https://github.com/google-deepmind/mujoco/releases/tag/3.4.0)
which introduces several parsing improvements (specifically for MJCF meshes and Euler angles)

**Describe the solution you'd like**
I would like to propose upgrading the MuJoCo dependency in the project to version **3.4.0**.
- Changelog: https://mujoco.readthedocs.io/en/3.4.0/changelog.html

saikishor:
I would love to go to the latest version. Let's trying to get in those PRs and then we can add some basic testing, later we can verify if this will build and work on the recent versions of MuJoCo without affecting anyone

PR #46, opened 2025-12-20
Bump MuJoCo library version to 3.4.0

saikishor:
This PR bumps the mujoco version to 3.4.0.

The idea is once the CI is good here, I will also bump the mujoco_vendor and make releases

Closes: https://github.com/ros-controls/mujoco_ros2_control/issues/22

### DIFF
--- a/src/mujoco_system_interface.cpp
+++ b/src/mujoco_system_interface.cpp
@@ -1721,8 +1721,15 @@ void MujocoSystemInterface::PhysicsLoop()
                 sim_->measured_slowdown = std::chrono::duration<double>(elapsedCPU).count() / elapsedSim;
                 measured = true;
               }
-              // inject noise
+// inject noise
+// Use mjVERSION_HEADER and if it is greater than 337 then do one thing or another
+// Needed due to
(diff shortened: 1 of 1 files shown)


src/mujoco_system_interface.cpp
```diff
@@ -1721,8 +1721,15 @@ void MujocoSystemInterface::PhysicsLoop()
                 sim_->measured_slowdown = std::chrono::duration<double>(elapsedCPU).count() / elapsedSim;
                 measured = true;
               }
-              // inject noise
+// inject noise
+// Use mjVERSION_HEADER and if it is greater than 337 then do one thing or another
+// Needed due to
+// https://github.com/google-deepmind/mujoco/commit/401bf431b8b0fe6e0a619412a607b5135dc4ded4#diff-3dc22ceeebd71304c41d349c6d273bda172ea88ff49c772dbdcf51b9b19bbd33R2943
+#if mjVERSION_HEADER < 337
```
eholum, line 1728: Forgive my ignorance here, should we just enforce that the header version match exactly what we expect in the CMakeLists.txt file? Would that make this conditional go away?

saikishor: Sure, but this allows people to go back to the older version and they don't have to touch the code to know what to fix. So, I would say it is better to have it handled also in the code

PR #49, opened 2025-12-21
Add pre-release workflow

christophfroehlich:
As in other ros-controls repositories

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -14,25 +14,25 @@ target_link_libraries(test_plugin
 find_package(launch_testing_ament_cmake REQUIRED)
 
 add_launch_test(src/robot_launch_test.py
-  TIMEOUT 70
+  TIMEOUT 50
 )
 
 add_launch_test(src/robot_launch_pid_test.py
-  TIMEOUT 70
+  TIMEOUT 50
 )
 
 add_launch_test(src/robot_launch_mjcf_generation_test.py
-  TIMEOUT 70
+  TIMEOUT 50
 )
 
 add_launch_test(src/robot_launch_mjcf_generation_from_robot_description_test.py
-  TIMEOUT 70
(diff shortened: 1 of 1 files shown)


saikishor (dismissed):
Looking good now!

test/src/robot_launch_test.py
```diff
@@ -43,6 +43,13 @@ def generate_test_description_common(use_pid="false", use_mjcf_from_topic="false
     os.environ["USE_MJCF_FROM_TOPIC"] = use_mjcf_from_topic
     os.environ["TEST_TRANSMISSIONS"] = test_transmissions
 
+    # Setup the venv needed for the make_mjcf_from_robot_description node
```
christophfroehlich: maybe this fits better into `setUpClass`?

saikishor: I added it here as all the tests call this common one. If not, we might have to add this to every setUpClass? Isn't it the case?

christophfroehlich: yes. as you prefer..

saikishor:
@christophfroehlich The tests are running properly, but not sure why exactly they fail with exit code -6, any idea?
```
2026-01-13T10:14:38.1969603Z 2: NO TESTS RAN
2026-01-13T10:14:38.1969951Z 2: terminate called without an active exception
2026-01-13T10:14:38.1970381Z 2: -- run_test.py: return code -6
2026-01-13T10:14:38.1973310Z 2: -- run_test.py: verify result file '/tmp/ws/test_results/mujoco_ros2_control/test_src_robot_launch_test.py.xunit.xml'
```

saikishor:
> no idea, any process ends with SIGABRT. pre-release runs in a different environment. Have you tested it locally in a native rolling environment?

I tested it on a rolling docker. I can check more this evening

PR #52, opened 2025-12-26
Setup virtualenv to handle Pip dependencies of make_mjcf_from_robot_description.py

saikishor:
This should handle the dependencies and allows to release the package. With the pip dependencies, we cannot release this package in buildfarm, so I came up with this solution. 

This PR should fix #49 and get it to work

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -24,5 +24,5 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-      - uses: actions/setup-python@v5
+      - uses: actions/setup-python@v6
       - uses: pre-commit/action@v3.0.1


eholum-nasa (commented):
> With the pip dependencies, we cannot release this package in buildfarm

I'm definitely ignorant here since I don't think I've ever run into this... but can you elaborate a little here? We basically need those tests to be runnable outside of the managed docker and/or pixi environment?

saikishor:
> I'm definitely ignorant here since I don't think I've ever run into this... but can you elaborate a little here? We basically need those tests to be runnable outside of the managed docker and/or pixi environment?

Ofcourse, we do. The thing is if we want to package things into debian and release them. We cannot do that for the python packages that are available via Pip and the buildfarm doesn't accept that.  Ofcourse, the source builds will work but you will need to install them by accepting to break system python packages.

So, what I was proposing here is to have a venv for us and then the packages use it from there. On the first run, it creates this venv and reutilizes it later on.

What do you think?

saikishor:
> Should this be mentioned somewhere here?
> https://github.com/ros-controls/mujoco_ros2_control/blob/main/docs/TOOLS.md#usage


Sure, I can add it. Moreover, anyone who has it system wide installed can directly run the python script without any issue. So, it serves both cases.

christophfroehlich (approved):
LGTM.

Note: If we split the tests in a packages, which we don't release, then there would be no need to wrap this in a bash script just for releasing it: We could remove the rosdep keys and it's the users responsibility to install it manually. But for sure, this is a convenient wrapper for first-time users of this package.

Issue #53, opened 2025-12-29
Mujoco model parameter susceptible to leading and trailing whitespaces in XML


sea-bass:
**Describe the bug**
I just spent a while debugging a "file not found" error because of leading/trailing whitespaces loaded from the XML file.

**To Reproduce**
I had a linter change this piece of code:

```xml
          <plugin>mujoco_ros2_control/MujocoSystemInterface</plugin>
          <param name="mujoco_model">$(find so_arm100_description)/mjcf/scene.xml</param>
```

to

```xml
          <plugin>mujoco_ros2_control/MujocoSystemInterface</plugin>
          <param name="mujoco_model">
              $(find so_arm100_description)/mjcf/scene.xml
          </param>
```

This caused the whitespaces before and after to register as part of the file path, and therefore my file was not found.

**Expected behavior**
Leading and trailing whitespaces should not break this loader.

**Environment (please complete the following information):**
 - OS: Ubuntu 24.04
 - Version: Kilted

PR #55, opened 2025-12-29
Fix trailing and leading whitespaces in the path + improve checks

saikishor:
Fixes: https://github.com/ros-controls/mujoco_ros2_control/issues/53

@sea-bass FYI


sea-bass (approved):
So quick, thank you!

Is the failing test unrelated/flaky?

PR #58, opened 2026-01-09
Main compiles on Humble

ndunkelb-nasa:
We have quite a few changes on main that we would like to have available on humble. Rather than backporting everything, this helps main run on humble. There aren't really any functional changes as far as I can tell, mostly just naming conventions. Maybe the biggest functional change is that it looks like there are different pid parameters available on Jazzy forward that will just be ignored in this scenario.

I'm not sure this was the best way to do it, but so far, I added a compile time definition of what ros distro we are on, then threw `#ifdef`s where necessary to change mostly just naming conventions. I am happy to change how this setup to something else if there is a more recommended way of doing so.

I have tested locally that this passes tests on Humble, Jazzy, Kilted, and Rolling, and I have run through a pretty comprehensive demo on one of our robots using this on Humble

### DIFF
--- a/src/mujoco_system_interface.cpp
+++ b/src/mujoco_system_interface.cpp
@@ -2209,6 +2209,18 @@ void MujocoSystemInterface::set_initial_pose()
   for (auto& actuator : mujoco_actuator_data_)
   {
     mj_data_->qpos[actuator.mj_pos_adr] = actuator.position_interface.state_;
+    if (actuator.is_position_control_enabled)
+    {
+      mj_data_->ctrl[actuator.mj_actuator_id] = actuator.position_interface.state_;
+    }
+    else if (actuator.is_velocity_control_enabled)
+    {
+      mj_data_->ctrl[actuator.mj_actuator_id] = actuator.velocity_interface.state_;
+    }
+    else if (actuator.is_effort_control_enabled)
+    {
+      mj_data_->ctrl[actuator.mj_actuator_id] = actuator.effort_interface.state_;
+    }
   }
 
   // Copy into the control data for reads


ndunkelb-nasa:
> @ndunkelb-nasa Can you review #56?. I think it is better to merge this after that one, as the conflicts to fix after this one will be complicated in that PR. Once that is merged, I can take the lead to finish rebasing this. Let me know what you think.

Yeah, I can take a look! I haven't used transmissions in ros2_control before, so I'll probably more be looking at how it integrates into the driver, but I wouldn't trust my thoughts on the transmission implementation.

saikishor:
> Yeah, I can take a look! I haven't used transmissions in ros2_control before, so I'll probably more be looking at how it integrates into the driver, but I wouldn't trust my thoughts on the transmission implementation.

Thank you. I understand. To give you an overview, if the transmission plugins are defined in the URDF, they are tried to be loaded and if not, it works as usual as before.

I refactored a bit of the code to separate the importation of the MuJoCo actuators and then importing ros2_control interfaces and then have a method to link the data between them

saikishor (dismissed):

saikishor:
@ndunkelb-nasa This is done. Can you verify and give me a confirmation, so we can  merge this too :)

saikishor (dismissed):

ndunkelb-nasa:
@saikishor, I put this in one of our repos to run to test, and realized the transmissions PR ended up messing up our tendon actuators. I didn't catch that those were being disabled in that. Were those disabled because you didn't think they were currently being used? Or because something in that change meant that tendons were unusable?

I would like to fix that before sending this through.

saikishor:
@ndunkelb-nasa Sorry about that. I didn't add them as I thought they were not being currently used and I've a usecase soon, so I thought of fixing it later or checking it by adding some tests

saikishor:
Can you try commenting this part and see if it works?

https://github.com/ros-controls/mujoco_ros2_control/blob/50c21b6679ad6385b41f5d0c5a363f58078312da/src/mujoco_system_interface.cpp#L1481-L1483

I believe it should.

ndunkelb-nasa:
> > Can you try commenting this part and see if it works?
> > https://github.com/ros-controls/mujoco_ros2_control/blob/50c21b6679ad6385b41f5d0c5a363f58078312da/src/mujoco_system_interface.cpp#L1481-L1483
> > 
> > I believe it should.
> 
> I had tried that, but I get an error about missing command interfaces. Im wondering if it doesn't have a good match to the bias types maybe? I am taking a look but have a meeting now and won't get to it for a bit.
> 
> ```
> [ros2_control_node-20] missing command interfaces:
> [ros2_control_node-20] ' left_finger_1_joint/position ' ' right_robotiq_85_left_knuckle_joint/position ```
> ```

Hmm on first check, that is not the case.

saikishor (approved):

PR #64, opened 2026-01-14
Added prismatic gripper to the test robot to validate tendons functionality

saikishor:
This PR fixes the current issue of tendons not working properly after merging the transmissions PR #56. Ideally, it is tested from now on.

### DIFF
--- a/src/mujoco_system_interface.cpp
+++ b/src/mujoco_system_interface.cpp
@@ -1518,7 +1518,10 @@ bool MujocoSystemInterface::register_mujoco_actuators()
       }
       else
       {
-        RCLCPP_ERROR(get_logger(), "Tendon actuator '%s' has no matching joint", act_name);
+        RCLCPP_ERROR(get_logger(),
+                     "Tendon actuator '%s' has no matching joint. Tendon actuators must be named the same as a joint "
+                     "that they will control.",
+                     act_name);
         return false;
       }
     }


src/mujoco_system_interface.cpp
```diff
@@ -1505,15 +1507,18 @@ bool MujocoSystemInterface::register_mujoco_actuators()
...
-        actuator_data.joint_name = std::string(tendon_name);
-        RCLCPP_INFO(get_logger(), "Registering MuJoCo actuator '%s' for tendon '%s'", act_name, tendon_name);
+        target_id = joint_id;
+        actuator_data.joint_name = std::string(act_name);
+        RCLCPP_INFO(get_logger(), "Registering MuJoCo tendon actuator '%s' using joint state", act_name);
       }
       else
       {
-        RCLCPP_ERROR(get_logger(), "Failed to find tendon name for actuator '%s'", act_name);
+        RCLCPP_ERROR(get_logger(), "Tendon actuator '%s' has no matching joint", act_name);
```
ndunkelb-nasa: Should this be updated to be something like this instead? At this point, I don't think there is any way a user would know of this requirement, and they would just be met with this error despite a technically correct mujoco model
```suggestion
        RCLCPP_ERROR(get_logger(), "Tendon actuator '%s' has no matching joint. Tendon actuators must be named the same as a joint that they will control", act_name);
```

src/mujoco_system_interface.cpp
```diff
@@ -1505,15 +1507,18 @@ bool MujocoSystemInterface::register_mujoco_actuators()
...
         return false;
```
ndunkelb-nasa, line 1525: Because this fails if we don't meet this criteria, I think this means that if you have a tendon actuator that we don't even care about in ROS, but it is not named the same as an joint name, this will still fail.

That doesn't sound like the right move? But maybe as long as we have it documented it is fine.

eholum-nasa: Maybe we just ignore it and continue with a warning?

ndunkelb-nasa: Yeah, I think that would be fine. I just tested out what happens if we do that, but the user messes up the tendon name as we had tested before. I think that there are plenty of errors to diagnose with?
```
[ros2_control_node-2] [ERROR] [1768405561.102307255] [MujocoSystemInterface]: Tendon actuator 'gripper_left_finger_join' has no matching joint
........
[ros2_control_node-2] [WARN] [1768405561.103061453] [MujocoSystemInterface]: Failed to find actuator for joint : gripper_left_finger_joint. This joint will be treated as a passive joint.
[ros2_control_node-2] [WARN] [1768405561.103069268] [MujocoSystemInterface]: Joint : gripper_left_finger_joint has command interfaces defined but no matching actuator in the MuJoCo model. This joint will be treated as a passive joint and no command interfaces will be exported.
........
[ros2_control_node-2] [INFO] [1768405561.121059154] [resource_manager]: Successful initialization of hardware 'MujocoSystem'
[ros2_control_node-2] [ERROR] [1768405561.121646876] [controller_manager]: The published robot description file (urdf) seems not to be genuine. The following error was caught:Wrong state or command interface configuration.
[ros2_control_node-2] missing state interfaces:
[ros2_control_node-2] 
[ros2_control_node-2] missing command interfaces:
[ros2_control_node-2] ' gripper_left_finger_joint/position '	
```
The sim technically starts up but is just frozen...

ndunkelb-nasa: Oh nvm, I didn't have the `continue` in there. It actually results in a seg fault... maybe not the best thing to do. Maybe we just leave it for now?

saikishor: Yes, I would leave it for now and get back to it with proper conventions and documentation.

PR #67, opened 2026-01-14
Handle passive joints properly and update the state information

saikishor:
This PR adds proper fixes inorder to have the passive joints properly integrated and readable from the mujoco part as well as ros2_control end

### DIFF
--- a/src/mujoco_system_interface.cpp
+++ b/src/mujoco_system_interface.cpp
@@ -1623,7 +1623,7 @@ bool MujocoSystemInterface::register_mujoco_actuators()
   if (override_mujoco_actuator_positions_)
   {
     RCLCPP_DEBUG(get_logger(),
-                 "Initializing acctuator position states from override start position file for %zu actuators.",
+                 "Initializing actuator position states from override start position file for %zu actuators.",
                  mujoco_actuator_data_.size());
 
     for (auto& actuator_data : mujoco_actuator_data_)
--- a/test/src/robot_launch_test.py
+++ b/test/src/robot_launch_test.py
@@ -240,7 +240,7 @@ class TestFixture(unittest.TestCase):
         )
         wait_for_topics.shutdown()
 
-        # Verify the same in the mujoco_actuaotrs_states topic
+        # Verify the same in the mujoco_actuators_states topic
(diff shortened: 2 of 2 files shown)


src/mujoco_system_interface.cpp
```diff
@@ -1599,6 +1595,52 @@ bool MujocoSystemInterface::register_mujoco_actuators()
...
+      passive_actuator.actuator_type = ActuatorType::PASSIVE;
+      mujoco_actuator_data_.push_back(passive_actuator);
+    }
+  }
+
+  // Set initial values if they are set in the info, or from override start position file
+  if (override_mujoco_actuator_positions_)
+  {
+    RCLCPP_DEBUG(get_logger(),
+                 "Initializing acctuator position states from override start position file for %zu actuators.",
```
eholum-nasa:
```suggestion
                 "Initializing actuator position states from override start position file for %zu actuators.",
```

test/config/controllers.yaml
```diff
@@ -9,7 +9,7 @@ controller_manager:
       type: position_controllers/JointGroupPositionController
 
     gripper_controller:
-      type: position_controllers/GripperActionController
+      type: position_controllers/JointGroupPositionController
```
saikishor, line 12: I modified this because this controller is available in a different package in the newest versiions of ros2_control. So, to have some consistency in tests, i modified it

test/src/robot_launch_test.py
```diff
@@ -181,6 +188,82 @@ def test_arm(self):
...
+        )
+        self.assertAlmostEqual(
+            msg.position[right_index],
+            0.04,
+            delta=0.005,
+            msg="gripper_right_finger_joint did not reach commanded position",
+        )
+        wait_for_topics.shutdown()
+
+        # Verify the same in the mujoco_actuaotrs_states topic
```
eholum-nasa:
```suggestion
        # Verify the same in the mujoco_actuators_states topic
```

eholum-nasa (dismissed):
Logic looks right to me and the tests pass! I just had the minor nits but otherwise seems GTG. Did not test explicitly...

Issue #68, opened 2026-01-15
Add support for contact sensor


Juliaj:
## Motivation

To create a locomotion controller from a learned policy (e.g., ONNX-based) for a bipedal robot often requires foot contact signals as part of the observation for gait control. See the Open Duck Playground MuJoCo example for the contact query: [Open_Duck_Playground mujoco_infer.py](https://github.com/apirrone/Open_Duck_Playground/blob/main/playground/open_duck_mini_v2/mujoco_infer.py#L96).

This is to add contact support for gait control and collision-oriented monitoring (unexpected contacts) using the MuJoCo contact results.

## Design (one primitive, two consumer modes)

Low-level primitive:
- Determine whether two bodies are in contact by iterating `mjData->contact[]` and mapping `contact.geom*` to body ids via `mjModel->geom_bodyid[]`.
- This is exposed as a raw boolean interface: `contact_raw`.

Consumer modes (how to produce the main `contact` signal from `contact_raw`):
- collision: `contact == contact_raw` (minimal filtering; default)
- gait: `contact` is a debounced/hysteresis version of `contact_raw` (stable for gait phase / state estimation)

## Configuration

Add to ros2_control URDF/xacro:

```xml
<sensor name="foot_floor_contact">
  <state_interface name="contact"/>      <!-- main contact output (depends on contact_consumer) -->
  <state_interface name="contact_raw"/>  <!-- raw/unfiltered contact output -->

  <param name="mujoco_type">contact</param>
  <param name="body1_name">foot</param>
  <param name="body2_name">world</param>  <!-- use "world" for static geometry in <worldbody> -->

  <!-- Optional: choose consumer behavior -->
  <!-- <param name="contact_consumer">collision</param> --> <!-- default -->
  <!-- <param name="contact_consumer">gait</param> -->

  <!-- Optional (steps, >=1): debounce/hysteresis for gait -->
  <!-- <param name="debounce_on_steps">2</param> -->
  <!-- <param name="debounce_off_steps">2</param> -->
</sensor>
```

State interfaces:
- `{sensor_name}/contact`: `1.0` or `0.0` (collision: equals raw; gait: debounced/hysteresis)
- `{sensor_name}/contact_raw`: `1.0` or `0.0` (always unfiltered)

## Notes (MuJoCo contact behavior)

- Requires `contype` / `conaffinity` configured on geoms for collision detection.
[...]

PR #69, opened 2026-01-15
support contact sensor

Juliaj:
Addresses 
- https://github.com/ros-controls/mujoco_ros2_control/issues/68


Juliaj:
Thanks, @saikishor, 

> I see more changes in this PR than the contact sensor integration itself.

The additional fix is for unit tests due to https://github.com/ros-controls/mujoco_ros2_control/issues/70

Juliaj:
> 3\. Do it entirely externally and wrap the HW interface with your own implementation. We exposed the [getData/getModel functions ](https://github.com/ros-controls/mujoco_ros2_control/pull/69/changes#diff-83b0fcafe0855f7b0540636774633f0da1e3688077f976e4a07d772fec56f86fR141), which would give external devs complete god-like access to the underlying sim if they wanted it.

Thanks @eholum-nasa!  By "wrap the HW interface with your own implementation," do you mean:

1, _Inheritance_: Create a derived class (e.g., OpenDuckMiniMujocoSystemInterface : public MujocoSystemInterface) that adds filtering?

2, _Composition_: Create an external package that holds a reference to MujocoSystemInterface and uses getData()/getModel() to implement filtering?

Or ... 
3, change the current PR and remove all consumer logic (debouncing/filtering) from the MujocoSystemInterface. The HW interface now only provides:
- Raw contact detection via contact_raw state interface
- Register contact detection separately from sensors (via register_contact_detection()) to reflect that it reads from mjData->contact[] rather than mjData->sensordata[].
- In my ros2_control motion controller, read contact_raw from state interfaces, use getData()/getModel() to access MuJoCo data directly, implement additional filtering/debouncing logic etc

Juliaj:
Closing, we're going with different route https://github.com/ros-controls/mujoco_ros2_control/pull/101.

Issue #70, opened 2026-01-15
Crash on shutdown/teardown in headless/CI environments due to camera rendering cleanup


Juliaj:
**Describe the bug**
Crash on shutdown/teardown in headless/CI environments due to camera rendering cleanup being done from the wrong thread/context, and because disable_rendering was not consistently preventing camera/lidar threads from starting.

When running with headless/CI settings (and/or disable_rendering=true), the system may still construct and start camera/lidar background threads that create an offscreen OpenGL context. During shutdown, camera cleanup can run from a thread without a current GL context. If mjr_freeContext() (or OpenGL resource cleanup inside it like glDeleteRenderbuffers) runs without a current GL context, it can segfault.

**To Reproduce**
Run contact sensor unit tests
```
colcon test --packages-select mujoco_ros2_control --ctest-args -R test_contact_sensors --event-handlers console_direct+
```
And tests will fail with -11.

saikishor:
Thanks for reporting it. Can we fix this in a separate PR?

Juliaj:
> [@Juliaj](https://github.com/Juliaj) just curious how often you see the above crash?

It is 100% repro for me with the new unit tests from contact sensor because they creates/destroys MujocoSystemInterface in-process with headless/disable_rendering.

PR #71, opened 2026-01-15
Fix the unit tests crash due to camera rendering cleanup

Juliaj:
Addresses https://github.com/ros-controls/mujoco_ros2_control/issues/70

**Summary for the fix**
- Honor disable_rendering=true by skipping creation and startup of camera/lidar wrappers (no rendering threads when disabled).
- Make camera cleanup GL-safe: free MuJoCo rendering context/scene inside the camera rendering thread while its OpenGL context is current; keep close() as “stop + join + glfwTerminate”.

Without this fix, the contact sensor unit tests fail constantly.


saikishor:
Let me check

PR #75, opened 2026-01-16
Publish the floating base state of the free joint attached to the robot

saikishor:
This PR makes the simulator publish a floating base state topic of the robot's ground truth within the simulator

### DIFF
--- a/src/mujoco_system_interface.cpp
+++ b/src/mujoco_system_interface.cpp
@@ -878,12 +878,12 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
   }
 
   // Check for free joint
-  std::string floating_base_joint_name =
-      get_hardware_parameter_or(get_hardware_info(), "floating_base_joint_name", "floating_base_joint");
+  std::string odom_free_joint_name =
+      get_hardware_parameter_or(get_hardware_info(), "odom_free_joint_name", "floating_base_joint");
   for (int i = 0; i < mj_model_->njnt; ++i)
   {
     const char* joint_name = mj_id2name(mj_model_, mjtObj::mjOBJ_JOINT, i);
-    if (floating_base_joint_name == joint_name)
+    if (odom_free_joint_name == joint_name)
     {
       if (mj_model_->jnt_type[i] == mjJNT_FREE)
       {
@@ -893,8 +893,9 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
       }
       else
       {
-        RCLCPP_FATAL(get_logger(), "Unable to use joint '%s' as floating base joint since it is not a free joint.",
-                     floating_base_joint_name.c_str());
+        RCLCPP_FATAL(get_logger(),
+                     "Unable to use joint '%s' to publish the floating base state since it is not a free joint.",
+                     odom_free_joint_name.c_str());
         return hardware_interface::CallbackReturn::FAILURE;
       }
     }
@@ -904,7 +905,7 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
   {
     // Odometry publisher
     std::string odom_topic_name =
-        get_hardware_parameter_or(get_hardware_info(), "floating_base_state_topic", "/simulator/floating_base_state");
+        get_hardware_parameter_or(get_hardware_info(), "odom_topic", "/simulator/floating_base_state");
     floating_base_publisher_ = mujoco_node_->create_publisher<nav_msgs::msg::Odometry>(odom_topic_name, 100);
     floating_base_realtime_publisher_ =
         std::make_shared<realtime_tools::RealtimePublisher<nav_msgs::msg::Odometry>>(floating_base_publisher_);
--- a/README.md
+++ b/README.md
@@ -104,10 +104,10 @@ Just specify the plugin and point to a valid MJCF on launch:
 
       <!--
         The optional parameters to choose the name of the floating base joint and the topic name to publish the
-        floating base state of the robot. The provided joint needs to be a MuJoCo free joint
+        odometry of the robot. The provided joint needs to be a valid MuJoCo free joint.
       -->
-      <param name="floating_base_joint_name">floating_base_joint</param>
-      <param name="floating_base_state_topic">/simulator/floating_base_state</param>
+      <param name="odom_free_joint_name">floating_base_joint</param>
+      <param name="odom_topic">/simulator/floating_base_state</param>
     </hardware>
   ...
 ```


src/mujoco_system_interface.cpp
```diff
@@ -877,6 +877,39 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
...
 
+  // Check for free joint
+  for (int i = 0; i < mj_model_->njnt; ++i)
+  {
+    if (mj_model_->jnt_type[i] == mjJNT_FREE)
+    {
+      free_joint_id_ = i;
+      free_joint_qpos_adr_ = mj_model_->jnt_qposadr[i];
+      free_joint_qvel_adr_ = mj_model_->jnt_dofadr[i];
+      break;  // Assume only one free joint for the robot base
```
ndunkelb-nasa: Hmm, I'm not sure I like this assumption. If you add any other free standing objects to the sim, this will be invalid. For example we do object interaction, and those objects are free joints. Could this instead be a param passed to the HWI? Maybe like `odom_publisher_joint_name`? This could make the logic a lot more explicit too.

saikishor: @ndunkelb-nasa Done

src/mujoco_system_interface.cpp
```diff
@@ -877,6 +877,50 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
     return hardware_interface::CallbackReturn::FAILURE;
   }
 
+  // Check for free joint
+  std::string floating_base_joint_name =
+      get_hardware_parameter_or(get_hardware_info(), "floating_base_joint_name", "floating_base_joint");
```
ndunkelb-nasa: Maybe could make this more specifically `odom_freejoint_name`? I think that gives some more clarity that it is used to publish an odom msg, and that it should be of type `freejoint` in the mjcf.

saikishor: I've documented that in the REAME.md. Do you want me to change the parameter naming too?

ndunkelb-nasa: I think it is technically fine either way, but is there a reason you prefer `floating_base_joint_name` over something that includes `odom`? Its good to have it in the readme, but a more descriptive name also helps.

saikishor: Nope no preference. I can change it

saikishor: Done

saikishor:
> Strangely, the test robot works fine for me, so I don't know why my robot without a "floating_base_joint" (just renamed it) has issues.

@ndunkelb-nasa  I see it too. For me it is failing here:
https://github.com/ros-controls/mujoco_ros2_control/blob/154cdb79b2cfa4c61c94a8ab9b9e8cddaa736c9e/src/mujoco_system_interface.cpp#L442-L448

```
[make_mjcf_from_robot_description-5] arm_right_tool_link ('pelvis_base_link', [-0.0918000000000001, 0, -0.0455], [1.0, 0, 0, 0])
[ros2_control_node-6] [INFO] [1768593593.078422121] [mujoco_node]: Error Error: body 'base_link' not found in bodypair 0
[ros2_control_node-6] Element name '', id 0, line 822
[ros2_control_node-6] [FATAL] [1768593593.078482512] [mujoco_node]: Failed to compile MuJoCo model: 
[ros2_control_node-6] Stack trace (most recent call last) in thread 8421:
[ros2_control_node-6] #16   Object "", at 0xffffffffffffffff, in 
[ros2_control_node-6] #15   Source "../sysdeps/unix/sysv/linux/x86_64/clone3.S", line 81, in __clone3 [0x7d2453f948bf]
[ros2_control_node-6] #14   Source "./nptl/pthread_create.c", line 442, in start_thread [0x7d2453f02ac2]
[ros2_control_node-6] #13   Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30", at 0x7d245427a252, in 
[ros2_control_node-6] #12   Object "/opt/ros/humble/lib/librclcpp.so", at 0x7d24544cf809, in rclcpp::executors::MultiThreadedExecutor::run(unsigned long)
[ros2_control_node-6] #11   Object "/opt/ros/humble/lib/librclcpp.so", at 0x7d24544c847e, in rclcpp::Executor::execute_any_executable(rclcpp::AnyExecutable&)
[ros2_control_node-6] #10   Object "/opt/ros/humble/lib/librclcpp.so", at 0x7d24544c7c7b, in rclcpp::Executor::execute_subscription(std::shared_ptr<rclcpp::SubscriptionBase>)
[ros2_control_node-6] #9    Object "/opt/pal/alum/lib/libcontroller_manager.so", at 0x7d245473bc8e, in 
[ros2_control_node-6] #8    Object "/opt/pal/alum/lib/libcontroller_manager.so", at 0x7d24546b3cda, in controller_manager::ControllerManager::robot_description_callback(std_msgs::msg::String_<std::allocator<void> > const&)
[ros2_control_node-6] #7    Object "/opt/pal/alum/lib/libcontroller_manager.so", at 0x7d24546b1b11, in controller_manager::ControllerManager::init_resource_manager(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
[ros2_control_node-6] #6    Object "/opt/pal/alum/lib/libhardware_interface.so", at 0x7d2453d3da52, in hardware_interface::ResourceManager::load_and_initialize_components(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned int)
[ros2_control_node-6] #5    Object "/opt/pal/alum/lib/libhardware_interface.so", at 0x7d2453d48efa, in 
[ros2_control_node-6] #4    Object "/opt/pal/alum/lib/libhardware_interface.so", at 0x7d2453d6623a, in hardware_interface::HardwareComponent::initialize(hardware_interface::HardwareComponentParams const&)
[ros2_control_node-6] #3    Object "/opt/pal/alum/lib/libhardware_interface.so", at 0x7d2453d6569e, in 
[ros2_control_node-6] #2    Object "/home/user/kangaroo_mujoco/build/mujoco_ros2_control/libmujoco_ros2_control.so", at 0x7d241eaf8b55, in mujoco_ros2_control::MujocoSystemInterface::on_init(hardware_interface::HardwareComponentInterfaceParams const&)
[ros2_control_node-6] #1    Object "/home/user/kangaroo_mujoco/build/mujoco_ros2_control/libmujoco_ros2_control.so", at 0x7d241eaf4913, in mujoco_ros2_control::LoadModel(char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, mujoco::Simulate&, std::shared_ptr<rclcpp::Node>)
[ros2_control_node-6] #0    Object "/home/user/kangaroo_mujoco/build/mujoco_ros2_control/libmujoco_ros2_control.so", at 0x7d241eaf43fe, in mujoco_ros2_control::loadModelFromTopic(std::shared_ptr<rclcpp::Node>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
[ros2_control_node-6] Segmentation fault (Address not mapped to object [0x10])

```

However, the same happens to me also with the `main` branch. Is that your case too?

ndunkelb-nasa:
Hmm, that looks like that is failing just on compilation of your model, which should be before any of these changes. You sure your model is still valid?

Main works fine for me.

Also worth noting though, we aren't using the on-the-fly mjcf creation yet. We are still using static mjcf files.

saikishor:
@ndunkelb-nasa ok no idea. I don't find anything weird. It works if I name the free joint differently and rerun it. We can leave it for later.

ndunkelb-nasa:
> Do you have the stack trace? Is it similar to mine?

Strangely its not seg-faulting, it is doing the thing again where it seems to open, but continuously reset. Let me see what version I was running before that was working so we can see what changed

ndunkelb-nasa:
> Ahh got it. Is there any other type of joint other than free joint in your case?

I think it has something to do with setting start positions. If I provide a `override_start_position_file` param, it actually works fine.

saikishor:
SO, if you are setting the joints from the ros2_control tags, then it should be from here

https://github.com/ros-controls/mujoco_ros2_control/blob/154cdb79b2cfa4c61c94a8ab9b9e8cddaa736c9e/src/mujoco_system_interface.cpp#L2098-L2109

We had to add it to have it working for our humnaoid though. That's weird. I tested it even today with this fix.

saikishor:
You should be able to pin point the joint by looking at the data in /joint_states or /mujoco_actuators_states

saikishor:
> Hmm, those topics actually never come up, and found that actually after I ctrl + c the node, I get this popping up
> 
> ```
> [ros2_control_node-20] WARNING: Nan, Inf or huge value in QPOS at DOF 8. The simulation is unstable. Time = 0.0000.
> ```

So, the initial position resulted in having NaNs already at the first cycles 🙈

saikishor:
@ndunkelb-nasa Can you try commenting these lines?
https://github.com/ros-controls/mujoco_ros2_control/blob/154cdb79b2cfa4c61c94a8ab9b9e8cddaa736c9e/src/mujoco_system_interface.cpp#L2098-L2109

PR #77, opened 2026-01-16
Fix invalid initial state + filter out ball joints 

saikishor:
After Nathan's discussion, may be it is also better filter out these joints as they have 3DoF

The initial state is invalid when the model has many passive joints and these joints are not exported into the ros2_control interfaces, then their interfaces are left out as NaNs initially, so if that's the case better set them to intial value automatically.


ndunkelb-nasa (dismissed):
Thanks for the quick fix! Works on my machine!

PR #80, opened 2026-01-20
Fix the clock publish rate based on the mj_step

saikishor:
FYI before this fix the clock is published at around 860 Hz and with this fix the deuplicates are removed and it is synced with the mujoco step

### DIFF
--- a/scripts/make_mjcf_from_robot_description.py
+++ b/scripts/make_mjcf_from_robot_description.py
@@ -259,6 +259,39 @@ def rename_material_textures(dir_path, modifier):
         path.rename(new_path)
 
 
+def extract_rgba(visual):
+    """
+    Extracts the rgba values for a trimesh object. This apparently is pretty tricky because the materials can
+    come in a number of forms. This is basically chat gpt's best guess of how to do this, but seems to work
+    for my cases.
+
+    :param visual: visual component of a trimesh scene
+    """
+    # 1) Material base color
+    mat = getattr(visual, "material", None)
+    if mat is not None:
+        # PBR
+        if isinstance(mat, trimesh.visual.material.PBRMaterial) and mat.baseColorFactor is not None:
+            return np.array(mat.baseColorFactor)
+
+        # Simple / Phong
+        if hasattr(mat, "diffuse") and mat.diffuse is not None:
+            rgb = np.array(mat.diffuse)
+            alpha = getattr(mat, "alpha", 1.0)
+            return np.concatenate([rgb, [alpha]])
+
+    # 2) Per-vertex colors
+    if visual.kind == "vertex" and visual.vertex_colors is not None:
+        vc = visual.vertex_colors
+        if vc.shape[1] == 4:
+            return vc[0] / 255.0
+        else:
+            return np.append(vc[0] / 255.0, 1.0)
+
+    # 3) I don't think this should ever happen, but just in case
+    return np.array([0.7, 0.7, 0.7, 1.0])
+
+
 def convert_to_objs(mesh_info_dict, directory, xml_data, convert_stl_to_obj, decompose_dict):
     # keep track of files we have already processed so we don't do it again
     converted_filenames = []
@@ -307,7 +340,9 @@ def convert_to_objs(mesh_info_dict, directory, xml_data, convert_stl_to_obj, dec
                     # make a material for the rgba values and export it
                     mtl_modifier = f"m{mtl_num}"
                     mtl_name = "mtl_" + mtl_modifier
-                    material = trimesh.visual.material.SimpleMaterial(name=mtl_name, diffuse=rgba)  # RGBA
+                    material = trimesh.visual.material.SimpleMaterial(
+                        name=mtl_name, diffuse=rgba, glossiness=1000, specular=[0.2, 0.2, 0.2]
+                    )  # RGBA
                     mesh.visual = trimesh.visual.TextureVisuals(material=material)
 
                     # increment material number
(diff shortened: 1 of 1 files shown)


src/mujoco_system_interface.cpp
```diff
@@ -2412,6 +2424,7 @@ void MujocoSystemInterface::PhysicsLoop()
       std::this_thread::sleep_for(std::chrono::milliseconds(1));
     }
 
+    bool publish_clock_now = false;
```
eholum-nasa: Maybe a little redundant with the comment below but would rather make things as obvious as possible

```suggestion
    // We only publish the /clock if the simulation was advanced to avoid duplicate timestamps
    bool publish_clock_now = false;
```

src/mujoco_system_interface.cpp
```diff
@@ -2408,6 +2422,8 @@ void MujocoSystemInterface::PhysicsLoop()
       std::this_thread::sleep_for(std::chrono::milliseconds(1));
     }
 
+    // We only publish the /clock if the simulation was advanced to avoid duplicate timestamps
+    bool publish_clock_now = false;
```
ndunkelb-nasa: Shouldn't we actually be publishing the clock here? In the current implementation, because of the while loop, if we get caught a bit behind, we might not actually publish the clock for a couple of cycles, even though the physics loop rate is 500hz for example. This means that the CM will not get hits every 2ms as hoped, it will instead get it sometimes at 2ms and sometimes at 4ms, etc. My clock looks like this (I assume because I am running at a heavy cpu load, it misses more frequenctly than others)
```sh
clock:
  sec: 791
  nanosec: 487999995
---
clock:
  sec: 791
  nanosec: 489999995
---
clock:
  sec: 791
  nanosec: 495999995
---
clock:
  sec: 791
  nanosec: 499999995
---
clock:
  sec: 791
  nanosec: 501999995
```

saikishor: This is weird that you are skipping the cycles completely. May be I can update the logic to only publish If there is a change in time?. Will that work for you?

ndunkelb-nasa: I imagine it is because of [this](https://github.com/ros-controls/mujoco_ros2_control/blob/main/src/mujoco_system_interface.cpp#L2475-L2477) while loop. Although I can't say I understand 100% what the condition is checking for the while loop. But the comment seems to support what I was saying. 

I don't think only publishing if there is a change in time will change what I am seeing. I think putting the `publish_clock()` call where the `publish_clock_now` param is set to true would solve it? That would never get called if the sim is not stepped

ndunkelb-nasa: I just tried it and I am getting 500 hz pub rates on /clock now (was 285 ish before) and not missing any time steps. 

If Yall are good with that, I can go ahead and commit it, but want to make sure I am understanding correctly.

saikishor: That's good. I'm curious about it. Please go ahead and push it

ndunkelb-nasa: pushed it! Go ahead and take a look

saikishor: It runs at 500 Hz very good

However, I see some instances, the clock is published at 0.02 seconds apart at the beginning and later instances 0.002 secs?  Is that normal? I'm sure it has nothing to do with your change

```
---
clock:
  sec: 3
  nanosec: 29999999
---
clock:
  sec: 3
  nanosec: 31999999
---
clock:
  sec: 3
  nanosec: 33999999
---
clock:
  sec: 3
  nanosec: 35999999
---
clock:
  sec: 3
  nanosec: 37999999
```

Later

```
clock:
  sec: 99
  nanosec: 199999999
---
clock:
  sec: 99
  nanosec: 201999999
---
clock:
  sec: 99
  nanosec: 203999999
---
clock:
  sec: 99
  nanosec: 205999999
---
clock:
  sec: 99
  nanosec: 207999999
---
clock:
  sec: 99
  nanosec: 209999999

```

saikishor: If this is normal, I'm happy with this changes. It could be the difference between the two loops we have

ndunkelb-nasa: Hmm, that is strange, and it doesn't happen for me (at least in the couple of times I tested). Does that happen as soon as you start? Even at the start, my times look good


```sh
clock:
  sec: 0
  nanosec: 2000000
---
clock:
  sec: 0
  nanosec: 4000000
---
clock:
  sec: 0
  nanosec: 6000000
---
clock:
  sec: 0
  nanosec: 8000000
---
clock:
  sec: 0
  nanosec: 10000000
```

saikishor: But yours is reporting 0.2 seconds? Instead of default 0.002?

ndunkelb-nasa: Actually the same misreading you did on my logs is the same thing you did on your logs. There are fewer 0s in the first couple because it won't report leading 0s to make it 0 digits for nanoseconds. So mine could be reading
```
clock:
  sec: 0
  nanosec: 002000000
---
clock:
  sec: 0
  nanosec: 004000000
---
clock:
  sec: 0
  nanosec: 006000000
---
clock:
  sec: 0
  nanosec: 008000000
---
clock:
  sec: 0
  nanosec: 010000000
```

Go back and check yours too! It is actually 2 ms as well. You just happened to catch it right at the start of the 3s logs, so there are only 8 digits instead of 9.

So we should be good to go!

saikishor: So dumb of me LOL. Sorry for wasting time 😅. I'm very much fine with the change. Can you approve it?

ndunkelb-nasa: All good, I was confused too for a bit!

PR #81, opened 2026-01-21
Fixing pixi install from recent updates

eholum-nasa:
A few rosdeps have been added recently for tests which was causing failures and the lock file had gone pretty out of date.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -43,7 +43,6 @@ mold = ">=2.40.2,<3"
 libcap = ">=2.76,<3"
 glfw = ">=3.4,<4"
 filelock = ">=3.20.0,<4"
-trimesh = ">=4.10.1"
 pip = "*"
 
 # ROS dependencies
@@ -68,4 +67,5 @@ ros-jazzy-launch-testing-ament-cmake = "*"
 xacro = ">=2.1.1, <3"
 obj2mjcf = ">=0.0.25"
 mujoco = "==3.4.0"
+trimesh = "*"
 scipy = ">=1.17.0, <2"
--- a/pixi.lock
+++ b/pixi.lock
@@ -543,7 +543,6 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/tinyxml2-11.0.0-h3f2d84a_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/tk-8.6.13-noxft_ha0e22de_103.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/tomli-2.4.0-pyhcf101f3_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/noarch/trimesh-4.11.1-pyh7b2049a_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/typeguard-4.4.4-pyhd8ed1ab_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/typing-extensions-4.15.0-h396c80c_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/typing_extensions-4.15.0-pyhcf101f3_0.conda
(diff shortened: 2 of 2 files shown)


pixi.toml
```diff
@@ -50,12 +50,14 @@ pip = "*"
...
 ros-jazzy-ament-cmake = "*"
 ros-jazzy-rclcpp = "*"
+ros-jazzy-ros2-control = "*"
 ros-jazzy-controller-manager = "*"
 ros-jazzy-control-toolbox = "*"
 ros-jazzy-hardware-interface = "*"
 ros-jazzy-transmission-interface = "*"
 ros-jazzy-joint-state-broadcaster = "*"
 ros-jazzy-position-controllers = "*"
+ros-jazzy-gripper-controllers = "*"
```
christophfroehlich: this [package](https://control.ros.org/jazzy/doc/ros2_controllers/doc/migration.html#gripper-action-controller) is deprecated, and not used here?
https://github.com/ros-controls/mujoco_ros2_control/blob/00de2a4fd3fbe07a9cf2e3ece7d710b6083ac6b6/test/config/controllers.yaml#L12
This is actually also [deprecated](https://control.ros.org/rolling/doc/ros2_controllers/doc/migration.html#position-controllers) on current rolling version.

christophfroehlich (changes requested):

pixi.toml
saikishor: Yess

saikishor: Actually, with the updated tests, this is not needed anymore

saikishor:
```suggestion
```

eholum-nasa:
I'm honestly not 100% sure how it could have worked before - since AFAICT it definitely should have failed? My best guess is that build environments were being cached magically somehow... The most significant change I can see from[ recently successful runs](https://github.com/ros-controls/mujoco_ros2_control/actions/runs/21001711546/job/60374940518?pr=65) (last week) vs [more recently failed runs](https://github.com/ros-controls/mujoco_ros2_control/actions/runs/21135078728/job/60775107699?pr=73) (two days ago) is runner version updates and cache invalidations... That or something could have changed in upstream pixi and robostack?

In any case, this is probably what we should be doing anyway... I would add the sourcing to the activation steps but that seems to have other consequences doing things before the workspace is built.

saikishor (dismissed):

saikishor:
Perfect. Good to know.

pixi.toml
```diff
@@ -50,6 +50,7 @@ pip = "*"
 ros-jazzy-ros-base = "*"
 ros-jazzy-ament-cmake = "*"
 ros-jazzy-rclcpp = "*"
+ros-jazzy-ros2-control = "*"
```
christophfroehlich: Sorry, I haven't had time to catch up with this repo. What was added to make this necessary? `ros2_control` is only a metapackage. either, remove the following lines, or add the missing package instead of "full" ros2_control.

```
ros-jazzy-controller-manager = "*"
ros-jazzy-hardware-interface = "*"
ros-jazzy-transmission-interface = "*"
```

pixi.toml
```diff
@@ -56,13 +56,17 @@ ros-jazzy-hardware-interface = "*"
...
 ros-jazzy-urdfdom = "*"
 ros-jazzy-urdfdom-py = "*"
 ros-jazzy-rviz2 = "*"
+ros-jazzy-launch-testing-ament-cmake = "*"
 
 [pypi-dependencies]
 xacro = ">=2.1.1, <3"
 obj2mjcf = ">=0.0.25"
 mujoco = "==3.4.0"
+trimesh = "*"
```
saikishor: The trimesh is already added in line 46
```suggestion
```

eholum-nasa: Oh whoops wrong button... I'll update but maybe we should pull it in from pypi rather than apt for consistency with the virtual env?

saikishor: Sure. I'm fine with both

pixi.toml
eholum-nasa: Ah thanks! Updated, that was laziness from odd test errors locally...

PR #82, opened 2026-01-21
Updated ament_index_cpp API

ahcorde:
Related with https://github.com/ament/ament_index/pull/104

Warning https://build.ros2.org/view/Rdev/job/Rdev__mujoco_ros2_control__ubuntu_noble_amd64/15/


src/mujoco_system_interface.cpp
```diff
@@ -758,8 +758,16 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
                                             /* is_passive = */ false);
 
       // Add ros2 control icon for the taskbar
-      std::string icon_location =
+      std::string icon_location;
+#if AMENT_INDEX_CPP_VERSION_GTE(1, 13, 0)
```
saikishor, line 762: @ahcorde Seems like this macro is not found. I don't see the ament_index_cpp package having version.h header

saikishor:
Addressed in https://github.com/ros-controls/mujoco_ros2_control/pull/229

PR #83, opened 2026-01-21
Add markdown link check, disable rolling, and minor tag improvements

eholum-nasa:
Sorry for stacking up a few changes for CI here.

* Added a markdown link check just to make sure we don't break any references when moving things around
* Using docker metadata actions instead of sed to get valid tags regardless of branch names
* Disabling rolling while that image seems to be broken...

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,7 +29,9 @@ jobs:
     timeout-minutes: 15
     strategy:
       matrix:
-        ros_distro: [humble, jazzy, kilted, rolling]
+        # TODO: Re-enable CI on rolling after the image is fixed:
+        # https://openrobotics.zulipchat.com/#narrow/channel/526029-ROS-PMC-.28Restricted-posting.29/topic/2026-01-22.20Rolling.20Docker.20image.20out-of-date/near/569502119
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -29,24 +29,29 @@ jobs:
...
+        uses: docker/metadata-action@v5
+        with:
+          images: mujoco_ros2_control
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
       - name: Build and push
         uses: docker/build-push-action@v6
         with:
+          context: .
```
saikishor, line 52: Do we need this here?


```suggestion
```

saikishor (commented):
Rest LGTM. 

I'm fine to disable rolling builds for the time being

.github/workflows/ci.yaml
eholum-nasa: There were some odd errors w.r.t. the path when GitHub was having CI issues so I pushed this... It's the default but maybe it doesn't hurt to make it painfully obvious?

https://github.com/docker/build-push-action?tab=readme-ov-file#path-context

saikishor: Sure thing!

PR #84, opened 2026-01-21
check that joint name exists before comparison

ndunkelb-nasa:
I found that in one of our robots, we had a freejoint defined without a name, and this just resulted in a segfault because we assume every joint has a name. 

This change resolves that issue, but also makes me wonder - should we **require** that every joint has a name? If the answer is no, then we have other places that will fail as well, at least [here](https://github.com/ros-controls/mujoco_ros2_control/blob/839e07ec31bab5406d5da0004a0dd1486b2c4482/src/mujoco_system_interface.cpp#L1691) when I did a quick test.

### DIFF
--- a/src/mujoco_system_interface.cpp
+++ b/src/mujoco_system_interface.cpp
@@ -969,6 +969,20 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
   // We can still turn this on manually if desired.
   sim_->opt.flags[mjVIS_RANGEFINDER] = false;
 
+#if !ROS_DISTRO_HUMBLE
+  // Verify the update rate
+  const mjtNum desired_timestep = 1.0 / static_cast<double>(get_hardware_info().rw_rate);
+  const bool under_sampled = mj_model_->opt.timestep > desired_timestep;
+  RCLCPP_WARN_EXPRESSION(
+      get_logger(), under_sampled,
+      "MuJoCo simulator frequency %lu Hz (timestep %.6f sec) is smaller than the controller manager's update rate %lu "
+      "Hz. The simulation may be under-sampled and this means that there will be some discrepancies in the rate at "
+      "which controllers update cycles run. Either increase the MuJoCo timestep or decrease the controller manager's "
+      "update rate.",
+      static_cast<unsigned long>(1.0 / mj_model_->opt.timestep), mj_model_->opt.timestep,
+      static_cast<unsigned long>(get_hardware_info().rw_rate));
+#endif
+
   // When the interface is activated, we start the physics engine.
   physics_thread_ = std::thread([this, headless]() {
     // Load the simulation and do an initial forward pass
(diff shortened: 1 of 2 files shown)


saikishor (dismissed):
I would say we should expect that the joints have proper naming. If not it is very hard to handle each and every case. 

This works for me. Thank you

ndunkelb-nasa:
> Agreed, plus isn't the free joint basically the only possible joint that wouldn't have a name if converting from a URDF (at least for now)?

I think that would even have a name on it. I think this only shows up if you add your own custom things, like maybe something an intern might do like [here](https://github.com/NASA-JSC-Robotics/chonkur_l_raile/blob/afe627eec4c8f2c3e12e18095581f9af424a2a2c/clr_mujoco_config/description/ctb.xml#L34) @njhart-nasa

src/mujoco_system_interface.cpp
```diff
@@ -870,6 +870,17 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
...
+  // We explicitly check to make sure that all joints have names, otherwise stuff down the line won't work
+  for (int i = 0; i < mj_model_->njnt; ++i)
+  {
+    const char* joint_name = mj_id2name(mj_model_, mjtObj::mjOBJ_JOINT, i);
+    if (!joint_name)
+    {
+      RCLCPP_FATAL(get_logger(), "Joint %d does not have a name. All joints must have names.", i);
+      return hardware_interface::CallbackReturn::FAILURE;
+    }
+  }
```
saikishor, line 882:
```suggestion
  bool all_joints_with_names = true;
  for (int i = 0; i < mj_model_->njnt; ++i)
  {
    const char* joint_name = mj_id2name(mj_model_, mjtObj::mjOBJ_JOINT, i);
    if (!joint_name)
    {
      RCLCPP_FATAL(get_logger(), "Joint %d does not have a name. All joints must have names.", i);
      all_joints_with_names = false;
    }
  }
  if(!all_joints_with_names)
  {
    return hardware_interface::CallbackReturn::FAILURE;
  }
```

Maybe this way all joints will be listed once?

ndunkelb-nasa: I like it! The only unfortunate part is that bc there are no names, it is hard to give useful information to the user. I don't have any ideas of how to do anything better though. Knowing that joint 17 doesn't have a name isn't super useful, other than knowing to go look through all the joints.

saikishor: You are right XD. Maybe we print the count of joints without names instead?. Like **there are 5 joints without name. All joints must have names**?

ndunkelb-nasa: Yeah, I think that is a fair compromise. Lets do this one more time!

src/mujoco_system_interface.cpp
```diff
@@ -883,6 +900,8 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
   for (int i = 0; i < mj_model_->njnt; ++i)
   {
     const char* joint_name = mj_id2name(mj_model_, mjtObj::mjOBJ_JOINT, i);
+
+    // first check that the joint name exists, then do comparison
```
saikishor:
```suggestion
```

saikishor: Last one, I think this is not needed anymore

PR #85, opened 2026-01-21
Set parameters per controller instead of central parameters

saikishor:
I observed this while testing the simulator at 1000 Hz, the joint_states are running at 500 Hz and later I realized that we are not setting the update_rate for the broadcaster, so it is running at the frequency of the controller manager. Can we add the parameters per controller instead?

### DIFF
--- a/src/mujoco_system_interface.cpp
+++ b/src/mujoco_system_interface.cpp
@@ -870,6 +870,23 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
   actuator_state_realtime_publisher_ =
       std::make_shared<realtime_tools::RealtimePublisher<sensor_msgs::msg::JointState>>(actuator_state_publisher_);
 
+  // We explicitly check to make sure that all joints have names, otherwise stuff down the line won't work
+  int num_joints_without_name = 0;
+  for (int i = 0; i < mj_model_->njnt; ++i)
+  {
+    const char* joint_name = mj_id2name(mj_model_, mjtObj::mjOBJ_JOINT, i);
+    if (!joint_name)
+    {
+      num_joints_without_name++;
+    }
+  }
+  if (num_joints_without_name)
+  {
+    RCLCPP_FATAL(get_logger(), "%d joints in the mjcf don't have names. All joints must have names.",
+                 num_joints_without_name);
+    return hardware_interface::CallbackReturn::FAILURE;
+  }
+
   // Register all MuJoCo actuators
   if (!register_mujoco_actuators())
   {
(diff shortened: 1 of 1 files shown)


test/config/controllers.yaml
```diff
@@ -1,23 +1,21 @@
...
-
-    position_controller:
-      type: position_controllers/JointGroupPositionController
+    update_rate: 1000
 
-    gripper_controller:
-      type: position_controllers/JointGroupPositionController
+joint_state_broadcaster:
+  ros__parameters:
+    update_rate: 50
```
ndunkelb-nasa, line 7: I actually haven't done this before, so I tried running this, and this was the result. Do you get the same? I am running on humble btw

```sh
$ ros2 topic hz /joint_states -w 50
average rate: 24.989
	min: 0.039s max: 0.041s std dev: 0.00033s window: 26
average rate: 25.006
	min: 0.039s max: 0.041s std dev: 0.00039s window: 50
average rate: 25.006
	min: 0.039s max: 0.041s std dev: 0.00047s window: 50
average rate: 25.012
	min: 0.039s max: 0.041s std dev: 0.00046s window: 50
```

eholum-nasa: You're not running at 50% speed or anything are you? This doubles/halves as I speed and slow the sim, but looks right on jazzy,

```
ros2 topic hz /joint_states -w 50
average rate: 49.995
        min: 0.019s max: 0.021s std dev: 0.00050s window: 50
average rate: 49.961
        min: 0.019s max: 0.021s std dev: 0.00060s window: 50
average rate: 49.995
        min: 0.019s max: 0.021s std dev: 0.00049s window: 50
average rate: 49.971
        min: 0.019s max: 0.021s std dev: 0.00048s window: 50
```

saikishor: Nope, for me it runs perfectly

```
$ ros2 topic hz  /joint_states
average rate: 50.016
	min: 0.019s max: 0.021s std dev: 0.00054s window: 49
average rate: 49.973
	min: 0.019s max: 0.022s std dev: 0.00056s window: 99
average rate: 49.981
	min: 0.019s max: 0.022s std dev: 0.00054s window: 149
average rate: 49.999
	min: 0.019s max: 0.022s std dev: 0.00053s window: 199
average rate: 49.989
	min: 0.019s max: 0.022s std dev: 0.00054s window: 249
```
$ ros2 param get /controller_manager update_rate
Integer value is: 100

saikishor: This is super weird the scheduling part should work the same on Humble. I can take a look on Humble

ndunkelb-nasa: Alright, I ran it again and it was 50. Super strange... Let me see if I can recreate it

ndunkelb-nasa: Alright I can't recreate it. The only thing I can guess is that maybe I had some other things going that I forgot to close. But the fact it was an exact multiple is pretty strange.... Im convinced enough that it is good to go!

saikishor: Awesome 🙌🏽🙌🏽

Issue #74, opened 2026-01-16
Key frame loading from scene


SergiMuac:
Right now, the robot simulation can start in different states:
- all joints and controls to 0.0
- If state_interface initial position is defined, it uses that.
- If a keyframe is saved in a file and you pass the filepath as in [here](https://github.com/ros-controls/mujoco_ros2_control/blob/main/README.md?plain=1#L77) then that position is used.

However, I think it is much better to provide a parameter where the name of an existing keyframe is provided, and it searches for it in the xml file and uses that information to initialize the simulation.
This feature is much more useful than providing the file path that contains the keyframe, because you can have many keyframes in the same file or even build them dynamically using xacro, and at the end, ROS2 Control will load the one with the specified name.

Thanks!

PR #88, opened 2026-01-23
Add `reset_world` service functionality

saikishor:
This PR adds reset_world functionality

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,6 +405,24 @@ The lidar sensor is then configurable through ROS 2 control xacro with:
   </ros2_control>
 ```
 
+## Simulation - Topics and Services
+
+### Topics
+
+* `/mujoco_actuators_states` (`sensor_msgs/msg/JointState`): Provides information on all internal MuJoCo joints,
+regardless of whether their interfaces are exposed via `ros2_control`.
+* `/clock` (`rosgraph_msgs/msg/Clock`): Contains the internal physics clock tracked by each MuJoCo simulation step.
+
+### Services
+
+* `~/reset_world` (`std_srvs/srv/Empty`): Resets the simulation to the robot's initial spawning state.
+
+> [!IMPORTANT]
+> If controllers are active during the service call, the robot may reset to the initial state and then immediately
+> snap back to its previous commanded position. To avoid this, it is recommended to deactivate any active joint
+> controllers before calling this service.
+
+
 ## Test Robot System
 
 While examples are limited, we maintain a functional example 2-dof robot system in the [test examples](./mujoco_ros2_control/test/test_resources/test_robot.urdf) space.


mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -951,14 +951,28 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
   initialize_initial_positions(get_hardware_info());
   set_initial_pose();
 
+  // Store initial state for reset_world service
+  {
+    const std::unique_lock<std::recursive_mutex> lock(*sim_mutex_);
+    initial_qpos_.assign(mj_data_->qpos, mj_data_->qpos + mj_model_->nq);
+    initial_qvel_.assign(mj_data_->qvel, mj_data_->qvel + mj_model_->nv);
+    initial_ctrl_.assign(mj_data_->ctrl, mj_data_->ctrl + mj_model_->nu);
+  }
```
ndunkelb-nasa, line 960: This definitely seems like a pretty straightforward way to do this.

However, I am curious if we make this a little bit more modular we could enable loading sim states in a variety of ways, not just from the initial position. For example, as mentioned in #74, it would be nice to be able to load several keyframes. And that could be integrated into this in a number of ways, including:
* provide a filepath to a key frame in the service call
* load a set of keyframes at initialization and choose from them through an integer in the service call
* just put several key frames in the mjcf file as suggested in #74 (although I am not the biggest fan of this idea)

And if we made the loading a little bit more modular, this could support that in the future (I don't think it necessarily needs to be part of this PR)

saikishor: Understood 🙌🏽🙌🏽. Sure thing. I have it in plan already

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -2419,13 +2433,98 @@ void MujocoSystemInterface::set_initial_pose()
...
+  // Restore simulation time to preserve ROS clock continuity
+  mj_data_->time = saved_time;
+
+  // Run forward dynamics to update derived quantities
+  mj_forward(mj_model_, mj_data_);
+
+  // Copy to control data for reads - this ensures the physics loop uses the reset state
+  mj_copyData(mj_data_control_, mj_model_, mj_data_);
+
+  // Reset command interfaces to initial position commands
```
eholum-nasa, line 2473: Should we also reset the PIDs if joints are PID controlled? Would accumulated error come back to bite us?

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_system_interface.hpp
```diff
@@ -243,6 +244,27 @@ class MujocoSystemInterface : public hardware_interface::SystemInterface
...
 
+  /**
+   * @brief Resets the simulation state to the initial configuration.
+   *
+   * This method resets all joints to their initial positions and velocities,
+   * resets all free bodies (objects) to their spawned positions, and updates
+   * all actuator and joint command/state interfaces.
+   *
+   * @note This method assumes the sim_mutex_ is already held by the caller.
+   * @note Simulation time (mj_data_->time) is preserved to maintain ROS clock continuity.
```
eholum-nasa, line 255: Genius.

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: Not sure. Maybe we could. That's a good idea

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -2444,6 +2553,32 @@ void MujocoSystemInterface::PhysicsLoop()
       // lock the sim mutex during the update
       const std::unique_lock<std::recursive_mutex> lock(*sim_mutex_);
 
+      // Detect if a reset occurred via the UI (Sync() processed pending_.reset)
+      // This is detected by simulation time jumping backwards to near-zero
+      // The render loop's Sync() calls mj_resetData() which resets time to 0
+      if (mj_model_ && mj_data_ && prevSimTime > 0.1 && mj_data_->time < 0.01)
```
ndunkelb-nasa, line 2559: Oh wow that is an interesting way to do it! We will just have to stress test it and see if we can get it to break!

saikishor: Sure! I tried ;)

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: > just put several key frames in the mjcf file as suggested in https://github.com/ros-controls/mujoco_ros2_control/issues/74 (although I am not the biggest fan of this idea)

The reason is we have a lot of variants of the robot, and it is more complicated to maintain all the keyframes per configuration, so instead we build them with xacro at runtime, this makes it easier to handle things around.

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -2419,13 +2433,108 @@ void MujocoSystemInterface::set_initial_pose()
...
+      actuator.pos_pid->reset();
+    }
+    if (actuator.vel_pid)
+    {
+      actuator.vel_pid->reset();
+    }
+
+    if (actuator.actuator_type != ActuatorType::PASSIVE)
+    {
+      // Set command to initial position to maintain position control at reset position
```
ndunkelb-nasa, line 2492: @saikishor , does this work the way you mention here? As @eholum-nasa mentioned in another comment, this doesn't seem to be the case for me, at least with JTCs active. The previous command stays around and immediately gets sent there after the reset. 

You would know more about that then we would I think.

saikishor: Yes, you need to disable the controller and then reset the world (The test also tests the same). If not, the command interfaces are reset, but then after a controller update cycle, they get back to their old position. There is no easy way for the hardware to reset the controllers. I can think of a way to do this in the future.

ndunkelb-nasa: Got it, sounds like it makes sense to have a robot-specific service which just uses this as a part of the process!

saikishor: Yup! We can provide tools that do that part in the future.

saikishor: I've documented it in the recent push

mujoco_ros2_control/src/mujoco_system_interface.cpp
ndunkelb-nasa: Hmm, if I have a TF buffer up, it seems to be detecting the jump back in time using the reset button still. I am going to dig a bit and see if I can see why

```
[rviz2-1] [WARN] [1769443087.254211770] [rviz2]: Detected jump back in time. Resetting RViz.
[rviz2-1] [WARN] [1769443087.254245273] [tf2_buffer]: Detected jump back in time. Clearing TF buffer.
[rviz2-1] [WARN] [1769443087.254692401] [tf2_buffer]: Detected jump back in time. Clearing TF buffer.
```

saikishor: Ohh crap!

ndunkelb-nasa: Alright, I moved the tracking of the previous time to the end of the physics loop, and that resolved the issue for me. @saikishor , do you want to take another look and make sure that is ok?

saikishor: The change makes sense. I trust you ;)

saikishor: Works perfectly. Thanks for fixing it

PR #89, opened 2026-01-24
Support for starting from a specific declared keyframe 

saikishor:
This PR adds support for the feature requested in #74, I've tested with the test_robot, it works fine, but when you define a keyframe for a robot with more than 1  free joint, forsome reason,  mujoco itself doesn't like it and doesn't start any simulation either with mujoco_ros2_control or with it's native simulate option

### DIFF
--- a/mujoco_ros2_control/src/mujoco_system_interface.cpp
+++ b/mujoco_ros2_control/src/mujoco_system_interface.cpp
@@ -1757,6 +1757,7 @@ bool MujocoSystemInterface::register_mujoco_actuators()
       if (!override_mujoco_actuator_positions_)
       {
         RCLCPP_ERROR(get_logger(), "Failed to apply initial keyframe: '%s'", keyframe_name.c_str());
+        return false;
       }
     }
   }


mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -1746,6 +1746,21 @@ bool MujocoSystemInterface::register_mujoco_actuators()
...
+  if (!override_mujoco_actuator_positions_)
+  {
+    const std::string keyframe_name = get_hardware_parameter_or(get_hardware_info(), "initial_keyframe", "");
+    if (!keyframe_name.empty())
+    {
+      RCLCPP_INFO(get_logger(), "Applying initial keyframe: '%s'", keyframe_name.c_str());
+      override_mujoco_actuator_positions_ = apply_keyframe(keyframe_name);
+      if (!override_mujoco_actuator_positions_)
+      {
+        RCLCPP_ERROR(get_logger(), "Failed to apply initial keyframe: '%s'", keyframe_name.c_str());
```
eholum, line 1759: Since this happens on startup maybe it's worth failing over?

Recognizing that we don't for the start positions, but given that these should set the initial state of the simulation maybe startup should abort rather than coming up in an unexpected state?

Also ignore me if this was already talked about!

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -1746,6 +1746,21 @@ bool MujocoSystemInterface::register_mujoco_actuators()
...
+      }
+    }
+  }
+
   // Set initial values if they are set in the info, or from override start position file
   if (override_mujoco_actuator_positions_)
```
eholum, line 1766: Question for @ndunkelb-nasa , but how necessary is the start positions file if we have the keyframe in the model itself?

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_system_interface.hpp
```diff
@@ -179,6 +179,16 @@ class MujocoSystemInterface : public hardware_interface::SystemInterface
...
 
+  /**
+   * @brief Applies a keyframe to the simulation.
+   *
+   * This method sets the simulation state to the specified keyframe defined in the MuJoCo model.
+   * @param key_frame_name Name of the keyframe to apply.
+   *
+   * @return true if the keyframe was applied successfully, false otherwise.
+   */
+  bool apply_keyframe(const std::string& keyframe_name);
```
eholum, line 190: Not necessary now, but I wonder if it's possible to save the simulations start pose as a "keyframe" that we could come back to and reuse this for resetting the sim? It's so tidy.

eholum (commented):
So handy!

Not sure about the multiple free joints issue, do you have a sample robot? Just tried it here:

https://github.com/eholum/mujoco_ros2_control/commit/b9c18bb7cb8f8a3122b7615fc54a4e9d0bd8931e

And it seems to work with the two added free cubes?

Also sorry for this account I do not have access to my work computer at the moment.

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: Makes sense. I'm fine with this. I can make this change

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_system_interface.hpp
saikishor: Yesssss, that would be ideal. We need to decide to switch the current reset service with that one or add a new service. We can do that in a different PR

eholum-nasa: Agreed. I think there's definitely some consolidation with the start positions file, too!

mujoco_ros2_control/src/mujoco_system_interface.cpp
eholum-nasa: Eh, I think we should consolidate some things in a future PR and can deal with the workflow then.

saikishor: Sorry. Fixed it already.

eholum-nasa: Haha works for me!

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_system_interface.hpp
eholum-nasa: Talked with Nathan offline about this and the added benefit to the config startfile is that it doesn't have to live in the model... We'll keep thinking about it. Definitely something that could be solved if/when we move our sims to using MJCF conversion on the fly...

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: Do you mind approving it again ;)

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_system_interface.hpp
saikishor: Thanks for sharing

Issue #43, opened 2025-12-20
Restructure repository before release


christophfroehlich:
I suggest to restructure the repository before we release it, this is "best-practice" where we iterated to with https://github.com/ros-controls/gz_ros2_control and other packages in ros-controls org.

- Move the package in a mujoco_ros2_control subfolder:
  - we can then maintain a repo-based README for installation and development topics
  - maintain a mujoco_ros2_control/README with topics about using it
  - add a rosdoc2 setup to let it publish this README on https://docs.ros.org/en/rolling/p/mujoco_ros2_control
  - another advantage: catkin's changelogs will not include everything CI related from .github folder
- move the example in a new mujoco_ros2_control_examples package and let it be released by bloom (users can then install the example and are good to test it)
- add a new mujoco_ros2_control_tests package (which is configured to be ignored by bloom in the release repo), so we can reuse the example in the test without a circular dependency. There all integration tests should go (but I haven't seen any unittests, which means all tests could go there).

Let me know what you think! I'm happy to open PRs for it, I just want to discuss this before.

christophfroehlich:
Thanks to @saikishor and all the others, I think this is resolved now!

PR #90, opened 2026-01-26
Separate tests to mujoco_ros2_control_tests package

saikishor:
Create mujoco_ros2_control_tests package as a part of #43

### DIFF
--- a/mujoco_ros2_control/package.xml
+++ b/mujoco_ros2_control/package.xml
@@ -25,6 +25,8 @@
   <depend>std_srvs</depend>
   <depend>transmission_interface</depend>
 
+  <test_depend>ament_cmake_gtest</test_depend>
+
   <exec_depend>python3-venv</exec_depend>
   <exec_depend>python3-pip</exec_depend>
 


mujoco_ros2_control_tests/src/test_plugin.cpp
eholum, line 1: My one very nit request is that we leave the more "unit test"-like tests as part of the core package, just so that there are lightweight tests that live with the core source, instead of with the heavier integration tests!

I made the change here, https://github.com/eholum/mujoco_ros2_control/commit/dc37ea5e7e14a58d659ed06e505b8ae69577b133, to show what I mean.

eholum-nasa: How do you feel about this, @saikishor ? I generally think it's nice to keep unit level tests with what they're testing, and higher level functional/integration tests separate. Does that make sense? I don't need to block on this though.

saikishor: Ahh Sure. It makes sense. Let's do that. Can you push it to this branch?

saikishor: Sorry, I didn't see this earlier. Somehow, I missed it

eholum-nasa: No worries! I'll push it in one second...

eholum-nasa (approved):
Assuming I didn't break anything too badly LGTM!

Thanks again for handling all these maintenance tasks.

PR #91, opened 2026-01-29
Re-enable rolling workflow

saikishor:
This PR aims to re-enable the rolling workflow

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -20,6 +20,8 @@ ARG CPU_ARCH=x86_64
 ARG DEBIAN_FRONTEND=noninteractive
 
 # Configure base dependencies
+# NOTE: A dist-update is required due to recently broken rolling images.
+# https://openrobotics.zulipchat.com/#narrow/channel/526029-ROS-PMC-.28Restricted-posting.29/topic/2026-01-22.20Rolling.20Docker.20image.20out-of-date/near/569502119
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -23,6 +23,7 @@ ARG DEBIAN_FRONTEND=noninteractive
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     --mount=type=cache,target=/var/lib/apt,sharing=locked \
     apt-get update && \
+    apt-get dist-upgrade -y && \
```
eholum-nasa, line 28: Ooo, do we really need this for rolling to work? The major benefit to relying on the upstream ros images is we're building with the same image/package versions as them.

saikishor: I know :(

This is the fix for now to have the rolling workflows working. That's what they discussed in the Zulip link
https://openrobotics.zulipchat.com/#narrow/channel/526029-ROS-PMC-.28Restricted-posting.29/topic/2026-01-22.20Rolling.20Docker.20image.20out-of-date/near/569502119

docker/Dockerfile
eholum-nasa, line 22: Oooh, yeah maybe just worth adding a note so we can remove it if it's ever fixed. 


```suggestion
# Configure base dependencies
# NOTE: A dist-update is required due to recently broken rolling images.
# https://openrobotics.zulipchat.com/#narrow/channel/526029-ROS-PMC-.28Restricted-posting.29/topic/2026-01-22.20Rolling.20Docker.20image.20out-of-date/near/569502119
```

PR #93, opened 2026-02-03
Feature: Add --no-fuse arg to preserve body hierarchy

SergiMuac:
### Description
This PR improves the `make_mjcf_from_robot_description.py` script by adding a new argument, `--no-fuse`, and fixing compatibility issues when combining fixed-base preservation with free-floating joints.

### Motivation and Context
Currently, when converting a URDF to MJCF, the MuJoCo compiler defaults to `fusestatic="true"`. This causes bodies with no degrees of freedom relative to their parent to be merged for optimization.

**Issue 1: Loss of Body Hierarchy**
When `fusestatic="true"`, fixed joints are optimized away.
Consider the following structure:
```xml
<link name="base_link"> ... </link>
<link name="pelvis_base_link"> ... </link>

<joint name="base_pelvis_fixed_joint" type="fixed">
  <parent link="base_link" />
  <child link="pelvis_base_link" />
</joint>
```
In this scenario, users "lose" their `base_link` as a distinct entity in the MJCF. This makes it impossible to reference these bodies later for sensors, specific constraints, or frame tracking.

**Issue 2: Crash with Free Joint (`-f`)**
Previously, if a user tried to enforce a specific hierarchy while using the `-f` (free floating) flag, the script would crash with: `ValueError: free joint can only be used on top level`.
This occurred because the script injects a "virtual" link in the URDF to simulate a floating base. When fusion is disabled (`no-fuse`), this virtual link becomes a real body, pushing the robot's actual root down the hierarchy. This violates MuJoCo's requirement that free joints must exist on a top-level body (a direct child of `worldbody`).

### Changes
*   **Added `--no-fuse` argument:**
    *   Passes `fusestatic="false"` to the MuJoCo compiler options.
    *   This ensures that static bodies in the URDF remain as separate `<body/>` elements in the MJCF, rather than being merged into the world or their parents.

*   **Improved Free Joint Logic (`-f`):**
    *   When both `-f` and `--no-fuse` are provided, the script now **skips** adding the temporary "virtual_base" link in the URDF (which was previously required to trick MuJoCo into creating a free joint).
    *   Instead, the script detects the actual root link of the robot and injects the `<freejoint/>` tag directly into that body in the generated MJCF, ensuring a valid top-level kinematic chain.


mujoco_ros2_control/scripts/make_mjcf_from_robot_description.py
```diff
@@ -1607,6 +1619,11 @@ def main(args=None):
         action="store_true",
         help="Adds a free joint before the root link of the robot in the urdf before conversion",
     )
+    parser.add_argument(
+        "--no-fuse",
+        action="store_true",
+        help="Prevents MuJoCo from merging static bodies (useful for fixed-base robots to preserve the root link)",
+    )
```
saikishor, line 1627:
```suggestion
    parser.add_argument(
        "--fuse",
        action="store_true",
        help="Prevents MuJoCo from merging static bodies (useful for fixed-base robots to preserve the root link)",
        action=argparse.BooleanOptionalAction,
    )
```

SergiMuac: Good! Done!

ndunkelb-nasa:
Out of curiosity, what was the driving use case for needing bodies themselves? We found that most things required **sites**, but didn't really use the **bodies**. For that reason, we duplicated every frame from the URDF as a site to use for sensors, etc. Although I would definitely like to have the body frames available in mujoco for troubleshooting.

Regardless, if i had known that `fusestatic` was an available option, I would have done it in the first place! That is awesome!

saikishor:
> Out of curiosity, what was the driving use case for needing bodies themselves? We found that most things required **sites**, but didn't really use the **bodies**. For that reason, we duplicated every frame from the URDF as a site to use for sensors, etc. Although I would definitely like to have the body frames available in mujoco for troubleshooting.
> 
> Regardless, if i had known that `fusestatic` was an available option, I would have done it in the first place! That is awesome!

The reason to do so is that we usually have a simulation of our robots with a fixed base, and this will allow us to run some trajectories or do some exploration without having to worry about the floating base. That's the only reason for now hahah

SergiMuac:
> Out of curiosity, what was the driving use case for needing bodies themselves? We found that most things required **sites**, but didn't really use the **bodies**. For that reason, we duplicated every frame from the URDF as a site to use for sensors, etc. Although I would definitely like to have the body frames available in mujoco for troubleshooting.
> 
> Regardless, if i had known that `fusestatic` was an available option, I would have done it in the first place! That is awesome!

Most of our robots use the base_link as the root. Typically, when the free-floating joint is added (via the -f arg), the base_link is preserved and only the internal fixed joints are merged.

​However, when we spawn the robot in a fixed position (without -f), all fixed joints are merged. This causes us to lose the base_link body and several others. While sensors aren't affected because they attach to sites, the <exclude> contact tags are a problem. These tags must reference specific bodies; because the body names change depending on whether -f is used, we end up with tags referencing non-existent bodies.

​By using fusestatic to keep the body hierarchy consistent, we solve this issue permanently.

PR #95, opened 2026-02-03
Add ResetWorld service to reset to a specific defined keyframe

saikishor:
As discussed, I've added a new service message type that would allow the users to switch to certain keyframe. If no keyframe is defined, it resets to the initial position


saikishor:
Thank you @eholum-nasa. I tried to handle Pixi CI, but couldn't solve it easily.

As per your instructions, I'll go ahead and merhe it 🫡

PR #99, opened 2026-02-06
Containerize pixi in CI

eholum-nasa:
Due to some combination of recent releases (pixi, GitHub action runner, and ubuntu images), the pixi workflow on bare GH action runners has become wildly unstable. I haven't found a root cause (nor has anyone else), so I say to isolate that workflow we just run in a bare ubuntu image.

This change does that, and removes the previously added hacks that made the integration tests work.


saikishor:
> Sorry to waste your time here, @saikishor! I'm just going to double check caching and then will merge it.

No no, don't worry. Yours is much better fix and better in long run

PR #101, opened 2026-02-07
feat: Make MujocoSystemInterface inheritable

Juliaj:
**Summary** 

Based on [discussion ](https://github.com/ros-controls/mujoco_ros2_control/pull/69#issuecomment-3860796603), the recommendation for adding contact detection support is to create a HW interface inheriting from `MujocoSystemInterface`. 

This PR enables `MujocoSystemInterface` to be properly inherited by external packages (e.g., `DuckMiniMujocoSystemInterface` in ros2_control_demos [example_18](https://github.com/Juliaj/ros2_control_demos/tree/onnx_demo_open_duck_mini/example_18/hardware)). The changes include moving `get_logger()` to the protected section for derived class access, and adding proper CMake target export setup so that derived classes can link against the library. 

**Main Changes**
- Inheritance Support:  moved `get_logger()` method from private to protected section in `mujoco_system_interface.hpp` to allow derived classes to access the logger
- Target Export Setup, the package now exports its CMake targets so that consuming packages can use `find_package(mujoco_ros2_control)` and link against `mujoco_ros2_control::mujoco_ros2_control` without manual workarounds
 - Internal dependencies like `libsimulate` and `lodepng` are also exported so they're automatically available to consumers
 - The library installation location was moved to the standard `lib/` directory to match ROS2 conventions and enable automatic RPATH handling in test executables
 - Conditional Export of mujoco_bundled. When MuJoCo is bundled locally (rather than system-installed), CMake needs a wrapper target to properly handle the difference between where the library lives during build versus after installation.

@eholum-nasa review please. 

Testing:
- Change tested via ros2_control_demos [example_18](https://github.com/Juliaj/ros2_control_demos/tree/onnx_demo_open_duck_mini/example_18/hardware)

### DIFF
--- a/mujoco_ros2_control/CMakeLists.txt
+++ b/mujoco_ros2_control/CMakeLists.txt
@@ -5,30 +5,20 @@ project(mujoco_ros2_control)
 find_package(ros2_control_cmake REQUIRED)
 
 # find dependencies
-# ROS2 package dependencies (matching package.xml)
-set(THIS_PACKAGE_INCLUDE_DEPENDS
-  backward_ros
-  control_toolbox
-  controller_manager
-  hardware_interface
-  nav_msgs
-  pluginlib
-  rclcpp
-  rclcpp_lifecycle
-  sensor_msgs
-  mujoco_ros2_control_msgs
-  transmission_interface
-)
-
 find_package(ament_cmake REQUIRED)
-foreach(Dependency IN ITEMS ${THIS_PACKAGE_INCLUDE_DEPENDS})
-  find_package(${Dependency} REQUIRED)
-endforeach()
-
-# Non-ROS2 dependencies
+find_package(controller_manager REQUIRED)
 find_package(Eigen3 REQUIRED)
 find_package(glfw3 REQUIRED)
+find_package(control_toolbox REQUIRED)
+find_package(hardware_interface REQUIRED)
+find_package(nav_msgs REQUIRED)
+find_package(pluginlib REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(rclcpp_lifecycle REQUIRED)
+find_package(mujoco_ros2_control_msgs REQUIRED)
 find_package(Threads REQUIRED)
+find_package(transmission_interface REQUIRED)
+find_package(backward_ros REQUIRED)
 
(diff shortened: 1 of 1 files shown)


mujoco_ros2_control/CMakeLists.txt
```diff
@@ -77,13 +102,43 @@ else()
...
   )
+  # Install mujoco headers so they're available when exporting targets
+  install(
+    DIRECTORY ${MUJOCO_INCLUDE_DIR}/
+    DESTINATION include/mujoco
+  )
+  install(
+    DIRECTORY ${MUJOCO_SIMULATE_DIR}/
+    DESTINATION include/mujoco_simulate
+  )
```
saikishor: I would avoid doing this as it is already taken care by mujoco_vendor

mujoco_ros2_control/CMakeLists.txt
```diff
@@ -5,40 +5,65 @@ project(mujoco_ros2_control)
...
   if(NOT MUJOCO_LIB)
-    message(FATAL_ERROR "Failed to find binary in $ENV{MUJOCO_DIR}")
+    message(FATAL_ERROR "Failed to find binary in ${MUJOCO_DIR}/lib")
   endif()
-  set(MUJOCO_INCLUDE_DIR $ENV{MUJOCO_DIR}/include)
-  set(MUJOCO_SIMULATE_DIR $ENV{MUJOCO_DIR}/simulate)
+  set(MUJOCO_INCLUDE_DIR ${MUJOCO_DIR}/include)
+  set(MUJOCO_SIMULATE_DIR ${MUJOCO_DIR}/simulate)
+  set(MUJOCO_NEEDS_INSTALL TRUE)
+  # Set install paths for use in generator expressions ($<INSTALL_INTERFACE:...>)
```
saikishor: Can we avoid it somehow?

saikishor:
> We also have odd workflows that are kind of borked. But I wouldn't mind cleaning this up a bit if you guys are ok with it?

@eholum-nasa The thing is it is already released in all distros, but not for pixi. 

What I've in mind is, we can merge the vendor right before the release and then after the tagging and releasing it, we simply revert back for pixi or find a way to make it work with pixi

PR #105, opened 2026-02-11
Cleanup the non-vendor mujoco install

eholum-nasa:
Picking up from @Juliaj's PR in #101, and pulling in the latest vendor package changes from main.

The main changes here are to break out the "fallback" logic if the mujoco vendor package isn't available to a separate cmake module, just to clean things up a bit. The separate module exports the same targets/variables as the vendor package, so for the rest of the build it _should_ be the same. I also complete removed the environment variable install method. It's been annoying and I don't think it's worth supporting now that the vendor package is available!

Then we have to make sure that a minimal set of dependencies (basically just mujoco things) are exposed so external consumers can get access to those functions (the model/data getters in particular). This should ensure that the exports are usable, and enable extremely dedicated developers to link against the hardware interface.

### DIFF
--- a/mujoco_ros2_control/CMakeLists.txt
+++ b/mujoco_ros2_control/CMakeLists.txt
@@ -136,10 +136,10 @@ PUBLIC
   transmission_interface::transmission_interface
   pluginlib::pluginlib
   ${MUJOCO_LIB}
-PRIVATE
-  mujoco::libsimulate
   control_toolbox::control_toolbox
   Eigen3::Eigen
+PRIVATE
+  mujoco::libsimulate
   Threads::Threads
   lodepng
   glfw
--- a/mujoco_ros2_control/test/CMakeLists.txt
+++ b/mujoco_ros2_control/test/CMakeLists.txt
@@ -1,6 +1,4 @@
 find_package(ament_cmake_gtest REQUIRED)
-find_package(Eigen3 REQUIRED)
-find_package(control_toolbox REQUIRED)
 
 ament_add_gtest(test_plugin test_plugin.cpp)
 target_link_libraries(test_plugin
@@ -13,6 +11,4 @@ target_link_libraries(test_headless_init
   mujoco_ros2_control
   rclcpp::rclcpp
   hardware_interface::hardware_interface
-  Eigen3::Eigen
-  control_toolbox::control_toolbox
 )


mujoco_ros2_control/CMakeLists.txt
```diff
@@ -20,90 +20,25 @@ find_package(Threads REQUIRED)
...
-  set(MUJOCO_INCLUDE_DIR ${MUJOCO_DIR}/include)
-  set(MUJOCO_SIMULATE_DIR ${MUJOCO_DIR}/simulate)
-
-  message(STATUS "Installing ${MUJOCO_LIB} to the install directory...")
-  install(
-    DIRECTORY ${MUJOCO_DIR}/lib/
-    DESTINATION lib
-    FILES_MATCHING PATTERN "libmujoco.so*"
-  )
+  # Otherwise we fallback to downloading the required version on the fly, but this requires
```
saikishor, line 36:
```suggestion
  set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
  # Otherwise we fallback to downloading the required version on the fly, but this requires
```

mujoco_ros2_control/CMakeLists.txt
```diff
@@ -20,90 +20,25 @@ find_package(Threads REQUIRED)
 find_package(transmission_interface REQUIRED)
 find_package(backward_ros REQUIRED)
 
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+
```
saikishor:
```suggestion
```

mujoco_ros2_control/test/test_headless_init.cpp
eholum-nasa, line 1: Throwing this in just to check the segfault in CI.

eholum-nasa: https://github.com/ros-controls/mujoco_ros2_control/actions/runs/21911522194/job/63266694143?pr=105#step:6:60

Yay!

saikishor: LOL, it did

mujoco_ros2_control/CMakeLists.txt
```diff
@@ -183,20 +115,21 @@ set_target_properties(mujoco_ros2_control PROPERTIES
...
-  control_toolbox::control_toolbox
   ${nav_msgs_TARGETS}
   ${sensor_msgs_TARGETS}
   ${mujoco_ros2_control_msgs_TARGETS}
   transmission_interface::transmission_interface
+  pluginlib::pluginlib
+  ${MUJOCO_LIB}
+PRIVATE
+  mujoco::libsimulate
+  controller_manager::controller_manager
```
saikishor:
```suggestion
```

mujoco_ros2_control/CMakeLists.txt
```diff
@@ -183,20 +115,21 @@ set_target_properties(mujoco_ros2_control PROPERTIES
...
+  hardware_interface::mock_components
```
saikishor:
```suggestion
```

mujoco_ros2_control/test/CMakeLists.txt
```diff
@@ -1,7 +1,18 @@
 find_package(ament_cmake_gtest REQUIRED)
+find_package(Eigen3 REQUIRED)
+find_package(control_toolbox REQUIRED)
```
saikishor:
```suggestion
```

mujoco_ros2_control/test/CMakeLists.txt
```diff
@@ -1,7 +1,18 @@
...
   rclcpp::rclcpp
 )
+
+ament_add_gtest(test_headless_init test_headless_init.cpp)
+target_link_libraries(test_headless_init
+  mujoco_ros2_control
+  rclcpp::rclcpp
+  hardware_interface::hardware_interface
+  Eigen3::Eigen
+  control_toolbox::control_toolbox
```
saikishor:
```suggestion
```
These are already linked to the parent library, do we need them still?

eholum-nasa: Ah, nope! But we should include them in the library's linking. Adjusted...

eholum-nasa:
> LGTM

Might as well let CI pass a few more times!

PR #107, opened 2026-02-11
Support camera rendering for headless environments using EGL

### LINKED ISSUE #27: Support camera rendering for headless environments
As noted in https://github.com/ros-controls/mujoco_ros2_simulation/pull/24, glfw will not initialize without a display. If possible, it would be nice to implement some kind of fallback or alternative to give Mujoco a valid OpenGL context with which to render camera images. In the absence of anything... I wonder if it would be possible to fallback to cpu rendering?

saikishor:
This adds support for camera rendering while being headless 

Should fix #27

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,6 +60,8 @@ jobs:
       - name: Run tests
         run: |
           docker run --rm \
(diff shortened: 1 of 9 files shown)


eholum-nasa:
Picked up in #197.

PR #108, opened 2026-02-12
Make register_sensors virtual for derived class to override

Juliaj:
**Changes**
- Make register_sensors() virtual and move it to protected so derived classes can override it.
- Treat unknown sensor types as debug instead of error so derived classes can add custom types (e.g. contact) without noisy logs.

**Testing**

[Example_18](https://github.com/Juliaj/ros2_control_demos/blob/fe8fac429a6a54ebc36ffb2619aed2aaad8f0876/example_18/hardware/duck_mini_mujoco_system_interface.cpp#L47-L51)

### DIFF
--- a/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_system_interface.hpp
+++ b/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_system_interface.hpp
@@ -153,45 +153,6 @@ public:
 protected:
   rclcpp::Logger get_logger() const;
 
-private:
-  /**
-   * @brief Loads actuator information from MuJoCo model into the SystemInterface.
-   *
-   * This function reads the actuator definitions from the MuJoCo model and initializes their corresponding
-   * state and command information of the actuator handles.
-   */
-  bool register_mujoco_actuators();
-
-  /**
-   * @brief Loads actuator information into the HW interface.
-   *
-   * Will pull joint/actuator information from the provided HardwareInfo, and map it to the appropriate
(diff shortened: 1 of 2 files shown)


mujoco_ros2_control/include/mujoco_ros2_control/mujoco_system_interface.hpp
```diff
@@ -242,7 +203,46 @@ class MujocoSystemInterface : public hardware_interface::SystemInterface
    *    <state_interface name="linear_acceleration.z"/>
    *  </sensor>
    */
-  void register_sensors(const hardware_interface::HardwareInfo& info);
+  virtual void register_sensors(const hardware_interface::HardwareInfo& info);
```
eholum-nasa, line 206: FWIW I _think_ we could get away with just exposing the base hardware interface + the model getter/setters in the interface. Rather than overriding this as you [do here](https://github.com/Juliaj/ros2_control_demos/blob/fe8fac429a6a54ebc36ffb2619aed2aaad8f0876/example_18/hardware/duck_mini_mujoco_system_interface.cpp#L47-L51), could you just add it to your overridden `on_init`?

That way we can keep the interface as close to the base hardware_interface as possible, and this could stay private and unique to the base class.

saikishor:
@Juliaj For now, it is better to expose the minimal API, as that keeps additional burden to maintain API of mujoco_ros2_control. For the initial release, I want to keep things simple. I hope you don't mind. We can get back to this soon and also to understand what we can integrate in near future.

PR #110, opened 2026-02-13
Update README.md before releasing

saikishor:
MInor updates to the readme


mujoco_ros2_control/README.md
eholum-nasa, line 1: Maybe worth first `git mv`ing this so that the history moves with it?

README.md
```diff
@@ -1,456 +1,96 @@
...
+- Example demos showing basic control, PID and transmission setups
 
-## URDF Model Conversion
+### Requirements
 
-Mujoco does not support the full feature set of xacro/URDFs in the ROS 2 ecosystem.
-As such, users are required to convert any existing robot description files to an MJCF format.
-This includes adding actuators, sensors, and cameras as needed to the MJCF XML.
+- ROS 2
+- MuJoCo 3.x (the CMake scripts try to use `mujoco_vendor`, if available)
```
eholum-nasa: Maybe not worth mentioning since it's handled internally now and users shouldn't be trying to install mujoco manually?

README.md
```diff
@@ -1,456 +1,96 @@
...
+There are two common ways to get running: build from source (recommended)
+or install prebuilt binaries (if available for your distribution).
 
-### Plugin
+- Build from source (recommended)
 
-This application is shipped as a ros2_control hardware interface, and can be configured as such.
-Just specify the plugin and point to a valid MJCF on launch:
+  1. Install MuJoCo and ensure it is available on your system. Optionally
+     install `mujoco_vendor` so CMake can find prebuilt MuJoCo binaries.
```
eholum-nasa: Yeah just because there is no longer an option to specify a local install.

```suggestion
  1. Install required dependencies manually or from rosdep, including the `mujoco_vendor` package, if available. Otherwise MuJoCo will will be downloaded at build time.
```

mujoco_ros2_control/README.md
saikishor: I did with mv and then did that with the first commit. Weirdly it messed up everything

README.md
saikishor: Copy that

mujoco_ros2_control/README.md
eholum-nasa: Oh man that's annoying. But /shrug! onward.

PR #111, opened 2026-02-13
Cleanup of Docs and Dev Guides

eholum-nasa:
Did a quick pass through and sanity checked some of the docs that hadn't been touched in a long while. There were some broken instructions and things have been updated/moved, so I just cleaned some of those things out, and fixed a few things in the Docker dev workflow.

### DIFF
--- a/mujoco_ros2_control/package.xml
+++ b/mujoco_ros2_control/package.xml
@@ -6,6 +6,12 @@
   <description>ros2_control wrapper for the Mujoco Simulate application</description>
   <maintainer email="nathan.b.dunkelberger@nasa.gov">Nathan Dunkelberger</maintainer>
   <maintainer email="erik.holum@nasa.gov">Erik Holum</maintainer>
+
+  <maintainer email="bence.magyar.robotics@gmail.com">Bence Magyar</maintainer>
+  <maintainer email="denis@stoglrobotics.de">Denis Štogl</maintainer>
+  <maintainer email="christoph.froehlich@ait.ac.at">Christoph Froehlich</maintainer>
+  <maintainer email="sai.kishor@pal-robotics.com">Sai Kishor Kothakota</maintainer>
+
   <license>Apache License, Version 2.0</license>
 
   <buildtool_depend>ament_cmake</buildtool_depend>


mujoco_ros2_control/docs/TOOLS.md
```diff
@@ -57,54 +59,35 @@ options:
 
 A sample URDF and inputs file are provided in [test_robot.urdf](../../mujoco_ros2_control_tests/test/test_resources/test_robot.urdf) and [test_inputs.xml](../../mujoco_ros2_control_tests/test/test_resources/test_inputs.xml).
 
-To convert the URDF, run the following from the repo root
```
ndunkelb-nasa, line 60: I actually think this is useful to have from repo root. Otherwise, we could toss in `$(ros2 pkg prefix mujoco_ros2_control_demos)/share/mujoco_ros2_control_demos` but that is super annoying...

mujoco_ros2_control/docs/TOOLS.md
```diff
@@ -57,54 +59,35 @@ options:
...
-
 ## Notes
 
 **NOTE** This has some heave non-ROS dependencies that could probably be cleaned up:
 
+* MuJoCo Python API
 * trimesh - Python library for loading and using triangular meshes.
 * obj2mjcf - A tool for converting Wavefront OBJ files to multiple MuJoCo meshes grouped by material.
-* MuJoCo Python API
 * xml.etree (not sure if this is already available)
```
ndunkelb-nasa: I actually think the only thing that uses etree is the find missing inertias... But I also think that doesn't really work, so maybe we should put a disclaimer or just take it away tbh

mujoco_ros2_control/docs/TOOLS.md
```diff
@@ -57,54 +59,35 @@ options:
...
-"${MUJOCO_DIR}"/bin/simulate /tmp/clr_output/scene.xml
-```
-
 ## Notes
 
 **NOTE** This has some heave non-ROS dependencies that could probably be cleaned up:
 
+* MuJoCo Python API
 * trimesh - Python library for loading and using triangular meshes.
 * obj2mjcf - A tool for converting Wavefront OBJ files to multiple MuJoCo meshes grouped by material.
```
ndunkelb-nasa, line 91: Is there a reason we can't or don't want to rosdep these? I guess just to not include python dependencies for a tool that is mainly just a shared library?

docker/Dockerfile
```diff
@@ -41,6 +37,11 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
       xterm \
       vim
 
+# Install MuJoCo specific pip dependencies, ignoring the conflict in kilted and rolling images
+ENV PIP_BREAK_SYSTEM_PACKAGES=1
+RUN pip install --ignore-installed typing-extensions && \
+    pip install mujoco obj2mjcf trimesh pycollada
+
```
saikishor: @eholum-nasa do we need this? I believe the bash script that we have already sets them up at execution time right? 

I believe that's how it was working till now. Isn't it?

mujoco_ros2_control/README.md
```diff
@@ -421,36 +407,11 @@ regardless of whether their interfaces are exposed via `ros2_control`.
...
-ros2 launch mujoco_ros2_control_tests test_robot.launch.py use_pid:=true
-
-# Launch an rviz2 window with the provided configuration
-rviz2 -d $(ros2 pkg prefix --share mujoco_ros2_control_tests)/config/test_robot.rviz
-```
-
-From there, command joints to move with,
-
-```bash
-ros2 topic pub /position_controller/commands std_msgs/msg/Float64MultiArray "data: [-0.25, 0.75]" --once
```
ndunkelb-nasa, line 446: Could be useful to toss this into the main page where the launch is included to show people that the test robot can actually do a thing

mujoco_ros2_control/docs/TOOLS.md
eholum-nasa: Oh lol yeah will revert. I had the prefix in there but it was too ugly and I forgot to change this back...

mujoco_ros2_control/docs/TOOLS.md
eholum-nasa: Yeah except not without breaking system packages... That's why the shell version wraps up installing these things in a virtualenv.

mujoco_ros2_control/docs/TOOLS.md
eholum-nasa: Yeah I'll just remove it... these "dependencies" are handled for people anyway with the conversion shell script.

mujoco_ros2_control/docs/TOOLS.md
saikishor: Yes, you can buy you have to accept to break system packages, that's why we should recommend to use the bash script that creates the venv in ROS_HOME and installs them over there.

mujoco_ros2_control/README.md
eholum-nasa: On the top level README? I just removed because the demo docs cover way more things, e.g.:

https://github.com/ros-controls/mujoco_ros2_control/tree/main/mujoco_ros2_control_demos#controlling-the-robot

mujoco_ros2_control/docs/TOOLS.md
eholum-nasa: Oh yeah in regards to your other comment let me update the demo docs to just use the shell script...

mujoco_ros2_control/README.md
ndunkelb-nasa: Ahh, all good, I was just looking at the changes, but didn't go into those docs.

docker/Dockerfile
eholum-nasa: Yep! Sorry just had to update the tools docs to use the shell script and now I can remove this... derp.

eholum-nasa: Ok it #JustWorks!

ndunkelb-nasa (approved):

PR #116, opened 2026-02-15
Use generate parameter library to configure the simulator

saikishor:
This PR  aims to also provide an alternative way to setup the simulator by parsing paramters via ROS parameters instead of the making changes to the URDF

### DIFF
--- a/mujoco_ros2_control/src/mujoco_system_interface.cpp
+++ b/mujoco_ros2_control/src/mujoco_system_interface.cpp
@@ -965,7 +965,7 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
     });
   }
 
-  if (sim_ready_future.wait_for(2s) == std::future_status::timeout)
+  if (sim_ready_future.wait_for(10s) == std::future_status::timeout)
   {
     RCLCPP_FATAL(get_logger(), "Timed out waiting to start simulation rendering!");
     return hardware_interface::CallbackReturn::ERROR;


ndunkelb-nasa:
I do have a couple of comments about how this is set up, and feel like it could be made a little bit less confusing. There are currently so many options that you can set, and I can see it being very easy to end up with a config where you think you have some sets of parameters, but you really have a different set active (which is why I have been specific about info/warning/error messages).

There are some params that are supported on both versions, but with different names, for example `camera_publish_rate` and `publish_rate.camera`, but some parameters are the same, like `mujoco_model`. I think that could lead users to think that they are the same in both sections and then end up accidentally setting a variable that doesn't exist. I think it would be good to keep them the same across the two options if possible.

There are also some options that are only available on one or the other. I think that if some options are available on both, we should make all options available on both so that we don't end up with confusion.

I think that potentially, something to help clean this up would be if we had some kind of helper function to process params in a standardized manner. I know it would be hard to handle all of the variety of cases, but it is very easy to get lost in the implementation in its current form. Maybe as pseudocode something like
```cpp
// instead of 
  const std::string mujoco_model_topic =
      !params_->mujoco_model_topic.empty() ?
          params_->mujoco_model_topic :
          get_hardware_parameter_or(get_hardware_info(), "mujoco_model_topic", "/mujoco_robot_description");
          
  sim_speed_factor_ = params_->sim_speed_factor > 0 ?
                          params_->sim_speed_factor :
                          std::stod(get_hardware_parameter(get_hardware_info(), "sim_speed_factor").value_or("-1"));

// generic function to get params from either generate param library or from urdf
type get_param(type param, function eval_if_param_exists, std::string hardware_param_name, type default){
  generate_param_exists = eval_if_param_exists(param)
  hardware_param_exists = hardware_parameter(hardware_param_name)
  if (generate_param_exists && hardware_param_exists){
    LOG("both parameters passed for {hardware_param_name}, using the parameter from the parameters file")
  }
  if (generate_param_exists){ 
    return param;
  }
  else if (hardware_param_exists){
    return get_hardware_parameter(hardware_param_name);
  }
  else {
    return default;
  }
}

auto is_not_empty [](const std::string& s) -> bool {
    return !s.empty(); 
}
auto is_positive [](const double d) -> bool {
    return d > 0; 
}
mujoco_model_topic = get_param(params_->mujoco_model_topic, is_not_empty, "mujoco_model_topic", "/mujoco_robot_description");
initial_keyframe = get_param(params_->initial_keyframe, is_not_empty, "initial_keyframe", "");
override_start_positions_file = get_param(params_->override_start_positions_file, is_not_empty, "override_start_positions_file", "");

sim_speed_factor = get_param(params_->sim_speed_factor, is_positive, "sim_speed_factor", "-1");
```

PR #118, opened 2026-02-17
Add proper logging for the mimic joints

saikishor:
**Current behaviour:**

```
Failed to find actuator for joint : gripper_outer_finger_right_joint. This joint will be treated as a passive joint.
Joint : gripper_outer_finger_right_joint is a passive joint
```

**With the proposed changes:**
```
[ros2_control_node-21] [INFO] [1771335814.029415094] [MujocoSystemInterface]: MuJoCo joint 'gripper_right_outer_finger_right_joint' is a mimic joint and has no associated actuator.
[ros2_control_node-21] [INFO] [1771329892.733762642] [MujocoSystemInterface]: Joint : 'gripper_right_outer_finger_right_joint' is a mimic joint mimicking joint 'gripper_right_finger_joint' with multiplier '-8.28'
```

### DIFF
--- a/mujoco_ros2_control/src/mujoco_system_interface.cpp
+++ b/mujoco_ros2_control/src/mujoco_system_interface.cpp
@@ -1882,7 +1882,7 @@ void MujocoSystemInterface::register_urdf_joints(const hardware_interface::Hardw
                      });
     const bool actuator_exists = actuator_it != mujoco_actuator_data_.end();
     // This isn't a failure the joint just won't be controllable
-    RCLCPP_WARN_EXPRESSION(get_logger(), !actuator_exists && !joint_data.is_mimic,
+    RCLCPP_INFO_EXPRESSION(get_logger(), !actuator_exists && !joint_data.is_mimic,
                            "Failed to find actuator for joint : %s. This joint will be treated as a passive joint.",
                            joint.name.c_str());
     RCLCPP_INFO_EXPRESSION(get_logger(), joint.command_interfaces.empty() && !joint_data.is_mimic,


mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -1867,8 +1858,47 @@ void MujocoSystemInterface::register_urdf_joints(const hardware_interface::Hardw
       {
         joint_data.mimic_multiplier = 1.0;
       }
+      RCLCPP_INFO_EXPRESSION(get_logger(), joint_data.is_mimic,
+                             "Joint : '%s' is a mimic joint mimicking joint '%s' with multiplier '%.2f'",
+                             joint.name.c_str(), hardware_info.joints.at(joint_data.mimicked_joint_index).name.c_str(),
+                             joint_data.mimic_multiplier);
+      if (!joint.command_interfaces.empty())
+      {
+        RCLCPP_WARN(
```
ndunkelb-nasa: I feel like this should be an error instead of a warning. I think if someone is expecting something to happen, and that thing is not happening, it should be an error instead of a warning.

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -1867,8 +1858,47 @@ void MujocoSystemInterface::register_urdf_joints(const hardware_interface::Hardw
...
+    const auto actuator_it =
+        std::find_if(mujoco_actuator_data_.begin(), mujoco_actuator_data_.end(),
+                     [&actuator_name, this](const MuJoCoActuatorData& actuator) {
+                       return (actuator.actuator_type != ActuatorType::PASSIVE) &&
+                              ((mj_id2name(mj_model_, mjOBJ_ACTUATOR, actuator.mj_actuator_id) == actuator_name) ||
+                               (actuator.joint_name == actuator_name));
+                     });
+    const bool actuator_exists = actuator_it != mujoco_actuator_data_.end();
+    // This isn't a failure the joint just won't be controllable
+    RCLCPP_WARN_EXPRESSION(get_logger(), !actuator_exists && !joint_data.is_mimic,
```
ndunkelb-nasa: I'm a bit confused by this expression. This shouldn't necessarily be a warning, unless there are command interfaces defined? Or am I misinterpreting this.

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -1867,8 +1858,47 @@ void MujocoSystemInterface::register_urdf_joints(const hardware_interface::Hardw
...
+                           "Failed to find actuator for joint : %s. This joint will be treated as a passive joint.",
+                           joint.name.c_str());
+    RCLCPP_INFO_EXPRESSION(get_logger(), joint.command_interfaces.empty() && !joint_data.is_mimic,
+                           "Joint : %s is a passive joint", joint.name.c_str());
+    if (!joint.command_interfaces.empty() && !actuator_exists)
+    {
+      RCLCPP_WARN(get_logger(),
```
ndunkelb-nasa: Again, I feel like this could be an error

saikishor: You are totally right!

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: You are totally right!

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: We print this message when we don't find any actuator for the joint and also if the joint is not mimic.

Are you saying we should only print when there are also command interfaces defined?

ndunkelb-nasa: Yeah, I think so? I think in this case, it could just be an info print if we wanted it here. Warning to implies that you might have done something wrong

saikishor: Sure. I can chnage it

PR #119, opened 2026-02-19
Improve testing for URDF<->MJCF tooling

saikishor:
I moved the methods that we have in the `make_mjcf_from_robot_description.py` script, and then made it a library and then I added tests for each and every method. I could see some improvements and edge cases to handle, but in this I only moved the methods and added tests for them

### DIFF
--- a/mujoco_ros2_control/CHANGELOG.rst
+++ b/mujoco_ros2_control/CHANGELOG.rst
@@ -1,0 +1,25 @@
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Changelog for package mujoco_ros2_control
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+0.0.1 (2026-02-24)
+------------------
+* Update documentation of URDF <-> MJCF tool (`#117 <https://github.com/ros-controls/mujoco_ros2_control/issues/117>`_)
+* Add proper logging for the mimic joints (`#118 <https://github.com/ros-controls/mujoco_ros2_control/issues/118>`_)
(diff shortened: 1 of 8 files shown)


mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
eholum-nasa, line 1: Ah nice. Was this just a straight cut/paste of all these utils functions?

(assuming yes and not really reading it....)

mujoco_ros2_control/tests/test_urdf_to_mujoco_utils.py
```diff
@@ -0,0 +1,1779 @@
...
+        self.assertEqual(result_dom.getElementsByTagName("camera")[0].getAttribute("quat"), "0.0 1.0 0.0 0.0")
+        self.assertEqual(
+            len(result_dom.getElementsByTagName("worldbody")[0].attributes), 0
+        )  # worldbody should have no attributes
+        self.assertEqual(
+            len(result_dom.getElementsByTagName("mujoco")[0].attributes), 0
+        )  # mujoco should have no attributes
+        self.assertEqual(
+            len(result_dom.getElementsByTagName("body")), 1
+        )  # there should be two body elements (base and camera body)
```
eholum-nasa: I think cameras don't get added as a body?


```suggestion
            len(result_dom.getElementsByTagName("body")), 1
        )  # there should still be a single body element (base)
```

mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
saikishor: Yes it is, I saw some functions could need some change, but I prefer to not touch them and add the tests as it is and then fix in the next PR. So, yes it is pure copy paste

PR #123, opened 2026-02-26
Changes necessary to also work for MuJoCo 3.5.0

saikishor:
I added some minor changes to the codebase to also have it working for MuJoCo 3.5.0 and these are also backward compatible

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,24 @@
+name: Coverage Build - Jazzy
+on:
+  workflow_dispatch:
+  pull_request: &event
+    branches:
+      - main
+    paths:
+      - '**.hpp'
+      - '**.h'
+      - '**.cpp'
+      - '**.py'
+      - '**.yaml'
+      - '.github/workflows/coverage.yml'
+      - '**/package.xml'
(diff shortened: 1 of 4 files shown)


mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -459,8 +459,8 @@ mjModel* loadModelFromTopic(rclcpp::Node::SharedPtr node, const std::string& top
       mj_deleteSpec(spec);
     }
     mj_deleteSpec(spec);
-    RCLCPP_INFO(node->get_logger(), "Model body count: %d", mnew->nbody);
-    RCLCPP_INFO(node->get_logger(), "Model geom count: %d", mnew->ngeom);
+    RCLCPP_INFO(node->get_logger(), "Model body count: %ld", static_cast<long>(mnew->nbody));
```
eholum-nasa, line 462: Were these actual type changes or was this just addressing warnings or something?

saikishor: It was failing as we use ros2_control_cmake, as it doesn't allow auto casting. It quite strict on the types

PR #129, opened 2026-02-27
Update license name to SPDX standard

eholum-nasa:
This should match https://spdx.org/licenses/Apache-2.0.html, rather than the full text string. This helps with automated parsers (like pixi) that will determine the license type from relevant package.xmls.

### DIFF
--- a/mujoco_ros2_control/package.xml
+++ b/mujoco_ros2_control/package.xml
@@ -12,7 +12,7 @@
   <maintainer email="christoph.froehlich@ait.ac.at">Christoph Froehlich</maintainer>
   <maintainer email="sai.kishor@pal-robotics.com">Sai Kishor Kothakota</maintainer>
 
-  <license>Apache-2.0</license>
+  <license>Apache License 2.0</license>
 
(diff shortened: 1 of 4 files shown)


mujoco_ros2_control/package.xml
```diff
@@ -12,7 +12,7 @@
   <maintainer email="christoph.froehlich@ait.ac.at">Christoph Froehlich</maintainer>
   <maintainer email="sai.kishor@pal-robotics.com">Sai Kishor Kothakota</maintainer>
 
-  <license>Apache License, Version 2.0</license>
+  <license>Apache-2.0</license>
```
saikishor: Maybe we should use `Apache License 2.0` as it is common for ROS 2 packages 

https://github.com/ros-controls/ros2_control/blob/master/hardware_interface/package.xml#L10

eholum-nasa: Oh yeah full name works too... updated. Sorry about that, bad copy/paste from a while ago just popped up...

saikishor (approved):

PR #130, opened 2026-03-02
Use pixi-build-ros as the backend for pixi builds

eholum-nasa:
The [pixi-build-ros](https://prefix-dev.github.io/pixi-build-backends/backends/pixi-build-ros/) does quite a bit of work for us, so it's worth including, IMO. Most notably is parsing `package.xml` to pull dependencies rather than managing them separately. This should make it a heck of a lot easier to build against the other ROS distros/push them to robostack as well.

While I'm at it I'm trying to unify some of the steps that were taken in https://github.com/RoboStack/ros-jazzy/pull/173 to get this building for robostack. There are some gnarly things that we're dealing with:

* The conda install of mujoco's simulate app actually exports symbols that we can link against, unfortunately the tar files do not. So depending on what is available (source vs no source) we still have to compile the simulate application during the build.
* I removed the download-on-the-fly as an option, this restricts us to using the vendor package to test version updates, but also removes one build patch to maintain. Given the complexity here, I'd lean towards removal but open to suggestions there. At least at this point it's dead simple to pull the vendor package in!
* Odd issues with pluginlib not being able to instantiate a class loader for the transmissions interface (that was causing CI failures). Added a note in the build file as to why. I haven't tested this yet but I expect that transmissions are currently broken in the robostack vendored package.

### DIFF
--- a/mujoco_ros2_control_tests/CMakeLists.txt
+++ b/mujoco_ros2_control_tests/CMakeLists.txt
@@ -18,27 +18,27 @@ if(BUILD_TESTING)
   find_package(launch_testing_ament_cmake REQUIRED)
 
   add_launch_test(test/robot_launch_test.py
-    TIMEOUT 50
+    TIMEOUT 75
   )
 
   add_launch_test(test/robot_launch_pid_test.py
-    TIMEOUT 50
+    TIMEOUT 75
   )
 
   add_launch_test(test/robot_launch_mjcf_generation_test.py
-    TIMEOUT 50
+    TIMEOUT 75
   )
 
   add_launch_test(test/robot_launch_mjcf_generation_from_robot_description_test.py
-    TIMEOUT 50
+    TIMEOUT 75
   )
 
   add_launch_test(test/robot_launch_transmission_test.py
-    TIMEOUT 50
+    TIMEOUT 75
   )
 
   add_launch_test(test/robot_launch_transmission_pid_test.py
-    TIMEOUT 50
+    TIMEOUT 75
   )
 endif()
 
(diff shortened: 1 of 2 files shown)


mujoco_ros2_control/package.xml
```diff
@@ -12,7 +12,7 @@
   <maintainer email="christoph.froehlich@ait.ac.at">Christoph Froehlich</maintainer>
   <maintainer email="sai.kishor@pal-robotics.com">Sai Kishor Kothakota</maintainer>
 
-  <license>Apache License 2.0</license>
+  <license>Apache-2.0</license>
```
eholum-nasa, line 15: The problem isn't the full name it's the spaces... https://github.com/prefix-dev/pixi/issues/5349. But IMO we should be using the machine names everywhere, given that it seems to be more widely parseable.

mujoco_ros2_control/package.xml
```diff
@@ -25,7 +25,6 @@
   <depend>hardware_interface</depend>
   <depend>nav_msgs</depend>
   <depend>libglfw3-dev</depend>
-  <!-- mujoco_vendor is optional - MuJoCo can also be provided via MUJOCO_INSTALL_DIR env or auto-download -->
```
saikishor, line 28: WHy do you want to remove it?

eholum-nasa: Now we get it everywhere!

Edit: That is between the apt/rosdep and pixi/conda options I am more than happy to abandon the download on the fly as another build path to maintain.

saikishor: Copy that

PR #133, opened 2026-03-03
MuJoCo ros2 control plugins

saikishor:
This PR proposes an approach to easily integrate the user's code to fetch information from the mujoco system hardware component, and decouples the complexity into the main code, and also gives the user some flexibility.

### DIFF
--- a/mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
+++ b/mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
@@ -39,8 +39,9 @@ public:
    * @param model Pointer to the MuJoCo model
    * @param data Pointer to the MuJoCo data
    * @return true if initialization was successful
-   * @note This method will be called once when the plugin is loaded. It can be used to read parameters, set up publishers/subscribers,
-   * etc. The node will be a child of the main mujoco_ros2_control node, so parameters should be namespaced accordingly.
+   * @note This method will be called once when the plugin is loaded. It can be used to read parameters, set up
+   * publishers/subscribers, etc. The node will be a child of the main mujoco_ros2_control node, so parameters should be
+   * namespaced accordingly.
    */
   virtual bool init(rclcpp::Node::SharedPtr node, const mjModel* model, mjData* data) = 0;
 
--- a/mujoco_ros2_control_plugins/src/heartbeat_publisher_plugin.cpp
+++ b/mujoco_ros2_control_plugins/src/heartbeat_publisher_plugin.cpp
@@ -21,24 +21,25 @@ namespace mujoco_ros2_control_plugins
 bool HeartbeatPublisherPlugin::init(rclcpp::Node::SharedPtr node, const mjModel* /*model*/, mjData* /*data*/)
 {
   node_ = node;
+  logger_ = node_->get_logger().get_child(node->get_sub_namespace());
 
   // Create publisher for heartbeat messages
   heartbeat_publisher_ = node_->create_publisher<std_msgs::msg::String>("mujoco_heartbeat", 10);
 
   // Initialize the last publish time
-  last_publish_time_ = std::chrono::steady_clock::now();
+  last_publish_time_ = node_->get_clock()->now();
 
-  RCLCPP_INFO(node_->get_logger(),
-              "HeartbeatPublisherPlugin initialized. Publishing to topic 'mujoco_heartbeat' every %ld second(s).",
-              publish_period_.count());
+  RCLCPP_INFO(logger_,
+              "HeartbeatPublisherPlugin initialized. Publishing to topic 'mujoco_heartbeat' every %.3f second(s).",
+              publish_period_.seconds());
 
   return true;
 }
 
 void HeartbeatPublisherPlugin::update(const mjModel* /*model*/, mjData* data)
 {
-  auto current_time = std::chrono::steady_clock::now();
-  auto elapsed = std::chrono::duration_cast<std::chrono::seconds>(current_time - last_publish_time_);
+  auto current_time = node_->get_clock()->now();
+  auto elapsed = current_time - last_publish_time_;
 
   // Check if it's time to publish
   if (elapsed >= publish_period_)
@@ -49,7 +50,7 @@ void HeartbeatPublisherPlugin::update(const mjModel* /*model*/, mjData* data)
 
     heartbeat_publisher_->publish(message);
 
-    RCLCPP_DEBUG(node_->get_logger(), "Published heartbeat #%lu at simulation time %.3f", message_count_, data->time);
+    RCLCPP_DEBUG(logger_, "Published heartbeat #%lu at simulation time %.3f", message_count_, data->time);
 
     message_count_++;
     last_publish_time_ = current_time;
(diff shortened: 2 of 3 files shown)


saikishor:
@Juliaj FYI

mujoco_ros2_control_plugins/package.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0"?>
```
eholum-nasa, line 1: @saikishor  you should just need to add this path to the pixi.toml like the others in:

https://github.com/ros-controls/mujoco_ros2_control/blob/main/pixi.toml#L32

Hopefully it just works with the build ros backend! I'll take a look at this real soon, too.

saikishor: Oops. You are right!

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -1050,6 +1062,71 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
...
+      }
+    }
+    RCLCPP_INFO_EXPRESSION(get_logger(), plugins_ns.empty(), "No 'mujoco_plugins' parameter found!");
+    RCLCPP_INFO_EXPRESSION(get_logger(), !plugins_ns.empty(),
+                           "Found 'mujoco_plugins' parameter with the following plugins: %s",
+                           fmt::format("{}", fmt::join(plugins_ns, ", ")).c_str());
+    if (plugins_ns.empty())
+    {
+      // Default: load the heartbeat publisher plugin
+      RCLCPP_INFO(get_logger(), "No 'mujoco_plugins' parameter found, loading default HeartbeatPublisherPlugin");
```
eholum-nasa: Did you mean to throw the heartbeat plugin into the list here?

That said, do we want this to be loaded by default? It's pretty innocuous I suppose.

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -1050,6 +1062,71 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
...
+    // Get list of plugins from parameter (if specified)
+    const std::string mujoco_plugins_param_prefix = "mujoco_plugins";
+    std::vector<std::string> plugins_ns;
+    const auto list_parameters = get_node()->list_parameters({ mujoco_plugins_param_prefix }, 0u);
+    const auto init_position = mujoco_plugins_param_prefix.size() + 1;  // +1 for the dot
+    for (const auto& param : list_parameters.names)
+    {
+      // find the motion key: after 'motions.', and before the next '.'
+      const auto plugin_key = param.substr(init_position, param.find_first_of('.', init_position) - init_position);
+      // Add the motion to the set of unique values
```
eholum-nasa:
```suggestion
      // find the plugin key: after 'motions.', and before the next '.'
      const auto plugin_key = param.substr(init_position, param.find_first_of('.', init_position) - init_position);
      // Add the plugin to the set of unique values
```

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -1050,6 +1062,71 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
     actuator_state_msg_.name.push_back(actuator.joint_name);
   }
 
+  // Load MuJoCo ROS2 Control plugins
+  try
```
eholum-nasa: I'm annoying, but would you mind kicking this out to an internal helper function just for clarity and to make the init method a little shorter?

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
```diff
@@ -0,0 +1,60 @@
...
+   * @brief Initialize the plugin
+   * @param node Shared pointer to the ROS 2 node for accessing parameters
+   * @param model Pointer to the MuJoCo model
+   * @param data Pointer to the MuJoCo data
+   * @return true if initialization was successful
+   */
+  virtual bool init(rclcpp::Node::SharedPtr node, const mjModel* model, mjData* data) = 0;
+
+  /**
+   * @brief Update the plugin (called every simulation step)
```
eholum-nasa, line 49: I think this is called during `read` right? So it will run at the control loop rate? Might be worth calling it out.

mujoco_ros2_control_plugins/src/heartbeat_publisher_plugin.cpp
```diff
@@ -0,0 +1,71 @@
...
+              "HeartbeatPublisherPlugin initialized. Publishing to topic 'mujoco_heartbeat' every %ld second(s).",
+              publish_period_.count());
+
+  return true;
+}
+
+void HeartbeatPublisherPlugin::update(const mjModel* /*model*/, mjData* data)
+{
+  auto current_time = std::chrono::steady_clock::now();
+  auto elapsed = std::chrono::duration_cast<std::chrono::seconds>(current_time - last_publish_time_);
```
eholum-nasa: Not sure if it was intentional or not but did you want to use the sim time here rather than wall clock time? They might be different! Not a big deal either way since it's an example, but I would explicitly call it out.

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: You are right. In my previous implementation i was using it to test it's functionality

mujoco_ros2_control_plugins/src/heartbeat_publisher_plugin.cpp
saikishor: Now that you say, it makes sense. 

Maybe I should also forward the rclcpp::Time here? That way everyone can use it for the timestamp to publish data

saikishor: What do you think about it?

eholum-nasa: I think users (and we) have options! Could either just use the node's time since that will pull the sim time, or even just the `data->time` if that's what people are interested in? I don't think we explicitly have to pass the time stamp through here?

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: Copy that

mujoco_ros2_control_plugins/src/heartbeat_publisher_plugin.cpp
saikishor: Done

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
saikishor: Done

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -2868,6 +2889,68 @@ rclcpp::Node::SharedPtr MujocoSystemInterface::get_node() const
...
+          RCLCPP_INFO(get_logger(), "Successfully loaded and initialized plugin: %s", plugin_name.c_str());
+        }
+        else
+        {
+          RCLCPP_ERROR(get_logger(), "Failed to initialize plugin: %s. Skipping it.", plugin_name.c_str());
+        }
+      }
+      catch (const pluginlib::PluginlibException& ex)
+      {
+        RCLCPP_ERROR(get_logger(), "Failed to load plugin '%s': %s", plugin_name.c_str(), ex.what());
```
eholum-nasa, line 2944: Last question is whether or not we should fail init if a plug fails to load? Should we just throw a runtime error?

But also if it becomes a problem it is easy to change later.

saikishor: IMO better not for the plugins as they are not the critical components right?.

I can add it for sure, not a problem. As you already approved this PR ;), I'll merge it and do it in a separate PR.

PR #136, opened 2026-03-05
throw on failing init of mujoco_ros2_control_plugins

saikishor:
As per comment https://github.com/ros-controls/mujoco_ros2_control/pull/133#discussion_r2887090763 of @eholum-nasa


mujoco_ros2_control/src/mujoco_system_interface.cpp
eholum-nasa, line 2953: Worth adding here, too?

saikishor: Done

PR #139, opened 2026-03-05
[Feature] Add ability to run the simulation progress by steps upon pause 

saikishor:
I've added a feature to be able to pause the simulation and then run the simulation by steps using the service or by pressing the **s** key on the simulator window. It would be a nice addition for debugging

### DIFF
--- a/mujoco_ros2_control/src/mujoco_system_interface.cpp
+++ b/mujoco_ros2_control/src/mujoco_system_interface.cpp
@@ -50,6 +50,7 @@
 #if !ROS_DISTRO_HUMBLE
 #include <hardware_interface/helpers.hpp>
 #endif
+#include <rclcpp/version.h>
 #include <hardware_interface/lexical_casts.hpp>
 #include <hardware_interface/types/hardware_interface_type_values.hpp>
 #include <rclcpp/rclcpp.hpp>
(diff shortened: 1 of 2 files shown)


mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -2857,6 +2994,26 @@ void MujocoSystemInterface::publish_clock()
 #endif
 }
 
+void MujocoSystemInterface::update_sim_display()
```
eholum-nasa, line 3093: Maybe just fast return if headless? Otherwise this is going to stack up work. 

Not that the sim is pause-able when headless... As an aside, should we add a pause service? I'd like to do that.

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -2646,6 +2684,58 @@ void MujocoSystemInterface::reset_world_callback(
...
+  step_diverged_.store(false);
+  pending_steps_.fetch_add(request->steps);
+
+  // Block until all steps are executed or the timeout expires.
+  // Timeout: at least 30 s, or 10 ms per step (whichever is larger).
+  const auto timeout =
+      std::chrono::milliseconds(std::max(static_cast<uint64_t>(30000), static_cast<uint64_t>(request->steps) * 10));
+
+  std::unique_lock<std::mutex> lock(steps_cv_mutex_);
+  const bool completed = steps_cv_.wait_for(lock, timeout, [this] { return pending_steps_.load() == 0; });
```
eholum-nasa, line 2796: Dumb edge case but, but if I unpause the sim while this is still counting down weird things seems to happen. It might be worth checking or learning the `pending_steps_` before unpausing?

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -2817,17 +2909,62 @@ void MujocoSystemInterface::PhysicsLoop()
...
+              mj_copyData(mj_data_control_, mj_model_, mj_data_);
+              sim_->AddToHistory();
+              pending_steps_.fetch_sub(1);
+              step_count_.fetch_add(1);
+              steps_cv_.notify_all();
+              update_sim_display();
+            }
+
+            // Force timing re-sync when/if simulation is resumed
+            if (pending_steps_.load() == 0)
```
eholum-nasa, line 3050: One solution to restarting while `pending_steps_` is non-zero is to add a state variable to the outside of this loop? Could be something like,

```
bool is_paused = false;
```

And then just check if the sim state has changed from paused to running at each iteration?

```
if (sim_->run && is_paused)
{
  pending_steps_.store(0);
}
is_paused = !sim_->run;
```

Or something?

eholum-nasa (commented):
Amazing.

* Tested on our robot and 🚀
* Love the docs updates

Just noted the one minor question about state management but otherwise this is just awesome. Will be so handy for RL applications too.

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: I thought of adding it, but then I thought you would say too many services XD

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: Ok I can take a look

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: Copy that!

mujoco_ros2_control/src/mujoco_system_interface.cpp
eholum-nasa: For better or worse you know me too well...

Is it a lot of services but look, so many nice docs!

saikishor: Hahhahaa. Sure

saikishor:
> Sorry I haven't had a chance to really look source yet, but was wondering if there was a reason to not use the arrow key, @saikishor? The default simulation window steps with the arrow key (although it doesn't play nice with ROS of course, which I'm sure yours does)

I'll be very frank with you, I didn't know the default simulation can step with arrow keys XD. Should I change?

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -2655,6 +2698,92 @@ void MujocoSystemInterface::reset_world_callback(
   RCLCPP_INFO(get_logger(), "%s", response->message.c_str());
 }
 
+void MujocoSystemInterface::set_pause_callback(
```
eholum-nasa, line 2725: I _think_ we might want this function to clear the step counter? If I call this while the sim is paused and stepping it will hang until the stepping completes.

That could also be the desired behavior... we could have it check if steps > - and return a failure or something?

saikishor: Yess, I'm already handling it. I have wrote a test to test the same. I'm handling it by setting them in different callback queues

saikishor: I'll push it in a while

PR #140, opened 2026-03-08
Fix reset simulation service for robots using PID control

saikishor:
The changes proposed in this PR fixes the reset simulation for all types of actuators type. We didn't realize it as the tests are not properly done: 

1. The joints are aligned with gravity, so evern resetting, they stayed in the same position
    - I modified the tests to have the joints perpendicular to the gravity to reproduce the issue
2. The tests resets the sim env, but immidiately checks for the joint states, instead wait for 1 sec to review the state.
3. I also observed redundant files in the demos repository, I cleaned them up

These above changes are done to the tests to reproduce the issue,

Current state:

https://github.com/user-attachments/assets/4ec2acd7-d8f1-4904-8c15-56a13fe3c736

With the proposed fix:

https://github.com/user-attachments/assets/ff0c28ad-818d-4927-8c5a-31041ea47c5f

### DIFF
--- a/mujoco_ros2_control/src/mujoco_system_interface.cpp
+++ b/mujoco_ros2_control/src/mujoco_system_interface.cpp
@@ -2939,12 +2939,15 @@ void MujocoSystemInterface::load_mujoco_plugins()
         }
         else
         {
-          RCLCPP_ERROR(get_logger(), "Failed to initialize plugin: %s. Skipping it.", plugin_name.c_str());
+          RCLCPP_ERROR(get_logger(), "Failed to initialize plugin: %s of type: %s", plugin_name.c_str(),
+                       plugin_type.c_str());
+          throw std::runtime_error("Failed to initialize plugin: " + plugin_name + " of type: " + plugin_type);
         }
       }
       catch (const pluginlib::PluginlibException& ex)
       {
         RCLCPP_ERROR(get_logger(), "Failed to load plugin '%s': %s", plugin_name.c_str(), ex.what());
+        throw;  // re-throw to be caught by the outer catch block
       }
     }
   }


mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -1357,24 +1357,19 @@ MujocoSystemInterface::perform_command_mode_switch(const std::vector<std::string
...
+      // write() branch active after the controller has been deactivated.
+      joint_it->is_position_control_enabled = false;
+      joint_it->is_velocity_control_enabled = false;
+      joint_it->is_effort_control_enabled = false;
+
+      actuator_it->is_position_control_enabled = false;
+      actuator_it->is_velocity_control_enabled = false;
+      actuator_it->is_effort_control_enabled = false;
+      actuator_it->is_position_pid_control_enabled = false;
+      actuator_it->is_velocity_pid_control_enabled = false;
```
eholum-nasa, line 1371: Should we just move this outside of the if block above since we disable everything before enabling, too?

mujoco_ros2_control_demos/demo_resources/mjcf_generation/test_inputs.xml
```diff
@@ -14,8 +14,8 @@
     </default>
 
     <actuator>
-      <position name="joint1" joint="joint1" kp="1000" dampratio="1.0" ctrlrange="-3.14159 3.14159" />
-      <position name="joint2" joint="joint2" kp="1000" dampratio="1.0" ctrlrange="-3.14159 3.14159" />
+      <position name="joint1" joint="joint1" kp="25000" dampratio="1.0" ctrlrange="-3.14159 3.14159" />
```
eholum-nasa, line 17: Once again I apologize for making these links 27 kg for some reason...

mujoco_ros2_control_tests/test/robot_launch_test.py
```diff
@@ -383,6 +383,10 @@ def test_reset_world_service(self):
         )
         self.node.get_logger().info("Clock continuity verified - clock was NOT reset")
 
+        # Sleep is needed here for a bit to allow the reset to take effect in the simulation (or)
+        # to see if the internal PIDs have a good setpoint after reset
+        time.sleep(1.0)
```
eholum-nasa, line 388: Do we actually need this or is the polling below sufficient?

saikishor: If I remove this, it doesn't work at all, as it receives the first message of queue and it stays with old position somehow

mujoco_ros2_control_demos/demo_resources/mjcf_generation/test_inputs.xml
saikishor: No worries ;)

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: I would need to test some things then. Can we do that in a separate PR?

Will that work for you?

mujoco_ros2_control_tests/test/robot_launch_test.py
eholum-nasa: ooooh weird. I wonder if there's some weird clock or timing thing? anyway can live with for now but we could come up with an alternative predicate if this ever causes intermittent failures!

saikishor: Sure thing. Thanks

PR #146, opened 2026-03-14
Cleanup duplicate files

christophfroehlich:
I compared them manually, should be equivalent with those in the subfolders.


saikishor (approved):
I'm fine to merge it and fix my PR

PR #148, opened 2026-03-15
Update internal links for myst-parser readthedocs

saikishor:
Related to https://github.com/ros-controls/control.ros.org/pull/588


mujoco_ros2_control/README.md
christophfroehlich, line 119: this note is not rendered correctly, because this is github markdown flavor only
https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

but I suppose this is only temporary, so I'm fine.

PR #151, opened 2026-03-20
Use requirements.txt to install python dependencies

saikishor:
This changes the way we install python dependencies to be able to freeze them to specific versions and install them.


mathias-luedtke (approved):

mujoco_ros2_control/scripts/robot_description_to_mjcf.sh
```diff
@@ -45,9 +45,9 @@ function check_dependencies() {
     fi
     # check if the dependencies trimesh, mujoco, obj2mjcf and scipy are installed, if not install them
     if ! python3 -c "import trimesh; import mujoco; import obj2mjcf" &> /dev/null; then
-        echo "Dependencies not found. Installing trimesh, mujoco, obj2mjcf, scipy...."
+        echo "Dependencies not found. Installing from requirements.txt...."
         start_time=$(date +%s)
-        pip3 install --no-input --no-cache-dir --disable-pip-version-check trimesh mujoco obj2mjcf
+        pip3 install --no-input --no-cache-dir --disable-pip-version-check -r "$(ros2 pkg prefix mujoco_ros2_control)/share/mujoco_ros2_control/requirements.txt"
```
eholum-nasa: pretty nit but since you're referencing the script location down below anyway, maybe just use it here too? Not that it will happen but this way installing deps can happen without `ros2 ...`.

```suggestion
        pip3 install --no-input --no-cache-dir --disable-pip-version-check -r "$(dirname "${BASH_SOURCE[0]}")/requirements.txt"
```

saikishor: Sounds good!

eholum-nasa (approved):

PR #154, opened 2026-03-27
Add support for "site" transmission type

christianrauch:
Support `mjTRN_SITE`, such as used with the [Skydio X2 MuJoCo model](https://github.com/google-deepmind/mujoco_menagerie/blob/main/skydio_x2/x2.xml):
https://github.com/google-deepmind/mujoco_menagerie/blob/feadf76d42f8a2162426f7d226a3b539556b3bf5/skydio_x2/x2.xml#L55-L60

```XML
  <worldbody>
    <body name="x2" pos="0 0 0.1" childclass="x2">
      <freejoint/>
      <site name="thrust1" pos="-.14 -.18 .05"/>
      <site name="thrust2" pos="-.14 .18 .05"/>
      <site name="thrust3" pos=".14 .18 .08"/>
      <site name="thrust4" pos=".14 -.18 .08"/>
    </body>
  </worldbody>

  <actuator>
    <motor class="x2" name="thrust1" site="thrust1" gear="0 0 1 0 0 -.0201"/>
    <motor class="x2" name="thrust2" site="thrust2" gear="0 0 1 0 0  .0201"/>
    <motor class="x2" name="thrust3" site="thrust3" gear="0 0 1 0 0 -.0201"/>
    <motor class="x2" name="thrust4" site="thrust4" gear="0 0 1 0 0  .0201"/>
  </actuator>
```

- [x] Fork the repository.
- [x] Pay attention to any automated CI failures reported in the pull request, and stay involved in the conversation.

### DIFF
--- a/mujoco_ros2_control_tests/test/site_velocity_launch_test.py
+++ b/mujoco_ros2_control_tests/test/site_velocity_launch_test.py
@@ -154,8 +154,8 @@ class TestSiteVelocityActuator(unittest.TestCase):
         self.assertTrue(
             self.spin_until(
                 lambda: self.get_joint_value(self._latest_actuator_js, "velocity", "slider_joint") is not None
-                and self.get_joint_value(self._latest_actuator_js, "velocity", "slider_joint") > 0.01,
-                timeout=1.0,
+                and self.get_joint_value(self._latest_actuator_js, "velocity", "slider_joint") > 0.4,
+                timeout=2.0,
             ),
             "Actuator state did not report sufficiently positive slider velocity",
         )
--- a/mujoco_ros2_control_tests/test/site_velocity_launch_test.py
+++ b/mujoco_ros2_control_tests/test/site_velocity_launch_test.py
@@ -146,16 +146,16 @@ class TestSiteVelocityActuator(unittest.TestCase):
             self.spin_until(
                 lambda: self.get_joint_value(self._latest_js, "position", "slider_joint") is not None
                 and self.get_joint_value(self._latest_js, "position", "slider_joint") > start_position + 0.1,
-                timeout=10.0,
+                timeout=1.0,
             ),
-            "Slider joint did not move after commanding the site velocity actuator",
+            "Slider joint did not move sufficiently after commanding the site velocity actuator",
         )
 
         self.assertTrue(
             self.spin_until(
                 lambda: self.get_joint_value(self._latest_actuator_js, "velocity", "slider_joint") is not None
-                and self.get_joint_value(self._latest_actuator_js, "velocity", "slider_joint") > 0.1,
-                timeout=5.0,
+                and self.get_joint_value(self._latest_actuator_js, "velocity", "slider_joint") > 0.01,
+                timeout=1.0,
             ),
-            "Actuator state did not report positive slider velocity",
+            "Actuator state did not report sufficiently positive slider velocity",
         )
--- a/mujoco_ros2_control_tests/test/site_velocity_launch_test.py
+++ b/mujoco_ros2_control_tests/test/site_velocity_launch_test.py
@@ -145,8 +145,8 @@ class TestSiteVelocityActuator(unittest.TestCase):
         self.assertTrue(
             self.spin_until(
                 lambda: self.get_joint_value(self._latest_js, "position", "slider_joint") is not None
-                and self.get_joint_value(self._latest_js, "position", "slider_joint") > start_position + 0.1,
-                timeout=1.0,
+                and self.get_joint_value(self._latest_js, "position", "slider_joint") > start_position + 0.5,
+                timeout=2.0,
             ),
             "Slider joint did not move sufficiently after commanding the site velocity actuator",
         )
--- a/mujoco_ros2_control_tests/test/site_velocity_launch_test.py
+++ b/mujoco_ros2_control_tests/test/site_velocity_launch_test.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python3
+
+# Copyright 2026 PAL Robotics S.L.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import time
+import unittest
+
+from ament_index_python.packages import get_package_share_directory
+from controller_manager.test_utils import check_controllers_running, check_if_js_published, check_node_running
+from controller_manager_msgs.srv import ListHardwareInterfaces
+from launch import LaunchDescription
+from launch.actions import IncludeLaunchDescription
+from launch.launch_description_sources import PythonLaunchDescriptionSource
+from launch_testing.actions import ReadyToTest
+from launch_testing.util import KeepAliveProc
+from launch_testing_ros import WaitForTopics
+import pytest
+import rclpy
+from rosgraph_msgs.msg import Clock
+from sensor_msgs.msg import JointState
+from std_msgs.msg import Float64MultiArray
+
+
+@pytest.mark.rostest
+def generate_test_description():
+    launch_include = IncludeLaunchDescription(
+        PythonLaunchDescriptionSource(
+            os.path.join(
+                get_package_share_directory("mujoco_ros2_control_tests"),
+                "launch/site_velocity_test_launch.py",
+            )
+        ),
+        launch_arguments={"headless": "true"}.items(),
+    )
+
+    return LaunchDescription([launch_include, KeepAliveProc(), ReadyToTest()])
+
+
+class TestSiteVelocityActuator(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        rclpy.init()
+
+    @classmethod
+    def tearDownClass(cls):
+        rclpy.shutdown()
+
+    def setUp(self):
+        self.node = rclpy.create_node("site_velocity_test_node")
+        self._latest_js = None
+        self._latest_actuator_js = None
+        self._js_sub = self.node.create_subscription(JointState, "/joint_states", self._joint_state_cb, 10)
+        self._actuator_sub = self.node.create_subscription(
+            JointState, "/mujoco_actuators_states", self._actuator_state_cb, 10
+        )
+
+    def tearDown(self):
+        self.node.destroy_node()
+
+    def _joint_state_cb(self, msg):
+        self._latest_js = msg
+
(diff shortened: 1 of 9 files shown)


saikishor (changes requested):
Hello,

Can you please add a test for the new behaviour?

saikishor (changes requested):
Thank you. Please add a test for the site type of actuation. You can inspire from the existing tests

saikishor (changes requested):
@christianrauch IMO This is not enough. I prefer you add a new demos based on the sites and then test it like how it is done with mujoco_ros2_control_tests, so we know thw whole physics pipeline would work

christianrauch:
I am a bit confused as to where tests have to be added. I added the new `site` actuator to the `mujoco_ros2_control` package tests, since this is the ROS package I added the feature to. The test in this package only check, if the plugin can be loaded and can be initialised with an in-source hard-coded URDF.

The `robot_launch_test.py` in `mujoco_ros2_control_tests` eventually uses `demo.launch.py` from `mujoco_ros2_control_demos`, which in turn launches `01_basic_robot.launch.py`, and this finally uses another hard-coded description from file `test_robot.urdf`.

What is the purpose of the tests in `mujoco_ros2_control`, if the actual tests happen in `mujoco_ros2_control_tests` and `mujoco_ros2_control_demos`? Shouldn't the tests for the `mujoco_ros2_control` package be inside the package, just like for the controllers in `ros2_controllers`?

How do I add a new robot to this test chain?

saikishor:
> I am a bit confused as to where tests have to be added. I added the new `site` actuator to the `mujoco_ros2_control` package tests, since this is the ROS package I added the feature to. The test in this package only check, if the plugin can be loaded and can be initialised with an in-source hard-coded URDF.

We have the `mujoco_ros2_control_tests` as functional tests, same as in the [gz_ros2_control](https://github.com/ros-controls/gz_ros2_control). What you have added in the mujoco_ros2_control serve maybe to see that nothing is crashing, but functionally if it is working or not, it is important to take a look.

> The `robot_launch_test.py` in `mujoco_ros2_control_tests` eventually uses `demo.launch.py` from `mujoco_ros2_control_demos`, which in turn launches `01_basic_robot.launch.py`, and this finally uses another hard-coded description from the file `test_robot.urdf`.

Yes, you would need to add a new robot URDF and add a new demo that uses site actuation. 

> What is the purpose of the tests in `mujoco_ros2_control`, if the actual tests happen in `mujoco_ros2_control_tests` and `mujoco_ros2_control_demos`? Shouldn't the tests for the `mujoco_ros2_control` package be inside the package, just like for the controllers in `ros2_controllers`?

As I mentioned above, the functional testing is what is happening with the `mujoco_ros2_control_tests` it is also similar to the tests within ros2_control_demos as well. We still need to improve the unit testing part in the mujoco_ros2_control package

christianrauch:
> Yes, you would need to add a new robot URDF and add a new demo that uses site actuation.

I don't really have the time (or motivation) to create a fully fledged simulation example with a new robot and duplicated launch infrastructure. The change I am proposing is rather trivial, and setting `actuator_data.joint_name = std::string(act_name);` could be the fallback for transmission types that do not require special handling anyway. In the end, the values are just forwarded to MuJoCo and their effect are depending on the model. Checking that the model behaves accordingly is rather a task of the MuJoCo tests.

saikishor:
> > Yes, you would need to add a new robot URDF and add a new demo that uses site actuation.
> 
> I don't really have the time (or motivation) to create a fully fledged simulation example with a new robot and duplicated launch infrastructure. The change I am proposing is rather trivial, and setting `actuator_data.joint_name = std::string(act_name);` could be the fallback for transmission types that do not require special handling anyway. In the end, the values are just forwarded to MuJoCo and their effect are depending on the model. Checking that the model behaves accordingly is rather a task of the MuJoCo tests.

Sorry I'm not interested to merge changes without having proper tests even so if they are trivial changes. The idea is to have proper functional testing. What's your opinion @ndunkelb-nasa @eholum-nasa ?


You can have these changes local and use it for your usecase :).

christianrauch:
I ended up vibe coding a test case (`site_velocity_launch_test`) with a new robot model that defines a `velocity` `actuator` at a `site`.

I checked that
```sh
colcon test --packages-select mujoco_ros2_control_tests --ctest -R site_velocity_launch_test -V --event-handlers console_direct+
```
fails with
```
[MujocoSystemInterface]: Unsupported transmission type '4' for actuator 'slider_joint'
[MujocoSystemInterface]: Failed to register MuJoCo actuators, exiting...
[controller_manager]: Failed to initialize hardware 'MujocoSystem'
```
without the `site` support, and that the `site_velocity_launch_test` test passes with the changes in this PR.

christianrauch:
@saikishor Is the test I added ok?

christianrauch:
@saikishor @ndunkelb-nasa @eholum-nasa Can any of you comment on the tests and whether this can be merged now?

mujoco_ros2_control_tests/test/site_velocity_launch_test.py
```diff
@@ -0,0 +1,164 @@
...
+        command.data = [0.5]
+        end_time = time.time() + 2.0
+        while time.time() < end_time:
+            pub.publish(command)
+            rclpy.spin_once(self.node, timeout_sec=0.1)
+
+        self.assertTrue(
+            self.spin_until(
+                lambda: self.get_joint_value(self._latest_js, "position", "slider_joint") is not None
+                and self.get_joint_value(self._latest_js, "position", "slider_joint") > start_position + 0.1,
```
saikishor:
```suggestion
                and self.get_joint_value(self._latest_js, "position", "slider_joint") > start_position + 0.5,
```

Should we set 0.5 the one corresponding to 1 sec of Spin?

saikishor (commented):
Looking good. Just a minor thing.

Should be also increase the velocity one?

mujoco_ros2_control_tests/test/site_velocity_launch_test.py
christianrauch: I think the concrete distance travelled does not matter so much here. The test only checks if the robot/cube moved more than 0.1 within 10 seconds and that its velocity reached 0.1 within 5 seconds.

We could check for an exact distance travelled in a given time. But do you think this is really necessary?

Increasing these values also means that the test will take longer.

saikishor:
@christianrauch from next time, don't force push changes on any of ros-controls repos, because for the maintainers we have to go through all the changes again and all the review is lost. 

Thank you ;)

saikishor:
@christianrauch what changes did you force push yesterday, I couldn't get catch of them

mujoco_ros2_control_tests/test/site_velocity_launch_test.py
saikishor: If your velocity is 0.5 m/sec, why would it take 10 seconds to move 0.1m, it should take 0.2+ seconds right? Increasing it a bit shouldn't affect it a lot

saikishor (commented):
I updated my comment

christianrauch:
> @christianrauch what changes did you force push yesterday, I couldn't get catch of them

I pushed the last commit. I removed the spinning after `pub.publish(command)` as this is actually not required and saves some time when running the test.

mujoco_ros2_control_tests/test/site_velocity_launch_test.py
christianrauch: It does not take 10 seconds. That is just the upper time limit at which this position should be reached. Essentially, it is a timeout. If `self.get_joint_value(self._latest_js, "position", "slider_joint") > start_position + 0.1` is not true within 10 seconds, the test fails.

christianrauch: I could reduce the timeout to make this test stricter. But right now, this only takes 10 seconds, and will fail, in the worst case.

Do you want to use a lower timeout here?

saikishor: Yup, why not increase it to be a realistic value?. As for the simulation, it should be able to reach 0.5 within 1 second, so,  it should be fine if we increase the tolerances. And also I would increase the velocity one to be closer

christianrauch: I pushed an additional commit that tunes the timeouts and thresholds.

saikishor: Why reduce the timeout?. I don't get your recent commits..

What I'm looking for is the joint moves approx 0.5m in a sec for a velocity command of 0.5 m/sec

mujoco_ros2_control_tests/test/site_velocity_launch_test.py
```diff
@@ -0,0 +1,161 @@
...
+                timeout=1.0,
+            ),
+            "Slider joint did not move sufficiently after commanding the site velocity actuator",
+        )
+
+        self.assertTrue(
+            self.spin_until(
+                lambda: self.get_joint_value(self._latest_actuator_js, "velocity", "slider_joint") is not None
+                and self.get_joint_value(self._latest_actuator_js, "velocity", "slider_joint") > 0.01,
+                timeout=1.0,
```
saikishor:
```suggestion
                and self.get_joint_value(self._latest_actuator_js, "velocity", "slider_joint") > 0.4,
                timeout=2.0,
```
It should reach atleast 0.4 given the timeout and the command

saikishor (changes requested):

mujoco_ros2_control_tests/test/site_velocity_launch_test.py
```diff
@@ -0,0 +1,161 @@
...
+
+        command = Float64MultiArray()
+        command.data = [0.5]
+        pub.publish(command)
+
+        self.assertTrue(
+            self.spin_until(
+                lambda: self.get_joint_value(self._latest_js, "position", "slider_joint") is not None
+                and self.get_joint_value(self._latest_js, "position", "slider_joint") > start_position + 0.1,
+                timeout=1.0,
```
saikishor:
```suggestion
                and self.get_joint_value(self._latest_js, "position", "slider_joint") > start_position + 0.25,
                timeout=1.0,
```

mujoco_ros2_control_tests/test/site_velocity_launch_test.py
christianrauch: Well, you suggested having more realistic and stricter tolerances. No?

In the end, this test only has to check that the robot slider moves based on the velocity command on the site. Everything else is optional. To reduce the test time, I would simply suggest using the smallest position and velocity that verify that the robot moved.

Increasing the position and velocity threshold increases the test duration without testing something new. I suggest we make the test run as fast as possible by reducing the position and velocity thresholds.

mujoco_ros2_control_tests/test/site_velocity_launch_test.py
saikishor: I modified my suggestion now

christianrauch:
@saikishor I applied your suggestions. Since the robot now has to travel 0.5 metres instead of only 0.1 metre, the test will run slightly longer, though.

saikishor (approved):

PR #159, opened 2026-03-28
improve debugging

christianrauch:
These are some minor tweaks to print more information about waiting for messages or missing interfaces to improve debugging.

### DIFF
--- a/mujoco_ros2_control_tests/test/robot_launch_test.py
+++ b/mujoco_ros2_control_tests/test/robot_launch_test.py
@@ -236,14 +236,14 @@ class TestFixture(unittest.TestCase):
             rclpy.spin_once(self.node, timeout_sec=0.1)
 
         # Wait for joints to reach target positions
-        self.wait_for_joint_positions({"joint1": 0.5, "joint2": -0.5}, delta=0.05, timeout=5.0)
+        self.wait_for_joint_positions({"joint1": 0.5, "joint2": -0.5}, delta=0.05, timeout=15.0)
 
         if os.environ.get("TEST_TRANSMISSIONS") != "true":
             expected_actuators = {"joint1": 0.5, "joint2": -0.5}
         else:
             expected_actuators = {"actuator1": 0.5 * 2.0, "actuator2": -0.5 * 0.5}
 
-        self.wait_for_joint_positions(expected_actuators, delta=0.05, timeout=5.0, topic="actuator_states")
+        self.wait_for_joint_positions(expected_actuators, delta=0.05, timeout=15.0, topic="actuator_states")
 
         # make sure the efforts field is non-zero (indicating PID/ effort reporting is working)
         self.assertTrue(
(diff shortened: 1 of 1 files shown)


mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -2489,13 +2489,30 @@ void MujocoSystemInterface::register_sensors(const hardware_interface::HardwareI
...
+
+      if (accel_id == -1)
+      {
+        RCLCPP_ERROR_STREAM(get_logger(), "Failed to find IMU sensor '" << sensor_data.linear_acceleration.name
+                                                                        << "' in MuJoCo model");
+      }
 
       if (quat_id == -1 || gyro_id == -1 || accel_id == -1)
       {
-        RCLCPP_ERROR_STREAM(get_logger(), "Failed to find IMU sensor in MuJoCo model, sensor name: " << sensor.name);
```
saikishor, line 2498: Add it into the condition and better use RCLCPP_ERROR_EXPRESSION for this

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -465,7 +465,7 @@ mjModel* loadModelFromTopic(rclcpp::Node::SharedPtr node, const std::string& top
 
   rclcpp::QoS qos_profile(rclcpp::QoSInitialization::from_rmw(rmw_qos_profile_default));
   qos_profile.reliable().transient_local().keep_last(1);
-  RCLCPP_INFO(node->get_logger(), "Trying to get the MuJoCo model from topic");
+  RCLCPP_INFO_STREAM(node->get_logger(), "Trying to get the MuJoCo model from topic '" << topic_name << "'");
```
saikishor, line 468:
```suggestion
  RCLCPP_INFO(node->get_logger(), "Trying to get the MuJoCo model from topic '%s'", topic_name.c_str());
```

mujoco_ros2_control/src/mujoco_system_interface.cpp
christianrauch: Done.

mujoco_ros2_control/src/mujoco_system_interface.cpp
christianrauch: Why not the `_STREAM` functions? They are already used in the code, are closer to the C++ API, and provide more convenient formatting.

eholum-nasa: Is there a reason to avoid them? I think I usually default to them but also acknowledge that no standard was ever enforced in the system interface one way or the other...

saikishor: I avoid them as much as I can, because the STREAM are not RT safe at all, as they first construct the string and then check for the type. The non stream version are better optimized for the compilers. I prefer that we stick to the INFO as much as we can

christianrauch: Ha. I didn't know that. I always preferred the "C++" API, as it provides better formatting and does not require converting the C++ strings to C strings.

Is there some consensus or style guide for this in ROS control?

saikishor: > Is there some consensus or style guide for this in ROS control?

Nope 🙈

christianrauch: Well, I also found these sorts of usages in the codebase combining `_STREAM` with `.c_str()`:
https://github.com/ros-controls/mujoco_ros2_control/blob/f7b494b5fc34e09f5640e4140742ef858cdcbf09/mujoco_ros2_control/src/mujoco_system_interface.cpp#L2520-L2521

So, can I keep my `_STREAM` for now and you fix this at another point for the whole codebase? :-)

eholum-nasa: Yeah there are a lot of them... I'm happy to fall on this hand grenade since I'm pretty sure they're like 99% mine!

That's good to know, @saikishor, I did not realize that.

PR #165, opened 2026-04-06
Add more useful plugins for the mujoco_ros2_control

saikishor:
Hello!

I've added 2 more new useful plugins for the mujoco_ros2_control_plugins.

1. Constraint violations publisher plugin - If there are closed loop soft constraints, this basically tells you how bad the constraint is at a given point of time

```
---
stamp:
  sec: 19
  nanosec: 539999999
names:
- leg_left_closed_kin_joint
- leg_right_closed_kin_joint
types:
- connect
- connect
violations:
- 4.296666276548677e-06
- 4.298455079931873e-06
---
```

2. External Wrench plugin - The major part of this plugin was made using AI, but this is mainly to be able to apply external wrenches to a given link, this is very useful inorder to test different control strategies such as cartesian impedance or the robustness of humanoid walking with pushes etc.


mujoco_ros2_control/include/mujoco_ros2_control/mujoco_system_interface.hpp
```diff
@@ -375,6 +375,16 @@ class MujocoSystemInterface : public hardware_interface::SystemInterface
   realtime_tools::RealtimePublisher<nav_msgs::msg::Odometry>::SharedPtr floating_base_realtime_publisher_ = nullptr;
   nav_msgs::msg::Odometry floating_base_msg_;
 
+  // Plugin marker publisher — created only when at least one plugin is loaded.
+  std::shared_ptr<rclcpp::Publisher<visualization_msgs::msg::MarkerArray>> plugin_marker_pub_raw_ = nullptr;
```
eholum-nasa: Should all of this just live in the plugin itself? I have serious concerns about coupling very specific visualization logic to the hardware system interface.

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -1574,6 +1575,35 @@ hardware_interface::return_type MujocoSystemInterface::read(const rclcpp::Time&
     plugin->update(mj_model_, mj_data_control_);
   }
 
+  // Collect and publish visualization markers from all plugins at a controlled rate.
+  // Markers are given a finite lifetime (2x the publish period) so they auto-expire
+  // in RViz when a plugin stops contributing them.
+  if (plugin_marker_pub_ && (time - last_marker_publish_time_).seconds() >= marker_publish_period_)
```
eholum-nasa: Yeah and I also worry that `read` and `write` are getting a little too bloated. We have a pretty well defined (and simple) hook for plugins that provides near god-like access (along with the mj_data getters and setters). I think this complicates that API in a way that I'm not sure is required, and it would probably be preferable to force people to use those interfaces if we can.

If users did want a shared constructs between multiple plugins then they could always setup some kind of static object or a singleton in another plugin that made it available across the board?

That said, for just these two plugins I would _probably_ lean towards each just having their own publisher?

mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
```diff
@@ -0,0 +1,339 @@
...
+  return true;
+}
+
+void ExternalWrenchPlugin::update(const mjModel* /*model_arg*/, mjData* data)
+{
+  // Step 1 – undo our previous contribution so other plugins / controllers
+  //          that also write qfrc_applied are not permanently affected.
+  for (int i = 0; i < nv_; ++i)
+  {
+    data->qfrc_applied[i] -= qfrc_prev_contribution_[i];
```
eholum-nasa: could weird things happen here if `qfrc_applied` is otherwise updated? By the sim, another plugin, or a controller? I'm really not sure but this smells funny to me. I worry that there are quite a few things than could go wrong here unless used in a very specific way (not unlike the rest of the drivers I guess...).

mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
```diff
@@ -0,0 +1,339 @@
...
+
+  RCLCPP_INFO(logger_,
+              "Applying wrench on body '%s' (id=%d) for %.3f s (ramp-down %.3f s). "
+              "Force [%.2f, %.2f, %.2f] N, Torque [%.2f, %.2f, %.2f] N·m, "
+              "Application point [%.3f, %.3f, %.3f] m (body frame).",
+              request->link_name.c_str(), body_id, duration.seconds(), w.ramp_down_duration.seconds(), w.force[0],
+              w.force[1], w.force[2], w.torque[0], w.torque[1], w.torque[2], w.application_point[0],
+              w.application_point[1], w.application_point[2]);
+
+  // Block the service thread until the wrench duration has elapsed so that
```
eholum-nasa: Does this actually need to wait to return or could it just return success if the wrench was added?

mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
```diff
@@ -0,0 +1,339 @@
...
+  {
+    node_->declare_parameter("force_arrow_scale", force_arrow_scale_);
+  }
+  if (!node_->has_parameter("torque_arrow_scale"))
+  {
+    node_->declare_parameter("torque_arrow_scale", torque_arrow_scale_);
+  }
+  if (!node_->has_parameter("marker_frame_id"))
+  {
+    node_->declare_parameter("marker_frame_id", marker_frame_id_);
```
eholum-nasa: What does this param do?

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/external_wrench_plugin.hpp
```diff
@@ -0,0 +1,138 @@
...
+#include <rclcpp/rclcpp.hpp>
+#include <visualization_msgs/msg/marker_array.hpp>
+
+#include "mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp"
+
+namespace mujoco_ros2_control_plugins
+{
+
+/**
+ * @brief Plugin that applies an external wrench (force + torque) to a named
```
eholum-nasa: This is like a throw back (or is it available on new gazebo?) to gazebo's `apply_body_wrench` or whatever it was called?

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/constraint_violations_publisher_plugin.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+/// Precomputed metadata for an equality constraint
+struct ConstraintMetadata
+{
+  std::string name;
+  std::string type;
+};
+
+/**
+ * @brief Plugin that reads MuJoCo equality-constraint violations every step and
```
eholum-nasa: I could see this being very useful for our parallel jaw grippers...

mujoco_ros2_control_plugins/src/constraint_violations_publisher_plugin.cpp
```diff
@@ -0,0 +1,267 @@
...
+  if (!realtime_publisher_)
+  {
+    return;
+  }
+
+  constraint_violations_msg_.stamp = now;
+  // Resize to actual violation count; vectors were preallocated in init()
+  constraint_violations_msg_.names.resize(active_count);
+  constraint_violations_msg_.types.resize(active_count);
+  constraint_violations_msg_.violations.resize(active_count);
```
eholum-nasa: What if we just published every constraints' status every single loop? Would that be too much noise or would it be beneficial to have the details for everything?

It would also simplify update slightly since we wouldn't have to resize.

eholum-nasa (commented):
I _really_ like seeing these. You should checkout the headers @ndunkelb-nasa based on our conversations this week.

I do have big concerns about adding the complexity to the hardware interface though. I think we should bias towards simplifying whatever we can and sticking to as simple a plugin interface as possible, until it becomes untenable. Does that make sense?

The other question I have is how to decide if/when a plugin goes into the main package vs when it should be owned by the user? I just don't want our maintenance burden to continue to grow unchecked. What do you think?

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_system_interface.hpp
saikishor: First of all I thought of the same, but then my final thoughts were to provide a single way to do things. For instance, If every plugin has to publish their own visualization message and then to visualize, we will have to basically add n marker arrays to the RViZ; instead, this way we have a single topic that can support multiple visualizations under different namespaces, and you are free to toggle the namespace within the plugin directly. That was my motivation behind moving it to generic place. Moreover, it leaves us less things to document for each and every plugin.

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: I've answered above my intention for a common publisher, if you don't like it I can switch to isolated one. I thought it would be nicer to have a common interface. What I thought it, one day if we want to visualize the constraint violation in the RViz directly all the time (Like if it is more than 0.1, show in red instead of white), then we can simply add it to a different namespace and 🎉  you already have it.

Regarding the bloating of read and write, one thing I can do is to place this in a separate method.

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/constraint_violations_publisher_plugin.hpp
saikishor: Nice. I wanted to have something like this for our closed kinematic humanoids, it is good to know if there is something is breaking these soft contraints and we should know it

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/external_wrench_plugin.hpp
saikishor: I don't know. We have something similar for our internal physics simulator. This is an inspiration from that

mujoco_ros2_control_plugins/src/constraint_violations_publisher_plugin.cpp
saikishor: Sure, we can publish all the constraints. I like the idea. I can implement it.

mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
saikishor: THanks for catching this. At the beginning I was publishing them in the world frame and then I encountered many issues, this should be removed

mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
saikishor: Now that you say, I think I can improve the logic here significantly to only perform when there is a service and not to affect other scenarios. What do you think?

mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
saikishor: I think we should block for the user's given time, that way the user don't have to sleep during this operation or scripting.

saikishor (commented):
> I do have big concerns about adding the complexity to the hardware interface though. I think we should bias towards simplifying whatever we can and sticking to as simple a plugin interface as possible, until it becomes untenable. Does that make sense?

Regd. the visualization markers, I've added my point of view in one of the comments. Let me know

> The other question I have is how to decide if/when a plugin goes into the main package vs when it should be owned by the user? I just don't want our maintenance burden to continue to grow unchecked. What do you think?

This is a very good question. We should see the generality of the plugin if it can be used for more use cases, and then if it meets the criteria and if it has appropriate tests, maybe we can consider having it in the original repo. If not, user has to maintain it on his own in their own repo. 

P.S: If one of these plugins is too much and doesn't follow the criteria, I can have them in our internal gitlab and maintain them myself.

Let me know what you think about my comments and based on that I can address them.

Thanks a lot for an amazing review @eholum-nasa

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_system_interface.hpp
eholum-nasa: I do like the idea of common tools for rviz access for plugins. Though for this publisher in particular, I think my preference would be to leave it in the plugin itself (at least for now), as the system interface is so complicated as it is. 

If we start to find that more and more plugins require more access to common resources I think some kind of container for rviz-related resources that can be accessed and managed by the plugins themselves would make sense. To me that is preferable to adding things directly to the system interface. This could potentially be some kind lazily instantiated static singleton. Say a publisher that only gets instantiated when needed by a plugin. Or some other kind of shared context that gets passed to the plugin on initialization. That would allow them to share resources without having to modify the API between plugins and the system interface, which I think is a high priority right now.

All that said, I do think that we should leave this in the plugin itself in the interest of scope and simplicity of this change.

mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
eholum-nasa: I think that makes sense! If no action is requested in the plug-in should do nothing, basically?

eholum-nasa (commented):
Thanks @saikishor! Added replies in threads.

> If one of these plugins is too much and doesn't follow the criteria, I can have them in our internal gitlab and maintain them myself.

The apply force plugin I could definitely see broad use cases. Constraints violations are probably less applicable to the kinds of things we generally do, but if you think that they meet the criteria listed, then I think it makes sense.

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_system_interface.hpp
saikishor: Agreed. Given that only there is one usecase for "now", I think it is in best interest of everyone that it only lives in the plugin itself. Noted 🫡

saikishor:
> The apply force plugin I could definitely see broad use cases. Constraints violations are probably less applicable to the kinds of things we generally do, but if you think that they meet the criteria listed, then I think it makes sense.

Personally,  I agree with you. Constraint violations is something that is interesting to us, but I don't see the generality of this plugin for the community. I see the broad use cases of the wrench plugin as you do. 

I'll remove the constraints violation plugin and keep this PR only on the wrench plugin. Thank you @eholum-nasa, for your insights

ndunkelb-nasa:
I think I agree with all of the above comments! Especially with the comments about the visualization living in the plugin. Like having a dependency on visualization_msgs in the core part of the hardware interface seems semantically wrong to me. I also feel like adding infrastructure in the main HWI loop to support arbitrary plugins could be a slippery slope for adding more and more components to support potential other implementations.

But it looks like you guys are already at the same conclusion, so good to go there.

mujoco_ros2_control_msgs/srv/ApplyExternalWrench.srv
```diff
@@ -0,0 +1,35 @@
+# Apply an external wrench to a named body/link in the MuJoCo simulation
+# for a given duration.
+
+# Standard ROS header (stamp, frame_id)
+std_msgs/Header             header
+
+# Name of the MuJoCo body to apply the wrench to (must match the MJCF body name)
+string                      link_name
+
+# Wrench expressed in the world frame.
```
ndunkelb-nasa: I do wonder whether this could be modified slightly to allow the user to express a wrench in a different frame? I think you already have the tough math in here so it wouldn't be that hard to modify. This could just be a WrenchStamped instead to support that?

mujoco_ros2_control_msgs/srv/ApplyExternalWrench.srv
ndunkelb-nasa, line 1: I also wonder about this being a service with a very specifically defined force profile. Could this instead maybe be a topic? Then it is the responsibility of the implementer to make a profile and pass it to the topic? Like maybe you want a sinusoidal profile? It seems like if you want to do that, you might end up making this service exceedingly complex.

Maybe that is a different plugin though, and this fits the use case of most people!

mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
```diff
@@ -0,0 +1,339 @@
...
+  }
+  force_arrow_scale_ = node_->get_parameter("force_arrow_scale").as_double();
+  torque_arrow_scale_ = node_->get_parameter("torque_arrow_scale").as_double();
+  marker_frame_id_ = node_->get_parameter("marker_frame_id").as_string();
+
+  service_ = node_->create_service<ApplyExternalWrench>("apply_wrench",
```
ndunkelb-nasa, line 49: It seems like the goal is to let multiple wrenches be applied at the same time? With that in mind, should this have a reentrant callback group or something? Or am I missing something here?

mujoco_ros2_control_msgs/srv/ApplyExternalWrench.srv
saikishor: Maybe yes, it could be nice to have the topic version too. We can take a look on how to extend the current one or to add a different plugin. I agree, the usecase is completely valid. 

We didn't have this ramp profile internally, I just added it thinking it is useful

mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
saikishor: Right now, it is not possible to set multiple wrenches at same time, I can take a look at it

ndunkelb-nasa: Got it, no big deal. There were just a couple of times with the format:
```cpp
  for (auto& w : active_wrenches_)
```
So I figured the intention was to support multiple.

saikishor: I'll try to see if I can fit it, in the current PR

ndunkelb-nasa: I think it is fine as is, and could always upgrade in another PR if wanted!

saikishor:
Just to inform, I'm already working on the changes, I've some changes locally, but once I'm done I'll push them during this week

saikishor:
https://github.com/user-attachments/assets/b007e46b-b944-4821-97f3-36751dee7555

Here is the final version, with this you can also visualize the forces both in the native viewer and also in the RViz

saikishor:
@eholum-nasa @ndunkelb-nasa Sorry for taking longer than expected. it took me quite sometime to fix some thing

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/external_wrench_plugin.hpp
```diff
@@ -0,0 +1,174 @@
...
+    rclcpp::Duration ramp_down_duration{ 0, 0 };
+    double current_scale{ 1.0 };  ///< ramp-down scale, used by markers too
+  };
+
+  /// Service callback — runs in a ROS executor thread.
+  void handleApplyWrench(const ApplyExternalWrench::Request::SharedPtr request,
+                         ApplyExternalWrench::Response::SharedPtr response);
+
+  /// Publish the result of publish_markers() via the realtime publisher. Called from update().
+  void publishMarkers();
```
eholum-nasa: camel case or no? It is `publish_markers` above!

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -2928,6 +2907,26 @@ void MujocoSystemInterface::PhysicsLoop()
       // run only if model is present
       if (mj_model_)
       {
+        // Determine the viewer (drag) forces for this outer iteration.
+        //
+        // mjv_updateScene in simulate.cc reads mj_data_->xfrc_applied BEFORE zeroing it, so
```
eholum-nasa, line 2912: Is this in reference to: https://github.com/google-deepmind/mujoco/blob/3.4.0/simulate/simulate.cc#L2222 ? I don't fully understand how `mjvPerturb` is used there but I'm assuming that's the mouse input perturbations?

My primary concern is that there are a growing number of logical components that are competing for access to the same buffers in `mj_data_control_` with what I would call "loose" safety mechanisms. Really what we're looking to do is offer users a way to inject energy into the system that meshes nicely with the underlying physics sim, right? `mj_data_control_`  was originally added as an efficiency mechanism to allow controllers state information without locking the sim. But it's growing into a more comprehensive holding area for both reading and writing, and we're into an awful lot of "you just have to know" kinds of things for writing plugins. e.g., the fact that `xfrc_applied` is zeroed at every read iteration, or that there are a few fields we've carved out as "protected" in the control data that don't get overwritten. It's gotten impossible to keep track of...

I think if we want to offer those kinds of features through the plugin interface then we should do it through a better defined interface. FWIW, I think the next step after https://github.com/ros-controls/mujoco_ros2_control/pull/175/ might be to make a write control buffer and a read control buffer, just to make sure that what is happening with each is better defined. This would make the `mj_data_control_` buffer less of an all in one staging area, and hopefully completely isolate `mj_data_` construct behind the simulation curtain. 

All that said, for now my vote would be to keep things simple and drop the native Simulate app rendering for these forces, even though it's an overall loss of a neat feature. We could just have the physics loop compose `xfrc_applied = mju_addTo(existing xfrc_applied, plugin forces, ...)` each step? I think that would really isolate the changes here until we can better isolate the underlying physics sim. Does that make sense? Sorry to push back on this I just think we need to make things a little cleaner before we carve the physics loop up more!

mujoco_ros2_control_plugins/README.md
```diff
@@ -16,16 +16,123 @@ The `mujoco_ros2_control_plugins` package is designed to contain plugins that ex
...
+| `wrench.wrench.force` | `geometry_msgs/Vector3` | Linear force [N] — see `apply_in_body_frame` for reference frame |
+| `wrench.wrench.torque` | `geometry_msgs/Vector3` | Angular moment [N·m] — see `apply_in_body_frame` for reference frame |
+| `application_point` | `geometry_msgs/Point` | Force application point in the **body (link) frame** (relative to body frame origin, metres). Zero → apply at the body frame origin. |
+| `duration` | `builtin_interfaces/Duration` | How long the wrench remains active. Zero → single simulation step. |
+| `ramp_down_duration` | `builtin_interfaces/Duration` | Duration over which the wrench linearly ramps from full magnitude to zero at the end of `duration`. Zero → no ramp-down. |
+| `apply_in_body_frame` | `bool` | **`false` (default) — World frame**: force/torque direction is fixed in the world frame and does not change as the body rotates. Use this for gravity-like or globally-directed pushes. **`true` — Body frame**: force/torque is expressed in the body's local frame and rotates with the body every simulation step. Use this for body-fixed thrusters or actuators. |
```
eholum-nasa: This got removed right?

eholum-nasa (commented):
The wrench implementation looks good to me! Sorry to be annoying about the PhysicsLoop. It took me a while to figure out what was going on under the hood (as I'm sure it did for you!!!). I think I'm selfishly protecting that code block until we can better define the relationship between the hardware interface and the underlying sim... if that makes sense...

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/external_wrench_plugin.hpp
saikishor: You are right. Fixing it

mujoco_ros2_control_plugins/test/test_external_wrench_plugin.cpp
```diff
@@ -0,0 +1,522 @@
...
+
+  plugin.update(model_, data_);
+
+  double total_one = 0.0;
+  for (int i = 0; i < model_->nbody * 6; ++i)
+  {
+    total_one += std::abs(data_->xfrc_applied[i]);
+  }
+
+  ASSERT_GT(total_one, 0.0) << "Single wrench should produce non-zero qfrc_applied";
```
eholum-nasa: Very nit!

```suggestion
  ASSERT_GT(total_one, 0.0) << "Single wrench should produce non-zero xfrc_applied";
```

mujoco_ros2_control/src/mujoco_system_interface.cpp
eholum-nasa: Talked offline about this! We will leave these in place for now and pull them into #175, and then have a planned change to refactor the physics loop in the newly created simulate class. So we will be refactoring this in the near future but the way this plugin works will give us something functional to refactor against!

eholum-nasa (dismissed):
Just the nit comment but I think we can move forward with the next two big things!

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: Yess!
we need a better way to handle the mj_data across the components

eholum-nasa (approved):
Again!

PR #170, opened 2026-04-21
Use pkg install dir for requirements in conversion script

eholum-nasa:
Otherwise it's looking in the lib when running the actual launch file.

```
[robot_description_to_mjcf.sh-1] ERROR: Could not open requirements file: [Errno 2] No such file or directory: '/opt/mujoco/ws/install/mujoco_ros2_control/lib/mujoco_ros2_control/requirements.txt'
[robot_description_to_mjcf.sh-1] Successfully installed dependencies in 0 seconds.
```

We should either do this or install `requirements.txt` into the lib, which feels wrong.


saikishor (approved):
This sounds like the right approach

PR #171, opened 2026-04-23
Handle duplicate mesh file names in the conversion tool

eholum-nasa:
We have simulations that are running multiple types of UR arms (ur5e + ur10e) which we want to use the conversion tool on. The problem is that UR has commonly named mesh files across their robots, https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/tree/rolling/meshes, so when converting we end up with duplicated asset names and the output gets overwritten.

To get around this, we propose adding a suffix when duplicate mesh files are found. We just add `_N` and increment as needed. I'm not sure if there's a better way of handling this that would actually work in all cases (like somehow deduping based on file structure), so this is pretty simple and solves it if using the conversion script.

Updated and did a little clean up of the tools docs while I was at it. Sorry to nitpick about markdown formatting... The only thing that actually changed in the docs is adding a note about the added underscore, updating the lidar example since it was out of date, and changing a reference to use a relative path.

### DIFF
--- a/mujoco_ros2_control/README.md
+++ b/mujoco_ros2_control/README.md
@@ -16,7 +16,7 @@ This includes adding actuators, sensors, and cameras as needed to the MJCF XML.
 
 This conversion can be done either offline or at run time.
 We have built a *highly experimental* tool to automate URDF conversion.
-For more information refer to the [documentation](https://github.com/ros-controls/mujoco_ros2_control/blob/main/mujoco_ros2_control/docs/TOOLS.md).
+For more information refer to the [documentation](./docs/TOOLS.md).
 
 ## Hardware Interface Setup
 
--- a/mujoco_ros2_control/README.md
+++ b/mujoco_ros2_control/README.md
@@ -375,7 +375,7 @@ The lidar sensor is then configurable through ROS 2 control xacro with:
 
 ```xml
     <!-- Lidar sensors are matched to a set of rangefinder sensors in the MJCF, which should be -->
-    <!-- generated with the "<lidar>" inputs in the conversion script. The consiste of N rangefinders -->
+    <!-- generated with the "<lidar>" inputs in the conversion script. The consists of N rangefinders -->
     <!-- and will generally be of the form "<sensor_name>-01". These parameters should line up with -->
     <!-- those documented in the laserscan message. -->
     <sensor name="lidar">
--- a/mujoco_ros2_control/README.md
+++ b/mujoco_ros2_control/README.md
@@ -375,19 +375,18 @@ The lidar sensor is then configurable through ROS 2 control xacro with:
 
 ```xml
     <!-- Lidar sensors are matched to a set of rangefinder sensors in the MJCF, which should be -->
-    <!-- generated with "replicate" and will generally be of the form "<sensor_name>-01". -->
-    <!-- We assume the lidar sensor starts at angle 0, increments by the specified `angle_increment`, and -->
-    <!-- that there are exactly `num_rangefinders` all named from <sensor_name>-000 to the max -->
-    <!-- <sensor_name>-<num_rangefinders> -->
+    <!-- generated with the "<lidar>" inputs in the conversion script. The consiste of N rangefinders -->
+    <!-- and will generally be of the form "<sensor_name>-01". These parameters should line up with -->
+    <!-- those documented in the laserscan message. -->
     <sensor name="lidar">
       <param name="frame_name">lidar_sensor_frame</param>
       <param name="angle_increment">0.025</param>
-      <param name="num_rangefinders">12</param>
+      <param name="min_angle">-0.3</param>
+      <param name="max_angle">0.3</param>
       <param name="range_min">0.05</param>
       <param name="range_max">10</param>
       <param name="laserscan_topic">/scan</param>
     </sensor>
-  </ros2_control>
 ```
 
 ## Simulation - Topics and Services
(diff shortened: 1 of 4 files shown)


eholum-nasa:
@esheetz-nasa tagging you here if you have the time to take a look?

esheetz-nasa:
I don't have permissions to review, but this looks good to me! :rocket:

christophfroehlich:
> I don't have permissions to review, but this looks good to me! :rocket: 

are you sure? everyone should be able to review from the "changes" tab

esheetz-nasa (approved):

esheetz-nasa:
> > I don't have permissions to review, but this looks good to me! 🚀
> 
> are you sure? everyone should be able to review from the "changes" tab

You're right, thank you!

mujoco_ros2_control/README.md
```diff
@@ -375,19 +375,18 @@ The lidar sensor is then configurable through ROS 2 control xacro with:
 
 ```xml
     <!-- Lidar sensors are matched to a set of rangefinder sensors in the MJCF, which should be -->
-    <!-- generated with "replicate" and will generally be of the form "<sensor_name>-01". -->
-    <!-- We assume the lidar sensor starts at angle 0, increments by the specified `angle_increment`, and -->
-    <!-- that there are exactly `num_rangefinders` all named from <sensor_name>-000 to the max -->
-    <!-- <sensor_name>-<num_rangefinders> -->
+    <!-- generated with the "<lidar>" inputs in the conversion script. The consiste of N rangefinders -->
```
saikishor:
```suggestion
    <!-- generated with the "<lidar>" inputs in the conversion script. The consists of N rangefinders -->
```

mujoco_ros2_control/docs/TOOLS.md
```diff
@@ -55,7 +56,7 @@ options:
                         If we should convert .stls to .objs
   -s, --save_only       Save files permanently on disk; without this flag, files go to a temporary directory
   -f, --add_free_joint  Adds a free joint before the root link of the robot in the urdf before conversion
-  --fuse, --no-fuse     Allows MuJoCo to merge static bodies. Use --no-fuse to prevent merging (Default: --fuse).
+  --fuse, --no-fuse     Allows MuJoCo to merge static bodies. Use --no-fuse to prevent merging.
```
saikishor: Why removing the mention of Default here?

mujoco_ros2_control/docs/TOOLS.md
```diff
@@ -80,21 +81,22 @@ The `/tmp/output/` directory will contain all necessary assets and MJCF files th
...
-/opt/ros/${ROS_DISTRO}/opt/mujoco_vendor/bin/simulate /tmp/output/mujoco_description_formatted.xml
+ros2 run mujoco_vendor simulate /tmp/output/mujoco_description_formatted.xml
 ```
 
 Of note, the test robot has a good chunk of supported functionality, and we recommend using it as a guide.
 
 > [!NOTE]
-> The `make_mjcf_from_robot_description.py` script requires `trimesh`, `mujoco`, and `obj2mjcf`. These must either be installed system-wide or available within a virtual environment that is sourced before running the command.
-
+> The `make_mjcf_from_robot_description.py` script requires `trimesh`, `mujoco`, and `obj2mjcf`.
```
saikishor:
```suggestion
> The `make_mjcf_from_robot_description.py` script requires `trimesh`, `mujoco`, `coacd`, and `obj2mjcf`.
```

mujoco_ros2_control/docs/TOOLS.md
```diff
@@ -106,6 +108,8 @@ Of note, the test robot has a good chunk of supported functionality, and we reco
 * reads a robot description URDF
 * add in mujoco tag that provides necessary info for conversion
 * replace package names from `package://` to absolute filepaths
+  * NOTE: duplicate mesh files will have an `_N` appended to them to avoid conflicts in the output `assets` folder
+  * E.g. if running multiple types of UR robots in one sim, there will be multiple `shoulder.dae` files
```
saikishor:
```suggestion
  * NOTE: duplicate mesh files will have an `_N` appended to them to avoid conflicts in the output `assets` folder. For instance, if running multiple types of UR robots in one sim, there will be multiple `shoulder.dae` files
```
Both the bullet points are related, why not in a single one?

mujoco_ros2_control/docs/TOOLS.md
eholum-nasa: oh theliterally just a copy/paste of the latest output from the help menu. Idk why the default got removed from the text... I'll update that in the script launcher.

mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
```diff
@@ -85,8 +85,30 @@ def remove_tag(xml_string, tag_to_remove):
...
+    Builds a dictionary of all unique visual meshes in the URDF and rewrites the
+    raw_xml so every mesh filename points at a disambiguated path. There are some
+    gotchas:
+
+    - Different URDF meshes can share a filename stem, so colliding stems get an
+      "__N" suffix so each resolves to its own asset directory. For example, if
+      running multiple UR types there will be a shoulder/ and shoulder__1/ directory.
+    - This happens BEFORE the pregen lookup. So if the asset_dir already
+      contains shoulder/ and shoulder__1/ from a prior run, the second source
+      will grab shoulder__1.
```
saikishor:
```suggestion
      contains shoulder/ and shoulder__1/ from a prior run, the first source
      will grab shoulder and the second source will grab shoulder__1.
```

saikishor (dismissed):
Rest LGTM

eholum-nasa:
Thanks @saikishor and @esheetz-nasa! Will wait for the runners to come back to life and then merge this in...

saikishor (approved):

Issue #168, opened 2026-04-14
Propose refactoring existing sensors as plugins


eholum-nasa:
I think this could go a long way towards simplifying the hardware interface. Now that we support plugins, I think the cameras and the lidar sensor would be reasonably straightforward to pull out of the base hardware interface and remade as plugins. The force/torque sensor and IMUs would be trickier, as they rely on exporting state interfaces. Though maybe it would be worth modifying the plugin interface to support that?

Opening this just as a place to discuss this.

eholum-nasa:
On the plugin refactoring note, given #165 I think getting a tighter yet still extensible definition in place for plugins would make sense. Right now there is so much ambiguity over `mj_data_control_` and what fields are available to be modified and what aren't. I think I would proposed updating the plugin interface to not allow users to modify the data ad hoc, and then add a structure that holds the "allowable" plugin contributions? 

Maybe something like:

```
struct PluginData
{
  std::vector<mjtNum> xfrc_applied;  // could essentially replace xfrc_plugin_desired_
  std::vector<mjtNum> qfrc_applied;  // or whatever other data fields we would want to stack...

  // And then we just have some kind of initialize and reset function....
  void allocate(const mjModel* m) { ... }
  void reset() { ... }
};
```

Then the plugins would get a read only copy of _all_ the data, but we could actually define how the simulation interacts with the contributions from the plugins. Something like:

```
class MuJoCoROS2ControlPluginBase {

  // No mods of the data allowed
  virtual bool init(rclcpp::Node::SharedPtr node, const mjModel* model, const mjData* data) = 0;

  // Still no mods of the mj_data allowed, but use plugin_data if there are inputs back to the sim
  virtual void update(const mjModel* model, const mjData* sim_data, PluginData& plugin_data) = 0;

  // cleanup is the same
};
```

Edit: @saikishor brings up the excellent point that mjData has all kinds of raw pointers inside of it, so the best we can do is _strongly_ recommend that plugins not change things. Maybe something like `mjModel* ro_model, mjData* ro_sim_data`. Something to think about...

eholum-nasa:
Cameras and lidar are now fully implemented as plugins.

I think the interface is pretty stable at this point. The FTS/IMU/Pose sensors all export state interfaces, which we aren't planning on enabling through plugins at all in the near future.

Calling this done.

PR #175, opened 2026-04-27
Refactor the core mujoco simulation to be in its own container

eholum-nasa:
I've had this on my wishlist for a while and using #168 as an excuse to get started on it. Apologies for the huge PR, hopefully it is 99% non-functional changes. Nothing here should change _anything_ w.r.t. the existing API. All ros params, services, topics, etc should be the same, and no existing classes should require changes.\

I would recommend just reviewing the changes in the header files to start, as that tells you what has moved where.

The goal is to separate what I think of as the "core mujoco simulation" and what is the "ros2_control hardware interface".  This PR starts that process by moving what I think of as the core simulation into its own class, called `MujocoSimulate`. In summary the goals are to:

* Move the physics loop to the wrapper class
  * This pulls in the model and data objects used in stepping the underlying simulation
  * Everything is still just exposed to the consuming hw interface through their raw pointers (as it was previously), this includes the relevant mjData containers.
* Move the mj::Simulate to the wrapper class
  * The included the UI rendering logic and threads, along with the "headless" renderer
* Pull over the minimal set of ros constructs that go along with those things
  * The clock publisher is owned by MujocoSimulate, which makes sense to me since it is a core part of the sim
  * The services for interacting with sim progression, include pause, reset, and step

Everything else should be left as is in the hardware interface. All the changes in the `MujocoSystemInterface` should just be deletions (or renaming entries that are now live in the wrapper).

The next task will be cleaning up the relationship between `mj_data_` and `mj_data_control_` in the MujocoSimulate class and the consumers of that data.

### DIFF
--- a/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
+++ b/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
@@ -88,7 +88,7 @@ public:
    * @brief Construct the Simulate application and start the UI thread (or HeadlessAdapter).
    *
    * This initializes the Simulate app and starts the UI thread in the background (if not
-   * running headless).
+   * running headless). It also sets up required publishers and services using the provided node.
    */
   bool initialize(rclcpp::Node::SharedPtr node, const std::string& model_path, const std::string& mujoco_model_topic,
                   double sim_speed_factor, bool headless);
@@ -122,11 +122,6 @@ public:
    * @brief Register a callback function to be called on `reset_world_state`.
    */
   void set_reset_callback(ResetCallback callback);
-
-  /**
-   * @brief Create the reset/pause/step services on the node passed to initialize().
-   */
-  void create_services();
 
   /**
    * @brief Start the physics thread. Must be called after load_model().
@@ -169,27 +164,6 @@ public:
   {
     return *sim_mutex_;
   }
-
-  /**
-   * @brief Accessor for actual Simulate application.
-   */
-  mujoco::Simulate* simulate() const
-  {
-    return sim_.get();
-  }
-
-  /**
-   * @brief Whether or not the simulation is being run in headless mode.
-   */
-  bool headless() const
-  {
-    return headless_;
-  }
-
-  // Snapshot accessors used by plugins/tests.
-  void get_model(mjModel*& dest);
-  void get_data(mjData*& dest);
-  void set_data(mjData* mj_data);
 
   /**
    * @brief Reset simulation state (qpos/qvel/ctrl/sensors/forces) to the captured initial state.
--- a/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
+++ b/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
@@ -1,0 +1,297 @@
+/**
+ * Copyright (c) 2026, United States Government, as represented by the
+ * Administrator of the National Aeronautics and Space Administration.
+ *
+ * All rights reserved.
+ *
+ * This software is licensed under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <atomic>
+#include <condition_variable>
+#include <functional>
+#include <memory>
+#include <mutex>
+#include <string>
+#include <thread>
+#include <vector>
+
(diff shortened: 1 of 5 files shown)


eholum-nasa:
As a sanity check of the plugin merge I added the wrench to the demo robot and seems good!

Updated the config:

```
/**:
  ros__parameters:
    mujoco_plugins:
      heart_beat_plugin:
        type: "mujoco_ros2_control_plugins/HeartbeatPublisherPlugin"
        update_rate: 1.0
      external_wrench:
        type: "mujoco_ros2_control_plugins/ExternalWrenchPlugin"
        force_arrow_scale: 0.01      # 100 N  → 1 m arrow
        torque_arrow_scale: 0.1      # 10 N·m → 1 m arrow
```

Then launched ` ros2 launch mujoco_ros2_control_demos 01_basic_robot.launch.py` and called the wrench service with:

```
ros2 service call /external_wrench/apply_wrench mujoco_ros2_control_msgs/srv/ApplyExternalWrench "{
  wrenches: {
    external_wrenches: [
      {
        header: {stamp: {sec: 0, nanosec: 0}, frame_id: 'world'},
        wrench: {
          header: {stamp: {sec: 0, nanosec: 0}, frame_id: 'gripper_right_finger'},
          wrench: {force: {x: 450.0, y: 0.0, z: 0.0}, torque: {x: 0.0, y: 0.0, z: 0.0}}
        },
        application_point: {x: 0.0, y: 0.0, z: 0.0},
        duration: {sec: 5, nanosec: 0},
        ramp_down_duration: {sec: 0, nanosec: 0}
      }
    ]
  }
}"
```

https://github.com/user-attachments/assets/5f4feac1-4aa4-4a5b-b094-0f4e32746842

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
```diff
@@ -0,0 +1,304 @@
...
+                  double sim_speed_factor, bool headless);
+
+  /**
+   * @brief Load the MuJoCo model and compile it.
+   *
+   * Loads the MJCF as a string either from disk, if a path is specified in `model_path`. If
+   * no model is found on disk, then we listen for a string topic. This additionally allocates
+   * both the data and the control data once the model is loaded.
+   */
+  bool load_model();
```
saikishor: Do we need this method?. Won't the above initialize do the same job?

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
```diff
@@ -0,0 +1,304 @@
...
+
+  /**
+   * @brief Stop the physics and UI threads if they are running.
+   */
+  void shutdown();
+
+  /**
+   * @brief Accessor for the mujoco model.
+   */
+  mjModel* model() const
```
saikishor:
```suggestion
  mjModel* model()
```

In this case, we should mark them as non-const as the internal data could change?

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
```diff
@@ -0,0 +1,304 @@
...
+  {
+    return mj_model_;
+  }
+
+  /**
+   * @brief Accessor for the mujoco data.
+   */
+  mjData* data() const
```
saikishor:
```suggestion
  mjData* data()
```

Same here?

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
```diff
@@ -0,0 +1,304 @@
...
+  {
+    return mj_data_;
+  }
+
+  /**
+   * @brief Accessor for the mujoco control data.
+   */
+  mjData* control_data() const
```
saikishor:
```suggestion
  mjData* control_data()
```

Same argument

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
```diff
@@ -0,0 +1,304 @@
...
+  {
+    return mj_data_control_;
+  }
+
+  /**
+   * @brief Accessor for the mutex which locks access to the data and model.
+   */
+  std::recursive_mutex& mutex() const
+  {
+    return *sim_mutex_;
```
saikishor, line 158:
```suggestion
    RCLCPP_WARN_EXPRESSION(logger_, sim_mutex_ == nullptr,  "Sim recursive mutex is still nullptr");
    return *sim_mutex_;
```

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
eholum-nasa: oh yeah I think that might break some use cases... good point!

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
eholum-nasa: Good point... I think in my head they were separate steps but moved it into `initialize()`.

saikishor (approved):
Tested my setup with these changes as well as the master and both work like a charm

Issue #178, opened 2026-04-29
CI Broken on rolling from missing rosdeps


eholum-nasa:
Just some hoopla with the noble sync and prep for resolute raccoon.

On rolling:

```
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
mujoco_ros2_control_plugins: No definition of [ros2_control_cmake] for OS version [noble]
mujoco_ros2_control_demos: No definition of [xacro] for OS version [noble]
mujoco_ros2_control_tests: No definition of [xacro] for OS version [noble]
mujoco_ros2_control: No definition of [transmission_interface] for OS version [noble]
```

Going to disable that build but opening this to remember to bring it back...

christophfroehlich:
rolling-testing should be fine, rolling-main needs a sync on the buildfarm

eholum-nasa:
> rolling-main needs a sync on the buildfarm

Still an issue?

https://github.com/ros-controls/mujoco_ros2_control/actions/runs/25866079536/job/76008419996?pr=194#step:5:2563

Ts-sound:
After some investigation, I identified the root cause.

The upstream project's `rosdistro` Rolling repository has been updated to:
```yaml
ubuntu:
  - resolute
```
(see: https://github.com/ros/rosdistro/blame/master/rolling/distribution.yaml)

However, the `ros:rolling` Docker images are still based on `noble` (not yet updated). Related issue: https://github.com/osrf/docker_images/issues/878

As a result, `rosdep` cannot install packages properly due to system mismatch, leading to build failures.

Related issues:
- https://github.com/ros-controls/ros2_control_ci/issues/821
- https://github.com/ros-controls/mujoco_ros2_control/issues/203
- https://github.com/ros-controls/mujoco_ros2_control/issues/178

Ts-sound:
@christophfroehlich @eholum-nasa

christophfroehlich:
the rosdep failures from the initial post are related to your description, but 
https://github.com/ros-controls/ros2_control_ci/issues/821
https://github.com/ros-controls/mujoco_ros2_control/issues/203
aren't. I use handcrafted docker images, and they use resolute raccoon already.

eholum-nasa:
Merged in #229

PR #179, opened 2026-04-29
Temp disable rolling build check

eholum-nasa:
Because of https://github.com/ros-controls/mujoco_ros2_control/issues/178

Resolving:

https://github.com/ros-controls/mujoco_ros2_control/actions/runs/25112185073/job/73595284152#step:5:407

Until deps are in place. I'd like to have the functional distros running while working on #175!

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -50,7 +50,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     source /opt/ros/${ROS_DISTRO}/setup.bash && \
     apt-get update && \
     rosdep update && \
-    rosdep install -iry --from-paths src/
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -50,7 +50,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     source /opt/ros/${ROS_DISTRO}/setup.bash && \
     apt-get update && \
     rosdep update && \
-    rosdep install -iry --from-paths src/
```
eholum, line 53: Might as well fail here rather than at compile time... I guess this has been here for a while!

saikishor (approved):

PR #182, opened 2026-05-01
Improve LiDAR validation in URDF-to-MJCF conversion

skywoodsz:
This PR improves validation for LiDAR handling in the URDF-to-MJCF conversion pipeline.

Currently, camera handling records matched sites and raises an error when a referenced site cannot be found. However, LiDAR handling does not have the same validation behavior. If a LiDAR `ref_site` is misspelled or missing, `add_lidar_from_sites()` silently skips it, and the generated MJCF contains no LiDAR without any failure signal.

In addition, `angle_increment` is used in LiDAR angle processing, but invalid values such as `0` were not explicitly checked before use. This could lead to a division-by-zero error later in the conversion process. This PR adds validation to ensure `angle_increment` is greater than zero, so invalid LiDAR configurations fail early with a clearer error message.

## Changes

1. Raise an error for unmatched LiDAR sites

   If a referenced LiDAR site cannot be found, the converter now raises a clear error instead of silently continuing:

   ```python
   ValueError("Lidar site(s) not found ...")


2. Validate `angle_increment`

   Since `angle_increment` is used in LiDAR angle processing, setting it to `0` could previously lead to a division-by-zero error. This PR now catches that invalid configuration earlier by raising a clear `ValueError`:

   ```python
   ValueError("'angle_increment' must be greater than zero for a 'lidar' tag!")

### DIFF
--- a/mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
+++ b/mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
@@ -991,11 +991,13 @@ def add_lidar_from_sites(dom, lidar_dict):
     """
 
     x_form = [0.5, 0.5, 0.5, 0.5]  # pi/2 around x, pi/2 about y
+    matched_sites = set()
 
     # Construct all lidar sensor bodies for relevant sites in xml and add them as children to the same parent
     for node in dom.getElementsByTagName("site"):
         site_name = node.getAttribute("name")
         if site_name in lidar_dict:
+            matched_sites.add(site_name)
             replicate = lidar_dict[site_name]
 
             # Handle conversion of the frames by applying the site transform, rangefinder transform, then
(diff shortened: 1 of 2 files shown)


eholum-nasa (commented):
Thanks, @skywoodsz!

Can you update the `test_add_lidar_from_sites_no_matching_sites` from the conversion tests accordingly? 

https://github.com/ros-controls/mujoco_ros2_control/blob/main/mujoco_ros2_control/tests/test_urdf_to_mujoco_utils.py#L968

I think just having it assert it catches the value error would make sense?

saikishor (commented):
Can you add tests to the newly added codebase handling both valid and invalid usecases

skywoodsz:
@saikishor @eholum-nasa 

Thanks for the review!

I have updated the tests accordingly:

- Added `test_get_processed_mujoco_inputs_lidar_zero_angle_increment`
  - Verifies that `angle_increment="0"` raises a `ValueError`
  - Checks that the error message contains `angle_increment`

- Updated `test_add_lidar_from_sites_no_matching_sites`
  - Previously, it asserted that no LiDAR body was generated
  - Now it asserts that an unmatched LiDAR site raises a `ValueError`
  - Checks that the error message contains `other_site`

eholum (approved):
Thanks, @skywoodsz !

As an aside our group is looking into an alternative, and hopefully significantly faster version of Lidar sensors, https://github.com/ros-controls/mujoco_ros2_control/issues/187. Just letting you know since it seems like you're using them!

eholum-nasa (approved):
Mixing accounts here...

PR #188, opened 2026-05-05
Migrate documentation to rst format for control.ros.org

saikishor:
I've migrated the documentation to the RST format for the control.ros.org rendering.


eholum-nasa (approved):
Artifacts _looked_ good to me! I added a pixi workflow to build these in isolation but can add that in a follow-on.

I'm assuming pretty much everything was a copy-paste / reformatting, and you didn't really update the content? Assuming that is the case, ship it.

saikishor:
@eholum-nasa Yes, it is a reformatted content to fix the RST style.

PR #189, opened 2026-05-06
Display the sim real time factor on the native viewer

saikishor:
Sometimes I realized that my sim is running slower due to complex meshes and it was just a sensation and now adding this info really helps to understand when something goes wrong

image
image
image

### DIFF
--- a/mujoco_ros2_control/src/mujoco_simulation.cpp
+++ b/mujoco_ros2_control/src/mujoco_simulation.cpp
@@ -1274,10 +1274,24 @@ void MujocoSimulation::update_sim_display()
     return;  // render thread hasn't consumed the last update yet, skip
   }
 
+  // Desired speed: from parameter override when set, otherwise from the UI slider.
+  const double desired_pct = sim_speed_factor_ < 0 ? static_cast<double>(sim_->percentRealTime[sim_->real_time_index]) :
+                                                     sim_speed_factor_ * 100.0;
+
+  // Actual speed: measured_slowdown = CPU_time / sim_time, so speed = 100 / slowdown.
+  // Falls back to 0 while paused (slowdown stays at its last measured value but we
+  // suppress the display when there is no forward progress).
+  const double actual_pct = sim_->run ? (100.0 / sim_->measured_slowdown) : 0.0;
+
+  char desired_buf[16], actual_buf[16];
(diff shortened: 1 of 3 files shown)


mujoco_ros2_control/src/mujoco_simulation.cpp
eholum-nasa, line 1063: Ah it's been a while since I've poked at this, but something to note is that this parameter was originally added to _speed the sim up_. If it is not set, then speed is delegated to the UI and users can slow the sim down lower than 100% in the normal "simulate" way.

So not sure how much we care, but we'd probably want a similar conditional in the sim display? Otherwise this might be confusing... where I have slowed the sim down to 40% from the UI but the widget still reports 100%.

Image

Tangentially, the sim app generally shows me when it's not hitting the target rate, does it not do that for you?

Image

eholum-nasa: edit: oh it's only when it is greater than 10% or so.

In any case, I like this but need to figure out how to get it to play nicely with requested speeds from the UI!

saikishor: Sure. I'll try to figure it out🫡

Thank for the nice info. I learn from you day to day 😃

saikishor: @eholum-nasa How are you setting the speeds from the UI?, I'm not able to do it with +/-. Which keys are you using?

eholum-nasa: oh gosh sorry, you have to omit the `sim_speed_factor` arg, which will delegate the speed determination to the `Simulate` application. I _think_ if you set it the plus and minus keys get ignored... iirc we don't set these for the demos should it should just work with `ros2 launch mujoco_ros2_control_demos 01_basic_robot.launch.py`.

saikishor:
image
image


Changing sim speed factor:

image

image

mujoco_ros2_control/src/mujoco_simulation.cpp
saikishor: Done

eholum-nasa (approved):
Thanks for the fix! WFM and I like that this would be easier to customize if we needed it.

PR #191, opened 2026-05-11
Control and plugin data input cleanup

eholum-nasa:
Working towards #168.

The focus of this PR is to clean up how `mj_data_control_` is used between the base physics simulation in `MujocoSimulation` and the hardware interface in `MujocoSystemInterface`. As of now, the simulation _must_ provide three things in a thread-safe manner to both the system interface and plugins,

1. Reading all fields from `mj_data_` / `mj_model_`
2. Adding control inputs to `mj_data_->ctrl` and `mj_data_->qfrc_applied`
3. Applying Cartesian forces to a body through `mj_data_->xfrc_applied`

`mj_data_control_` is now the primary mechanism for 1 and 2. I added a new mechanism `PluginData` for 3, which basically takes the place of `xfrc_plugin_desired_`. I lean towards wanting to expose that buffer / swapping ugliness out to the plugins, just because it keeps that data in an accessible location rather than getting overwritten by whatever happens in the simulate application. I put it in a struct, just for extensibility, though realistically given the other control inputs in `mjData`, maybe it never expands past it? I am very pro-externally facing buffer for the data, but less tied to the struct itself... just some minor future-proofing.

Regardless, the new order of operations from the ros2_control interface:

1. `read()` - Locks the sim, deep copies `mj_data_` into `mj_data_control_` for providing state information
2. `update()` - Controllers can use `mj_data_control_` to update their values
3. `write()`
  * commands are written into `mj_data_control_` (`ctrl` and `qfrc_applied`)
  * Plugins are `updated` given them access to control and state variables prior to sending to the sim
  * Control data is copied from `mj_data_control_` into `mj_data_`

The most significant changes are to the plugin interface. First, the `update(...)` call now happens at the end of `write`, rather than `read`. I think this is critical and probably a long standing bug... because previously any values set by plugins in `mj_data_control_` were being immediately overwritten by the controllers. This way, plugins get access to the data immediately before it is fired down to the "hw interface" and can adjust as needed knowing the values won't be touched. 

Other than that, this shouldn't change anything about how external users interact with the system interface. However, there are possibly significant gains both for performance and readability,

1. The `physics_loop` is no longer doing extraneous `mj_copyData` calls. Once the hardware interface updates the control inputs, they can be used directly.
2. Removed mutex locking from the control loop excepting when data is explicitly moved between the control and sim data containers. We're close to getting the sim mutex entirely contained within `MujocoSimulation`, which would be a big win for clarity, IMO.
3. Fixed a LONG STANDING BUG in qfrc_applied copy. The old physics loop used mj_model_->nu instead of mj_model_->nv when copying qfrc_applied, which has size nv. This silently copied the wrong number of elements. Not sure how we ever missed that...

Apologies for the wall of text.

TL;DR - The physics loop is way cleaner and the control thread now fully owns management of control data. Existing tests provide pretty decent coverage of the changes. And the only difference users should expect to see slightly improved performance of the physics loop.

### DIFF
--- a/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
+++ b/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
@@ -28,17 +28,19 @@
 #include <thread>
 #include <vector>
 
-#include <mujoco/mujoco.h>
-
-#include <mujoco_ros2_control_msgs/srv/reset_world.hpp>
-#include <mujoco_ros2_control_msgs/srv/set_pause.hpp>
-#include <mujoco_ros2_control_msgs/srv/step_simulation.hpp>
 #include <rclcpp/rclcpp.hpp>
 #include <realtime_tools/realtime_publisher.hpp>
 #include <rosgraph_msgs/msg/clock.hpp>
 
 #include "glfw_adapter.h"  // for mj::GlfwAdapter
 #include "simulate.h"      // must be on your include path, handled by CMake
+
+#include <mujoco/mujoco.h>
+
+#include <mujoco_ros2_control/data.hpp>
+#include <mujoco_ros2_control_msgs/srv/reset_world.hpp>
+#include <mujoco_ros2_control_msgs/srv/set_pause.hpp>
+#include <mujoco_ros2_control_msgs/srv/step_simulation.hpp>
 
 namespace mujoco_ros2_control
 {
@@ -150,22 +152,11 @@ public:
   }
 
   /**
-   * @brief Accessor for the mutex which locks access to the data and model.
+   * @brief Accessor for the plugin_data.
    */
-  std::recursive_mutex& mutex() const
+  PluginData& plugin_data()
   {
-    RCLCPP_WARN_EXPRESSION(logger_, sim_mutex_ == nullptr, "Sim recursive mutex is still nullptr");
-    return *sim_mutex_;
-  }
-
-  /**
-   * @brief Accessor for the stacking applied forces, these values will be added to `xfrc_applied`.
-   *
-   * TODO: Remove this and provide consisted access for write-able mujoco data.
-   */
-  std::vector<mjtNum>& xfrc_plugin_desired()
-  {
-    return xfrc_plugin_desired_;
+    return plugin_data_;
   }
 
   /**
@@ -173,6 +164,41 @@ public:
    * @note Caller must hold the sim mutex.
    */
   void reset_world_state(bool fill_initial_state);
+
+  /**
+   * @brief Copies `mj_model_` into the provided container in a thread safe way.
+   *
+   * This locks the sim mutex and will pause the physics loop, so should be used sparingly.
+   */
+  void copy_mj_model(mjModel* destination);
+
+  /**
+   * @brief Copies `mj_data_` into the provided container in a thread safe way.
+   *
+   * This locks the sim mutex and will pause the physics loop, so should be used sparingly.
+   */
+  void copy_mj_data(mjData* destination);
+
+  /**
+   * @brief Copies the provided mjData into mj_data_ om a thread safe way.
+   */
+  void set_mj_data(mjData* source);
+
+  /**
+   * @brief Copies relevant control data into the sim data in a thread safe way.
+   *
+   * This locks the sim mutex and will pause the physics loop, so should be used sparingly.
+   */
+  void copy_control_data();
+
+  /**
+   * @brief Accessor for the mutex which locks access to the data and model.
+   */
+  std::recursive_mutex& mutex() const
+  {
+    RCLCPP_WARN_EXPRESSION(logger_, sim_mutex_ == nullptr, "Sim recursive mutex is still nullptr");
+    return *sim_mutex_;
+  }
 
 private:
   /**
@@ -218,29 +244,14 @@ private:
   mjModel* mj_model_{ nullptr };
   mjData* mj_data_{ nullptr };
 
-  // Data container for control data. The physics data is copied into this container every
-  // controller loop. This is where the controller should add commands.
+  // Data container for control data. The physics data from `mj_data_` is copied into this
+  // on every loop. The external interface can read as needed from this information.
+  // This is where the controller should add commands for `ctrl` and
+  // `qfrc_applied`.
   mjData* mj_data_control_{ nullptr };
 
-  // TODO: Conslidate the control and these buffer to provide consistent, clear access for
-  //       for mujoco data.
-  //
-  // Dedicated buffer for plugin xfrc contributions.
-  //
-  // mj_copyData contaminates mj_data_control_->xfrc_applied with viewer forces, so we cannot
-  // rely on mj_data_control_->xfrc_applied to hold plugin forces across physics iterations.
-  // Instead, the control thread (read()) zeroes mj_data_control_->xfrc_applied before every
-  // plugin update, lets plugins write fresh forces, then copies the result here.  The physics
-  // thread uses this buffer (not mj_data_control_->xfrc_applied) when composing each mj_step.
-  // Because this buffer is never touched by mj_copyData, it holds the last plugin contribution
-  // cleanly until the next control cycle — no restore and no undo mechanism required.
-  std::vector<mjtNum> xfrc_viewer_capture_;  ///< viewer-only forces (drag), used per inner step
-  std::vector<mjtNum> xfrc_plugin_desired_;  ///< plugin forces, set once per control cycle
-  /// Last value written to mj_data_->xfrc_applied by the physics loop (viewer + plugin).
-  /// Used at the start of each outer iteration to detect whether the render thread ran
-  /// (and zeroed/re-applied drag) since the last physics step.  If mj_data_->xfrc_applied
-  /// matches this buffer, the render thread did not run and xfrc_viewer_capture_ is still valid.
-  std::vector<mjtNum> xfrc_last_restore_;
+  // Data structure to provide access to control and force inputs from plugins.
+  PluginData plugin_data_;
 
   // For rendering
   mjvCamera cam_;
--- a/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
+++ b/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
@@ -57,6 +57,21 @@ namespace mujoco_ros2_control
  * underlying simulation - including publishing simulated time to /clock, as well as services
  * for pausing, stepping, and resetting the simulation.
  *
+ * Importantly, the physics loop is intended to run at whatever speed (relative realtime) is
+ * requested by the user. It is important to not interrupt the loop with locking calls that
+ * interact with either the physics sim data, `mj_data_`,  or the model, `mj_model_`.
+ * Instead, consumers of this class are provided with functions to read all sim data and provide
+ * control inputs.
+ *
+ * `copy_mj_data` will lock the sim and do a full copy of the existing `mj_data_` into the
+ * provided container, which can be used as the caller requires.
+ *
+ * `update_control_data` will copy control inputs from `mj_data_control_` in `mj_data_`, add is
+ * the recommended way to send control commands to the underlying sim.
+ *
+ * Lastly, `plugin_data_` is provided to update control inputs or applied forces from plugin
+ * interfaces in a standalone, well structured manner.
+ *
  * Thread safety is still somewhat messy, as callers are provided with a simulation mutex that
  * locks the model and data while the actual mujoco engine moves the sim forward. Callers
  * need to be wary of locking that mutex external to this class, as it can have significant
--- a/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
+++ b/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
@@ -87,8 +87,6 @@ public:
    * Triggered by the ~/reset_world service or by a any other reset detected in the physics
    * loop. The callback will be run with the sim mutex but after all data has been restored.
    *
-   * // TODO: Retire this when mujoco data is split.
-   *
    * @param fill_initial_state When true, the caller has not already populated
    *        mj_data_->qpos/qvel/ctrl from a keyframe and the callback should restore the captured
    *        initial state. When false, a keyframe has already been applied.
--- a/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
+++ b/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
@@ -136,7 +136,11 @@ public:
   }
 
   /**
-   * @brief Accessor for the mujoco data.
+   * @brief Accessor for the raw mujoco simulation data.
+   *
+   * Users should generally not interact with the physics sim data excepting during setup and
+   * other special circumstances. For "normal" processing it is recommended to use
+   * `control_data` or other containers populated by `copy_mj_data`.
    */
   mjData* data()
   {
@@ -145,6 +149,9 @@ public:
 
   /**
    * @brief Accessor for the mujoco control data.
+   *
+   * Standard container for both providing control inputs through `ctrl` and `qfrc_applied`, along
+   * with access to the underlying mujoco simulation's data.
    */
   mjData* control_data()
   {
@@ -176,20 +183,24 @@ public:
    * @brief Copies `mj_data_` into the provided container in a thread safe way.
    *
    * This locks the sim mutex and will pause the physics loop, so should be used sparingly.
+   * @note If `destination` is empty a new mjData will be allocated.
    */
   void copy_mj_data(mjData* destination);
 
   /**
-   * @brief Copies the provided mjData into mj_data_ om a thread safe way.
+   * @brief Copies the provided mjData into mj_data_ in a thread safe way.
+   *
+   * @note This will completely overwrite the existing data, use with caution!
    */
   void set_mj_data(mjData* source);
 
   /**
-   * @brief Copies relevant control data into the sim data in a thread safe way.
+   * @brief Copies control fields from `mj_data_control_` into the sim data in a thread safe way.
    *
-   * This locks the sim mutex and will pause the physics loop, so should be used sparingly.
+   * Specifically, copies `mj_data_control_->ctrl` and `mj_data_control_->qfrc_applied` into
+   * `mj_data_` to update the hw control inputs for the next iteration of the physics loop.
    */
-  void copy_control_data();
+  void update_control_data();
 
   /**
    * @brief Accessor for the mutex which locks access to the data and model.
@@ -242,12 +253,12 @@ private:
   // MuJoCo data pointers, these are the primary containers used by the physics simulation.
   // It is generally not recommended to interact with them directly.
   mjModel* mj_model_{ nullptr };
+
+  // Primary data container for the physics loop. We do not recommend interacting with this
+  // directly unless you are sure of what you are doing.
   mjData* mj_data_{ nullptr };
 
-  // Data container for control data. The physics data from `mj_data_` is copied into this
-  // on every loop. The external interface can read as needed from this information.
-  // This is where the controller should add commands for `ctrl` and
-  // `qfrc_applied`.
+  // This container provides both state information and control inputs for the system interface.
   mjData* mj_data_control_{ nullptr };
 
   // Data structure to provide access to control and force inputs from plugins.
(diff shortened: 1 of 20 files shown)


msavchen-nasa:
Still catching up on the full context of this PR, so I apologize for any silly questions in advance

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
```diff
@@ -28,18 +28,20 @@
...
 #include <rclcpp/rclcpp.hpp>
 #include <realtime_tools/realtime_publisher.hpp>
 #include <rosgraph_msgs/msg/clock.hpp>
 
 #include "glfw_adapter.h"  // for mj::GlfwAdapter
 #include "simulate.h"      // must be on your include path, handled by CMake
 
+#include <mujoco/mujoco.h>
+
+#include <mujoco_ros2_control_msgs/srv/reset_world.hpp>
```
msavchen-nasa, line 40: Is it necessary for these services to be within the `simulation.hpp` or can they be part of the mujoco_ros2_control_plugins?

mujoco_ros2_control/src/mujoco_simulation.cpp
```diff
@@ -1214,20 +1183,17 @@ void MujocoSimulation::physics_loop()
...
-
             // run mj_forward, to update rendering and joint sliders
             mj_forward(mj_model_, mj_data_);
             sim_->speed_changed = true;
             update_sim_display();
           }
         }
 
+        // Restore plugin xfrc into mj_data_ so mjv_updateScene shows force arrows
+        // in the native viewer (it reads xfrc_applied before zeroing it).
```
msavchen-nasa: Im having trouble parsing this comment.
`mju_addTo` combines the forces from the plugin data with mj_data but it doesnt zero out the plugin data's forces itself. 
I think that is done in the `write` method of `MuJocoSystemInfrace`...

mujoco_ros2_control/src/mujoco_simulation.cpp
```diff
@@ -1214,20 +1183,17 @@ void MujocoSimulation::physics_loop()
...
+        mju_addTo(mj_data_->xfrc_applied, plugin_data_->xfrc_applied.data(), 6 * mj_model_->nbody);
```
msavchen-nasa: Why are we adding the force multiple times in the same physics loop.
If we are running (unpaused) we are performing the the same calculation on line [1051](https://github.com/ros-controls/mujoco_ros2_control/pull/191/changes#diff-2da00f09a30eeaf87c64df6eb0315d7e53226c20d9235494bcef538359ce2a18R1051)

In the case of the external wrench plugin this would result the applied wrench being twice as big, right?

msavchen-nasa: Why are we adding the force multiple times in the same physics loop.
If we are running (unpaused) we are performing the the same calculation on line [1051](https://github.com/ros-controls/mujoco_ros2_control/pull/191/changes#diff-2da00f09a30eeaf87c64df6eb0315d7e53226c20d9235494bcef538359ce2a18R1051)

mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
```diff
@@ -147,8 +133,8 @@ void ExternalWrenchPlugin::update(const mjModel* /*model_arg*/, mjData* data)
     const int base = w.body_id * 6;
     for (int j = 0; j < 3; ++j)
     {
-      data->xfrc_applied[base + j] += force_world[j];
-      data->xfrc_applied[base + 3 + j] += torque_at_xipos[j];
+      plugin_data->xfrc_applied[base + j] += force_world[j];
```
msavchen-nasa: This makes me a bit nervous,
We are relying on the user to make sure that they add their forces rather than setting them directly.
I think this would be a place where we would track errors to.
This design is still a vast improvement and separation of dealing with mj_data_ directly.

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
```diff
@@ -21,6 +21,41 @@
...
+ *
+ * For now, we include `xfrc_applied` as a mechanism to inject external forces directly into the physics
+ * simulation, while providing rendering capabilities that work in tandem with the Simulate app's
+ * force rendering arrows. Of note, these buffers exist separately from `mj_data_` and `mj_data_control_`
+ * as the Simulate's render thread's `Sync()` function can zero out applied Cartesian forces, which
+ * otherwise complicates determining desired forces from external plugins.
+ *
+ * @param xfrc_applied Applied Cartesian force/torque (nbody x 6)
+ */
+struct PluginData
```
msavchen-nasa: `PluginData` is extensible, but currently it provides access to xfrc_applied. 
Does it make sense to require all mujoco_control_plugins use this struct?
`HeartBeatPublisherPlugin` for example doesnt use it at all....

msavchen-nasa: `PluginData` is extensible, but currently it provides access to xfrc_applied. 
Does it make sense to require all mujoco_control_plugins use this struct?
`HeartBeatPublisherPlugin` for example doesnt use it at all....

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
```diff
@@ -48,14 +83,23 @@ class MuJoCoROS2ControlPluginBase
...
+   *       and qfrc_applied take priority. Changes are copied to the physics data before
+   *       the next mj_step.
    * @note This method will be called at the end of the mujoco_ros2_control read loop, before the update loop of
    * controllers and the write loop. This means that changes to the data here will be visible to controllers and will
    * affect the next simulation step.
    * @note This method will be called in a real-time thread, so it should avoid blocking operations and should be
    * efficient.
    */
-  virtual void update(const mjModel* model, mjData* data) = 0;
+  virtual void update(const mjModel* model, mjData* data, std::shared_ptr<PluginData> plugin_data) = 0;
```
msavchen-nasa: Pointer mjData is still being passed to the plugin, 
the user can potentially just use the mjData to access the xfrc_applied instead of using the plugin data right?

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
eholum-nasa: IMO they are fundamental in terms of how they interact with the simulation in `physics_loop` so I put them in here. That is, they drastically impact the progression of the simulation and need access to the lowest levels of data so I think it makes sense to have them as part of this bottom layer ROS interface.

The intention of `mujoco_ros2_control_plugins` is to provide external consumers access to the data and a subset of inputs, so it would be a little messy to move them there!

mujoco_ros2_control/src/mujoco_simulation.cpp
eholum-nasa: Ah yes this is very, very deep lore from https://github.com/ros-controls/mujoco_ros2_control/pull/165.

Basically, the `Simulate` application (which is the UI and what we are wrapping up) has the ability to apply forces to bodies using the mouse. So it will occasionally zero out the entries in `mj_data_->xfrc_applied` in the event that drag forces are being applied from a different location. It also handles rendering those applied forces as arrows in the UI (which is very neat).

We provide ways for plugins to modify those forces as well, but now there are two possible mechanisms by which `mj_data_->xfrc_applied` can be modified! It's a bit of a headache but that is basically why we have to re-apply them every single loop here.

Before we had a triple buffer mechanism,

```
  std::vector<mjtNum> xfrc_viewer_capture_; 
  std::vector<mjtNum> xfrc_plugin_desired_; 
  std::vector<mjtNum> xfrc_last_restore_;
```

to track updates from both sources and merge them.

For this change, I separated out how plugins have access to `xfrc_applied` by putting it into `PluginData`. This way they won't step on each other toes, and we just have to applies those directly in this loop, as before.

mujoco_ros2_control/src/mujoco_simulation.cpp
eholum-nasa: Ah good catch! I think one of these is where it shouldn't be... I _think_...

Basically in the paused case the force restoration was in every single conditional block, so I pulled it out here, though I believe what is on line 1155 should get moved into the step conditional. I'll double check...

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
eholum-nasa: Well, not all plugins are expected to use all of the data that is provided. The heartbeat plugin is just an example, for instance!

But that said, I still have mixed feelings about this data structure. I do think that it is critical to better define how both the `MujocoSystemInterface` and external plugins interact with both `mjData` AND this odd side object that has competing threads writing to it. This is just one option to deal with it that I hope clarifies things, but also drastically simplifies the internals of the physics sim.

I just don't know what other invisible entries in mjData we might be mucking with or might break in the future. So another BIG piece of the motivation is to define a plugin interface that won't break the next time we want to add another entry to this structure. Does that make sense?

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
eholum-nasa: yep, they certainly can! but there's no guarantee of results if they go that route so that's in their hands!

mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
eholum-nasa: Yeah the plugins are kind of the Wild West / use at your own risk. Just trying to make them as clear as possible...

eholum-nasa (commented):
Thanks for jumping in @msavchen-nasa ! Appreciate the questions and will take a deeper look here. Let me know what makes sense and what doesn't.

mujoco_ros2_control/src/mujoco_simulation.cpp
eholum-nasa: Ok I don't really want to think about that odd "optimization" for the sake of this PR so choosing to leave as is...

Edit: blah - now I'm back to the original problem of needing to keep plugin contributions separated... this will need more tweaking :/

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
```diff
@@ -213,32 +265,24 @@ class MujocoSimulation
...
-  /// (and zeroed/re-applied drag) since the last physics step.  If mj_data_->xfrc_applied
-  /// matches this buffer, the render thread did not run and xfrc_viewer_capture_ is still valid.
-  std::vector<mjtNum> xfrc_last_restore_;
+  // Data structure to provide access to control and force inputs from plugins.
+  std::shared_ptr<mujoco_ros2_control_plugins::PluginData> plugin_data_;
+
+  // Buffers to track actively applied Cartesian forces from both the plugins and the Simulate /
+  // viewer-only drag forces.
+  std::vector<mjtNum> xfrc_viewer_capture_;  // Tracks forces from the viewer
+  std::vector<mjtNum> xfrc_last_written_;    // tracks the last value written to xfrc_applied
```
eholum-nasa, line 273: Someday, if it ever becomes necessary, it might be worth completely forking the simulate application just to change that small bit of code! Until then I can't figure any way to get away from requiring three buffers.

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
```diff
@@ -143,36 +164,67 @@ class MujocoSimulation
...
+   * @brief Copies `mj_model_` into the provided container in a thread safe way.
    *
-   * TODO: Remove this and provide consisted access for write-able mujoco data.
+   * This locks the sim mutex and will pause the physics loop, so should be used sparingly.
    */
-  std::vector<mjtNum>& xfrc_plugin_desired()
-  {
-    return xfrc_plugin_desired_;
-  }
+  mjModel* copy_mj_model(mjModel* destination);
```
saikishor:
```suggestion
  void copy_mj_model(mjModel* destination);
```

Do we need to return the same pointer again?

mujoco_ros2_control/src/mujoco_simulation.cpp
```diff
@@ -922,6 +933,31 @@ void MujocoSimulation::step_simulation_callback(
   }
 }
 
+mjModel* MujocoSimulation::copy_mj_model(mjModel* destination)
+{
+  const std::unique_lock<std::recursive_mutex> lock(*sim_mutex_);
+  return mj_copyModel(destination, mj_model_);
```
```suggestion
void MujocoSimulation::copy_mj_model(mjModel* destination)
{
  const std::unique_lock<std::recursive_mutex> lock(*sim_mutex_);
  mj_copyModel(destination, mj_model_);
```

mujoco_ros2_control/src/mujoco_simulation.cpp
```diff
@@ -982,24 +1012,23 @@ void MujocoSimulation::physics_loop()
...
-        // the render thread has not run since our last step, so the viewer forces are unchanged
-        // and xfrc_viewer_capture_ remains valid.  If it differs, the render thread ran: it
-        // zeroed xfrc_applied and called mjv_applyPerturbForce, so mj_data_->xfrc_applied now
-        // holds only drag forces — use that as the new viewer capture.
-        const int nbody6 = 6 * mj_model_->nbody;
-        if (std::memcmp(mj_data_->xfrc_applied, xfrc_last_restore_.data(), nbody6 * sizeof(mjtNum)) != 0)
+        // record it in xfrc_last_written_. We can then combine the desired forces from the plugins
+        // as well as the viewers prior to stepping, without either of them stacking in
+        // undesirable ways.
+        const uint32_t nbody6 = 6 * mj_model_->nbody;
```
```suggestion
        const auto nbody6 = 6 * mj_model_->nbody;
```

saikishor (commented):
I'm still a bit skeptic about some part of the approach. I'll give it a thought again and come back to you. 2 minor nitpicks in the meantime ;)

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
eholum-nasa: The short answer is no it does not _need_ to. If `destination` is null [mj_copyModel](https://mujoco.readthedocs.io/en/stable/APIreference/APIfunctions.html#mj-copymodel) will allocate and return a pointer to the new model. I'll just update the signatures to pass a reference to the pointer, too.

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
msavchen-nasa: Yeah, I definitely see the difficulty in that...

eholum-nasa:
@saikishor I think I got to the conclusion that the plugin data structure wasn't worth it, so I removed in 80c631f930c83b13a033c9f76c5e94a88fe10181. If we ever want to resurrect it it can live there...

I re-added `xfrc_plugin_desired_` whose use is hopefully a little cleaner because all control data is updated in a single place. Let me know if that is cleaner/makes more sense to you.

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
```diff
@@ -143,36 +162,61 @@ class MujocoSimulation
...
+   * @brief Copies `mj_model_` into the provided container in a thread safe way.
    *
-   * TODO: Remove this and provide consisted access for write-able mujoco data.
+   * This locks the sim mutex and will pause the physics loop, so should be used sparingly.
    */
-  std::vector<mjtNum>& xfrc_plugin_desired()
-  {
-    return xfrc_plugin_desired_;
-  }
+  void copy_mj_model(mjModel*& destination);
```
ndunkelb-nasa: Could these be `get_mj_model` and `get_mj_data`? Just to be a bit more clear. I think `copy` would normally have a destination and a source, and doesn't really imply that you are grabbing it from the member variable.

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
```diff
@@ -143,36 +162,61 @@ class MujocoSimulation
...
+
+  /**
+   * @brief Copies control fields from `mj_data_control_` into the sim data in a thread safe way.
+   *
+   * Specifically, copies `mj_data_control_->ctrl` and `mj_data_control_->qfrc_applied` into
+   * `mj_data_` to update the hw control inputs for the next iteration of the physics loop.
+   * `mj_data_control_->xfrc_applied` is copied into `xfrc_plugin_desired_` to avoid conflicts
+   * from the simulate app.
+   */
+  void update_control_data();
```
ndunkelb-nasa: Similarly, I think this could be set_control_data()

mujoco_ros2_control/src/mujoco_simulation.cpp
```diff
@@ -1172,10 +1189,7 @@ void MujocoSimulation::physics_loop()
           // (try_publish) has time to flush between steps, matching play mode behavior.
           if (pending_steps_.load() > 0)
           {
-            mju_copy(mj_data_->ctrl, mj_data_control_->ctrl, static_cast<int>(mj_model_->nu));
-            mju_copy(mj_data_->qfrc_applied, mj_data_control_->qfrc_applied, static_cast<int>(mj_model_->nu));
-
-            // TODO: Cleanup when mujoco data is split
+            // Add the viewer and plugin applied forces
             mju_copy(mj_data_->xfrc_applied, xfrc_viewer_capture_.data(), nbody6);
             mju_addTo(mj_data_->xfrc_applied, xfrc_plugin_desired_.data(), nbody6);
```
ndunkelb-nasa: I guess I a not as deep in the weeds here, but is there a reason these stayed in this loop whereas other stuff got moved out of this loop?

ndunkelb-nasa (commented):
I'm still going through, but mostly a couple of minor changes and curious about some clarification. I'm not sure I understand fully why everything needs to be done exactly where it is, but I agree with the idea to separate things. I think the main part I still want to understand is - anything that stayed in the simulation loop itself, why does it need to be there instead of being handled outside?

mujoco_ros2_control/src/mujoco_simulation.cpp
eholum-nasa: Yeah this is the deep woods. Basically because `xfrc_applied` can be written two from two locations plugins/hw interface AND the `Simulate` app (like clicking and dragging arrow). That function can also zero out applied forces. We need to make sure that the values are applied at every iteration of the physics_loop to ensure that they are consistently applied, so they stay inside the loop. This is in contrast to the other control values in `mj_data_`, which are only copied over above.

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
eholum-nasa: FWIW these are explicitly running `copy` functions under the hood, not just returning pointers or a value. So real work is happening. We could name it `copy_physics_model` or `copy_physics_data`?

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
eholum-nasa: Yeah, I wonder if this even needs to maintain the pointer to `mj_data_control_`, maybe that is the consumers responsibility...

eholum-nasa: eh, no, the reset logic gets real fuzzy with that. I think I'll leave it but will update these function names.

eholum-nasa: Update: I changed my mind. I think it makes total sense for the things that interact with the sim to own their data! Moved it out of the `MujocoSimulation` class and back to the hw interface. Hopefully that creates a _very clear separation_ of responsibilities. The Simulation just updates the physics data in `mj_data_`, consumers can interact with it from their own data containers.

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
eholum-nasa: updated. and hopefully the args make it more clear...

saikishor (approved):
LGTM

PR #197, opened 2026-05-20
Support/cameras rendering/headless

msavchen-nasa:
Resolves #21.

Building on top of https://github.com/ros-controls/mujoco_ros2_control/pull/107.


mujoco_ros2_control/README.md
```diff
@@ -11,42 +11,495 @@ Parts of this library are also based on the MoveIt [mujoco_ros2_control](https:/
...
 
-The simulator exposes ROS 2 topics and services for interacting with the simulation at runtime (pause, reset, step-by-step control, etc.).
-See the [simulation topics and services documentation](docs/hardware_interface.rst#simulation-topics-and-services) for details.
+This allows camera topics to be published even when running in headless mode (e.g., on a server, in Docker containers, or in CI environments).
+
+> [!NOTE]
+> EGL requires proper GPU drivers and EGL libraries to be installed (e.g., `libegl1-mesa` on Ubuntu).
+> If both GLFW and EGL fail to initialize, camera publishing will be disabled with a warning.
+
+
```
saikishor: This part of the documentation should go here: https://github.com/ros-controls/mujoco_ros2_control/blob/main/mujoco_ros2_control/docs/hardware_interface.rst

msavchen-nasa: thank you!

mujoco_ros2_control/docs/hardware_interface.rst
```diff
@@ -287,6 +287,21 @@ Frame and topic names can be overridden via ``ros2_control`` xacro:
 
    MuJoCo's camera coordinate conventions differ from ROS. Refer to the MuJoCo documentation for details.
 
+Headless Rendering
+~~~~~~~~~~~~~~~~~~
+
+Camera rendering is supported in headless environments (without a display). The system automatically detects whether a display is available:
```
eholum-nasa:
```suggestion
Camera rendering is supported in headless environments (without a display).
The system automatically detects whether a display is available:
```

docker/Dockerfile
```diff
@@ -35,7 +35,15 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
...
-      vim
+      vim \
+      libegl1 \
+      libgbm1 \
+      libosmesa6 \
+      libosmesa6-dev
+
+# Enable software rendering with LLVMpipe for headless environments (CI, Docker without GPU)
+ENV LIBGL_ALWAYS_SOFTWARE=1
+ENV GALLIUM_DRIVER=llvmpipe
```
eholum-nasa: Duplicated by CI run, I would omit here

```suggestion
```

mujoco_ros2_control/src/mujoco_cameras.cpp
```diff
@@ -150,15 +166,171 @@ void MujocoCameras::close()
...
+  {
+    RCLCPP_ERROR(node_->get_logger(), "EGL: Failed to get display");
+    return false;
+  }
+
+  // Initialize EGL
+  EGLint major, minor;
+  if (!eglInitialize(egl_display_, &major, &minor))
+  {
+    RCLCPP_ERROR(node_->get_logger(), "EGL: Failed to initialize (error: 0x%x)", eglGetError());
```
eholum-nasa, line 173: Interesting, failing in the pixi build here:

https://github.com/ros-controls/mujoco_ros2_control/actions/runs/26166897599/job/76973442345#step:6:2032

.github/workflows/ci.yaml
```diff
@@ -106,6 +108,8 @@ jobs:
             -v ${{ github.workspace }}/.sccache:/workspace/.sccache \
             -w /workspace \
             -e SCCACHE_DIR=/workspace/.sccache \
+            -e LIBGL_ALWAYS_SOFTWARE=1 \
+            -e GALLIUM_DRIVER=llvmpipe \
```
eholum-nasa: Building and running this locally egl is still failing to get a context, I think this should be

```suggestion
            -e EGL_PLATFORM=surfaceless \
```

https://mesa-docs.readthedocs.io/en/latest/egl.html, or https://registry.khronos.org/EGL/extensions/MESA/EGL_MESA_platform_surfaceless.txt,

>  The platform's intended use case is for enabling OpenGL and OpenGL ES     applications on systems where no window system exists. However, the     platform's permitted usage is not restricted to this case.  Since the     platform is independent of any native window system, it may also be used on     systems where a window system is present.

Maybe worth it on the docker side too?

docker/Dockerfile.pixi
```diff
@@ -8,11 +8,20 @@ RUN apt-get update && apt-get install -y \
...
+    libosmesa6 \
+    libosmesa6-dev \
     && rm -rf /var/lib/apt/lists/*
 
 RUN curl -fsSL https://pixi.sh/install.sh | bash
 ENV PATH="/root/.pixi/bin:${PATH}"
 
+# Enable software rendering with LLVMpipe for headless environments (CI, Docker without GPU)
+ENV LIBGL_ALWAYS_SOFTWARE=1
+ENV GALLIUM_DRIVER=llvmpipe
```
eholum-nasa: Same as above

```suggestion
```

docker/Dockerfile.pixi
```diff
@@ -8,11 +8,20 @@ RUN apt-get update && apt-get install -y \
     git \
     build-essential \
     ca-certificates \
+    libegl1 \
+    libgbm1 \
+    libgl1-mesa-dri \
+    libosmesa6 \
+    libosmesa6-dev \
```
eholum-nasa: in the build log these aren't actually being linked (noticed locally).

So given what egl can do with `surfaceless` do we even need the osmesa fallback?

mujoco_ros2_control/docs/hardware_interface.rst
```diff
@@ -287,6 +287,21 @@ Frame and topic names can be overridden via ``ros2_control`` xacro:
...
+
+* With display: Uses GLFW for OpenGL context creation (default behavior)
+* Without display: Falls back to EGL for GPU-accelerated headless rendering
+
+This allows camera topics to be published even when running in headless mode (e.g., on a server, in Docker containers, or in CI environments).
+
+.. note::
+   EGL requires proper GPU drivers and EGL libraries to be installed (e.g., libegl1-mesa on Ubuntu).
+   If both GLFW and EGL fail to initialize, camera publishing will be disabled with a warning.
```
eholum-nasa, line 303: I wonder if there's a way we could figure this out before starting the rendering thread... seems hard.

.github/workflows/ci.yaml
msavchen-nasa: Yup, managed to replicate this in my setup as well. The only problem is that `Coverage Build - Jazzy` still fails.

docker/Dockerfile.pixi
msavchen-nasa: technically no, I did see a couple of cases where OSMesa being used instead of EGL. I also tried making sure that these OSMesa could be found in the pixi environment but CMakeLists wasnt finding those environment.
Im still trying to map everything out for myself...

.github/workflows/ci.yaml
eholum-nasa: Same issue probably?

docker/Dockerfile.pixi
eholum-nasa: My vote is that if we don't _need_ it for headless environments then we shouldn't include it? Especially since it's not being pulled in now. If it offers something else that is functionally necessary then maybe we should re-evaluate? I only briefly read through the mesa docs and those are quite limited, so if `surfaceless` is insufficient or otherwise borked in some way that would be worth knowing...

.github/workflows/ci.yaml
msavchen-nasa: Yes, basically.

mujoco_ros2_control/src/mujoco_cameras.cpp
msavchen-nasa: This is resolved changing how we get display.

mujoco_ros2_control/docs/hardware_interface.rst
```diff
@@ -287,6 +287,22 @@ Frame and topic names can be overridden via ``ros2_control`` xacro:
...
    MuJoCo's camera coordinate conventions differ from ROS. Refer to the MuJoCo documentation for details.
 
+Headless Rendering
+~~~~~~~~~~~~~~~~~~
+
+Camera rendering is supported in headless environments (without a display).
+The system automatically detects whether a display is available:
+
+* With display: Uses GLFW for OpenGL context creation (default behavior)
+* Without display: Falls back to EGL for GPU-accelerated headless rendering
```
eholum-nasa, line 297: So fancy... will this actually work with hw acceleration? Have you tested?

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_cameras.hpp
```diff
@@ -142,6 +144,23 @@ class MujocoCameras
   // Camera processing thread
   std::thread rendering_thread_;
   std::atomic_bool publish_images_;
+
+  // EGL context for headless rendering (used when GLFW is unavailable)
+  EGLDisplay egl_display_;
+  EGLContext egl_context_;
+  EGLSurface egl_surface_;
+  bool use_egl_;
```
eholum-nasa: Keep defaults here

```suggestion
  // EGL context for headless rendering (used when GLFW is unavailable)
  EGLDisplay egl_display_{ EGL_NO_DISPLAY };
  EGLContext egl_context_{ EGL_NO_CONTEXT };
  EGLSurface egl_surface_{ EGL_NO_SURFACE };
  bool use_egl_{ false };
```

mujoco_ros2_control/src/mujoco_cameras.cpp
```diff
@@ -34,6 +36,10 @@ MujocoCameras::MujocoCameras(rclcpp::Node::SharedPtr node, std::recursive_mutex*
   , mj_data_(mujoco_data)
   , mj_model_(mujoco_model)
   , camera_publish_rate_(camera_publish_rate)
+  , egl_display_(EGL_NO_DISPLAY)
+  , egl_context_(EGL_NO_CONTEXT)
+  , egl_surface_(EGL_NO_SURFACE)
+  , use_egl_(false)
```
eholum-nasa: Other defaults already live in the header

```suggestion
```

mujoco_ros2_control/src/mujoco_cameras.cpp
```diff
@@ -136,14 +142,20 @@ void MujocoCameras::init(GlfwInitFn glfw_init_fn)
...
+  else
+  {
+    RCLCPP_WARN(node_->get_logger(), "Failed to initialize GLFW. Attempting EGL for headless rendering.");
+    use_egl_ = true;
+    publish_images_ = true;
+    rendering_thread_ = std::thread(&MujocoCameras::update_loop, this);
   }
-  publish_images_ = true;
-  rendering_thread_ = std::thread(&MujocoCameras::update_loop, this);
 }
```
eholum-nasa: Duplicates

```suggestion
  }
  else
  {
    RCLCPP_WARN(node_->get_logger(), "Failed to initialize GLFW. Attempting EGL for headless rendering.");
    use_egl_ = true;
    publish_images_ = true;
}
rendering_thread_ = std::thread(&MujocoCameras::update_loop, this);
}
```

mujoco_ros2_control/src/mujoco_cameras.cpp
```diff
@@ -155,15 +167,144 @@ void MujocoCameras::close()
...
+  {
+    RCLCPP_ERROR(node_->get_logger(), "EGL: Failed to create context (error: 0x%x)", eglGetError());
+    eglTerminate(egl_display_);
+    egl_display_ = EGL_NO_DISPLAY;
+    return false;
+  }
+
+  // Create PBuffer surface for offscreen rendering
+  const EGLint pbuffer_attribs[] = { EGL_WIDTH, 1, EGL_HEIGHT, 1, EGL_NONE };
+  egl_surface_ = eglCreatePbufferSurface(egl_display_, egl_config, pbuffer_attribs);
```
eholum-nasa, line 228: Does this break in the surfaceless case or is it just empty work?

mujoco_ros2_control/src/mujoco_cameras.cpp
```diff
@@ -155,15 +167,144 @@ void MujocoCameras::close()
...
+
+  // Initialize EGL
+  EGLint major, minor;
+  if (!eglInitialize(egl_display_, &major, &minor))
+  {
+    RCLCPP_ERROR(node_->get_logger(), "EGL: Failed to initialize (error: 0x%x)", eglGetError());
+    return false;
+  }
+  RCLCPP_INFO(node_->get_logger(), "EGL: Initialized version %d.%d", major, minor);
+
```
eholum-nasa: Found this helpful

```suggestion
  const char* client_apis = eglQueryString(egl_display_, EGL_CLIENT_APIS);
  const char* vendor = eglQueryString(egl_display_, EGL_VENDOR);
  RCLCPP_INFO(node_->get_logger(), "EGL: Initialized version %d.%d", major, minor);
  RCLCPP_INFO(node_->get_logger(), "EGL: Vendor: %s, APIs: %s", vendor, client_apis);
```

eholum-nasa (commented):
Much headless. Very neat.

Just some minor nits and questions but confirmed this on my laptop, does it really connect with Nvidia drivers?

mujoco_ros2_control/docs/hardware_interface.rst
msavchen-nasa: ooo good call,
Yes, it does!

mujoco_ros2_control/src/mujoco_cameras.cpp
msavchen-nasa: It doesnt break with surfaceless, 
From what I understand we need to create a pbuffer because that is where the display is storing the pixel values when there is no screen available.

mujoco_ros2_control/src/mujoco_cameras.cpp
```diff
@@ -136,13 +138,18 @@ void MujocoCameras::init(GlfwInitFn glfw_init_fn)
...
+  }
+  else
+  {
+    RCLCPP_WARN(node_->get_logger(), "Failed to initialize GLFW. Attempting EGL for headless rendering.");
+    use_egl_ = true;
+    publish_images_ = true;
   }
-  publish_images_ = true;
   rendering_thread_ = std::thread(&MujocoCameras::update_loop, this);
 }
```
eholum-nasa: Yeah my bad for the half suggestion, but we always publish!

```suggestion
  }
  else
  {
    RCLCPP_WARN(node_->get_logger(), "Failed to initialize GLFW. Attempting EGL for headless rendering.");
    use_egl_ = true;
  }
  publish_images_ = true;
  rendering_thread_ = std::thread(&MujocoCameras::update_loop, this);
}
```

eholum-nasa (approved):

PR #200, opened 2026-05-24
Fix glfw initialization on headless run

### LINKED ISSUE #198: MujocoCameras::init() starts OpenGL rendering thread even when model has no cameras (breaks headless / CI)
**Summary**

In headless:=true, MujocoSystemInterface::on_activate() always calls cameras_->init(). MujocoCameras::init() starts an offscreen GLFW/OpenGL thread when glfwInit() succeeds, even if cameras_ is empty (no MJCF <camera> elements and no camera sensors in ros2_control).

That thread calls mjr_makeContext() / gladLoadGL, which can fail on headless or CI hosts without a usable OpenGL stack. The process then aborts and ros2_control_node exits after hardware activation.

Models with no MuJoCo cameras should not need OpenGL for physics and joint control in headless mode.

**Steps to reproduce**
- Use a regression test added to branch https://github.com/Juliaj/mujoco_ros2_control/tree/test-headless-activate-no-cameras-repro, commit [2e23bca](https://github.com/Juliaj/mujoco_ros2_control/commit/2e23bca5237a0372f46d1cf57012c2ee56976680)
- Run HeadlessInitTest.HeadlessActivateWithoutCameras (e.g. on an X11 machine with the repro branch).
```
colcon test --packages-select mujoco_ros2_control --ctest-args -R test_headless_init
```
- Observe crash after activation: 
```
[INFO] ... Initializing rendering for cameras
[INFO] ... Starting the camera rendering loop, publishing at 5.000000 Hz
ERROR: gladLoadGL error
terminate called after throwing ... rclcpp::exceptions::RCLError
  what(): failed to publish message: cannot publish data ...
```

***Expected behavior***

With no MJCF cameras and no camera sensors, MujocoCameras::init() should not create a GLFW window or OpenGL context. headless:=true should run simulation and ros2_control without a display or OpenGL 3.3+ (typical CI).

**Cause**

MujocoSystemInterface::on_activate() always calls cameras_->init() (and lidar_sensors_->init()). register_cameras() only fills cameras_ from mj_model_->ncam; an empty list is valid.
MujocoCameras::init() does not check cameras_.empty() before starting the offscreen GLFW thread.
Headless mode skips the viewer via HeadlessAdapter but still starts camera offscreen rendering when there are no cameras.

- Fix candidate, on the repro branch, commit [9540dcf](https://github.com/Juliaj/mujoco_ros2_control/commit/9540dcf8d26005d6e5daddbc74cc731488206af5)

saikishor:
Fixes: #198

### DIFF
--- a/mujoco_ros2_control/src/mujoco_cameras.cpp
+++ b/mujoco_ros2_control/src/mujoco_cameras.cpp
@@ -128,10 +128,10 @@ void MujocoCameras::register_cameras(const hardware_interface::HardwareInfo& har
   }
 }
 
-void MujocoCameras::init()
+void MujocoCameras::init(GlfwInitFn glfw_init_fn)
 {
   // Start the rendering thread process
-  if (!glfwInit())
+  if (!glfw_init_fn())
   {
     RCLCPP_WARN(node_->get_logger(), "Failed to initialize GLFW. Disabling camera publishing.");
     publish_images_ = false;
--- a/mujoco_ros2_control/src/mujoco_cameras.cpp
+++ b/mujoco_ros2_control/src/mujoco_cameras.cpp
@@ -130,6 +130,11 @@ void MujocoCameras::register_cameras(const hardware_interface::HardwareInfo& har
 
 void MujocoCameras::init(GlfwInitFn glfw_init_fn)
 {
+  if (cameras_.empty())
+  {
+    return;
+  }
+
   // Start the rendering thread process
   if (!glfw_init_fn())
   {
--- a/mujoco_ros2_control/src/mujoco_lidar.cpp
+++ b/mujoco_ros2_control/src/mujoco_lidar.cpp
@@ -211,6 +211,11 @@ bool MujocoLidar::register_lidar(const hardware_interface::HardwareInfo& hardwar
 
 void MujocoLidar::init()
 {
+  if (lidar_sensors_.empty())
+  {
+    return;
+  }
+
   // Start the rendering thread process
   publish_lidar_ = true;
   rendering_thread_ = std::thread(&MujocoLidar::update_loop, this);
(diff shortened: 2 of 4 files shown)


mujoco_ros2_control/src/mujoco_lidar.cpp
```diff
@@ -211,6 +211,11 @@ bool MujocoLidar::register_lidar(const hardware_interface::HardwareInfo& hardwar
 
 void MujocoLidar::init()
 {
+  if (lidar_sensors_.empty())
+  {
+    return;
```
eholum-nasa, line 216: ah good call.

mujoco_ros2_control/tests/test_headless_init.cpp
```diff
@@ -154,6 +157,111 @@ TEST_F(HeadlessInitTest, HeadlessInitialization)
...
+#else
+  hardware_interface::HardwareComponentInterfaceParams params;
+  params.hardware_info = info;
+  auto init_result = interface_->on_init(params);
+#endif
+  ASSERT_EQ(init_result, hardware_interface::CallbackReturn::SUCCESS);
+
+  // Wait for model and data to be available.
+  auto start = std::chrono::steady_clock::now();
+  auto timeout = std::chrono::seconds(5);
```
eholum-nasa: Does this take 5 seconds? Only asking because the test above is just 1s and unsure if we've seen CI failures from it.

mujoco_ros2_control/src/mujoco_cameras.cpp
```diff
@@ -128,10 +128,15 @@ void MujocoCameras::register_cameras(const hardware_interface::HardwareInfo& har
   }
 }
 
-void MujocoCameras::init()
+void MujocoCameras::init(GlfwInitFn glfw_init_fn)
```
eholum-nasa, line 131: Relevance to #197, FYI @msavchen-nasa

eholum-nasa (dismissed):
LGTM! Just some minor questions.

FWIW a handful of changes coming down the pipeline here too.

mujoco_ros2_control/tests/test_headless_init.cpp
saikishor: Good one, changing it

eholum-nasa (approved):
LGTM (assuming tests are good!)

I like the conditionals on no cameras being added... I think this might end up being kind of a stop-gap until headless rendering lands and we make cameras a plugin.

PR #202, opened 2026-05-25
Add changes for mujoco vendor bump

saikishor:
The recent changes introduced in MuJoCo 3.4.0+, they have changed the types of some of the variables, and this is making our code fail to compile to to the COnversion warnings set to error, we will need these changes to work on both the existing version  of mujoco_vendor - 3.4.0 and also to bump it to 3.8.1

### DIFF
--- a/mujoco_ros2_control/package.xml
+++ b/mujoco_ros2_control/package.xml
@@ -24,6 +24,7 @@
   <depend>controller_manager</depend>
   <depend>hardware_interface</depend>
   <depend>fmt</depend>
+  <depend>ros2pkg</depend>
   <depend>nav_msgs</depend>
   <depend>libglfw3-dev</depend>
   <depend>mujoco_vendor</depend>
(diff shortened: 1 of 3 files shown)


eholum-nasa (commented):
Ah sorry I broke it...

Out of curiosity what is the process for the bump? Do we need to support both 3.4.0 _and_ 3.8.*? Or can we bump both the source and mujoco vendor prior to the release and just make a note that folks will need a newer version of the vendor package manually installed until the release?

saikishor:
> Ah sorry I broke it...
> 
> Out of curiosity what is the process for the bump? Do we need to support both 3.4.0 _and_ 3.8.*? Or can we bump both the source and mujoco vendor prior to the release and just make a note that folks will need a newer version of the vendor package manually installed until the release?

The problem is the release will fail as one of its dependency fail to  build. So, it is better this way, so we can release the mujoco_ros2_control first and then later bump mujoco_vendor. If not, they might rollback

eholum-nasa (approved):

PR #204, opened 2026-05-26
Isolate flakey functional tests and update pixi

### LINKED ISSUE #169: [CI] `test_set_pause_and_step_simulation` is flaky
**Describe the bug**
```
2026-04-19T04:38:01.3483252Z     -- run_test.py: verify result file '/home/runner/work/ros2_control_ci/ros2_control_ci/.work/target_ws/build/mujoco_ros2_control_tests/test_results/mujoco_ros2_control_tests/test_robot_launch_mjcf_generation_from_robot_description_test.py.xunit.xml'
2026-04-19T04:38:01.3484428Z   >>>
2026-04-19T04:38:01.3485456Z build/mujoco_ros2_control_tests/test_results/mujoco_ros2_control_tests/test_robot_launch_mjcf_generation_from_robot_description_test.py.xunit.xml: 13 tests, 0 errors, 1 failure, 1 skipped
2026-04-19T04:38:01.3486375Z - robot_launch_test.TestFixture test_set_pause_and_step_simulation
2026-04-19T04:38:01.3486688Z   <<< failure message
2026-04-19T04:38:01.3486889Z     Traceback (most recent call last):
2026-04-19T04:38:01.3494581Z       File "/home/runner/work/ros2_control_ci/ros2_control_ci/.work/target_ws/src/ros-controls/mujoco_ros2_control/mujoco_ros2_control_tests/test/robot_launch_test.py", line 539, in test_set_pause_and_step_simulation
2026-04-19T04:38:01.3495704Z         self.assertAlmostEqual(paused_clock_time, clock_time, delta=MUJOCO_TIMESTEP + 1e-7)
2026-04-19T04:38:01.3496330Z     AssertionError: 11.876 != 11.88 within 0.0020001 delta (0.004000000000001336 difference)
2026-04-19T04:38:01.3496759Z   >>>
```

**To Reproduce**
https://github.com/ros-controls/ros2_control_ci/issues/746

eholum-nasa:
Possibly resolves #169 (fingers crossed)?

Basically the reset world and pause unpause tests at the _functional_ level are troublesome. Though with the recent refactor these are more testable in isolation, so I rewrote (with AI's help) those tests to be directly at the `MujocoSimulation` level, rather than requiring all of the launch_ros infrastructure that normally comes with it. The logic is still basically the same, but hopefully reducing the burden on those functional tests will improve our CI pipelines.

While I was at it we need a pixi dep bump so doing it here.


mujoco_ros2_control/tests/test_mujoco_simluation.cpp
```diff
@@ -0,0 +1,365 @@
...
+  pause_req->paused = true;
+  auto pause_future = pause_client->async_send_request(pause_req);
+  ASSERT_EQ(pause_future.wait_for(std::chrono::seconds(5)), std::future_status::ready);
+  ASSERT_TRUE(pause_future.get()->success);
+
+  // Once paused, time should not advance
+  double prev = sim_->data()->time;
+  ASSERT_TRUE(wait_until([&]() {
+    double now = sim_->data()->time;
+    bool settled = (now == prev) && (now > 0.0);
```
ndunkelb-nasa: technically don't need the `now > 0` here bc we confirmed we had started before, right? Not that it really matters, just got a bit confused why that was important in this case. Also, is it possible this accidentally passes because on the first run, `prev` and `now` happen to both run before the next timestep?

mujoco_ros2_control/tests/test_mujoco_simluation.cpp
```diff
@@ -0,0 +1,365 @@
...
+  auto pause_req = std::make_shared<mujoco_ros2_control_msgs::srv::SetPause::Request>();
+  pause_req->paused = true;
+  auto pause_future = pause_client->async_send_request(pause_req);
+  ASSERT_EQ(pause_future.wait_for(std::chrono::seconds(5)), std::future_status::ready);
+  ASSERT_TRUE(pause_future.get()->success);
+
+  // Wait for pause to take effect
+  double prev = sim_->data()->time;
+  ASSERT_TRUE(wait_until([&]() {
+    double now = sim_->data()->time;
```
ndunkelb-nasa: Same potential question as above.

eholum-nasa: good shout, it's a redundant check... removed. Also removed some redundant clean up...

mujoco_ros2_control/tests/test_mujoco_simluation.cpp
ndunkelb-nasa: I saw you fixed the extra 0s, but @eholum-nasa , can you comment on my second statement? Not sure if I am thinking about it correctly.

eholum-nasa: > Also, is it possible this accidentally passes because on the first run, prev and now happen to both run before the next timestep?

Ah, I think it is _extremely_ unlikely but you are definitely right. There's no lock around `data()` because... reasons? So it is definitely possible that this grabs a duplicate timestamp during one execution of the `physics_loop`... Maybe we should use the copy mechanism to be sure of that... or make it actually settle like we did in the python version.

ndunkelb-nasa: I mean I think if you just read once, then wait 50ms or something, it should have plenty of time to step, and a check after that would be fine.

eholum-nasa: yeah definitely a lot of silly games to play, but we are keeping track of the step count in the physics loop! I just exposed that through a getter so that we can verify the sim is actually progressing and the time stays constant. That should be deterministic.

ndunkelb-nasa (approved):

PR #206, opened 2026-05-28
Re-add descoped ResetWorld functional test

eholum-nasa:
From talking offline with @saikishor, this was originally an important regression test against the expected behavior when calling reset world in combination with controller activation/deactivation.

The re-adds the test from #204, but removes the pausing checks and just trusts that portion is true, as that now has the lower level test to cover it. I also bumped up the joint deltas... it's a big delta but all we really care about is that the JTC is doing something close to right, and not just blowing up.


eholum-nasa:
Have run this a few times, @saikishor! It's scaled down but it should cover the movement cases pre/post reset that we talked about.

saikishor (approved):
LGTM. Thank you for readding it

PR #208, opened 2026-05-29
Add --cache-dir option for the URDF->MJCF generation script

saikishor:
I've added a `--cache-dir`because we are saving the output of the generated mjcf and right now modifying to add capsule in certain parts and we want the script to use the information automatically and publish the modified file instead of recreating it. 

I don't know if the --cache-dir is the right naming here hahah. I'm open to names


mujoco_ros2_control/docs/tools.rst
```diff
@@ -66,6 +66,34 @@ A complete list of options is available from the argument parser:
      -a ASSET_DIR, --asset_dir ASSET_DIR
                            Optionally pass an existing folder with pre-generated OBJ meshes. (default: None)
      --scene SCENE         Optionally pass an existing xml for the scene (default: None)
+     --cache-dir CACHE_DIR
+                           Optionally pass a directory containing a previously generated MJCF. If it already holds the necessary
+                           information, the conversion is skipped and the cached MJCF is read and published/copied directly
+                           instead of being regenerated. (default: None)
+
+A cache directory is considered complete when it contains a non-empty
```
eholum-nasa, line 74: Depending on the problem you guys are trying to solve, have you used `--assets_dir`? This is provided to reused existing converted mesh files to save on the conversions.

E.g. for CLR: https://github.com/NASA-JSC-Robotics/chonkur_l_raile/blob/jazzy-devel/clr_mujoco_config/launch/clr_mujoco.launch.py#L100, 

This has the advantage of still using the conversion, but not running `obj2mjcf` on anything that already has its mesh files in the provided `assets/` directory. So we can still add little things on the fly or tweak our URDFs, and end up with a changed scene without having to run everything.

Would that be sufficient for what you need? It doesn't have the converted `xml` but that portion is fast enough that we generally don't care.

Alternatively, the sim drivers also still supporting loading a pre-generated model from file, which we added a [conditional](https://github.com/NASA-JSC-Robotics/chonkur_l_raile/blob/jazzy-devel/clr_mujoco_config/launch/clr_mujoco.launch.py#L124) to our xacro to support. So I think one way to "cache" everything would be to use that and just point to the generated assets directory?

saikishor: @eholum-nasa the issue we are having is not at level of the `--asset-dir` for obj regeneration but more about the final modifications we had to do in the regenerated mjcf file. The main thing was having the capsules at the feet for our robots. 

Atleast this should help for the time being, unless we have a way to replace the MuJoCo collisions properly. 


The `--cache-dir` mainly looks for the existing mjcf file inside the folder

mujoco_ros2_control/docs/tools.rst
```diff
@@ -66,6 +66,34 @@ A complete list of options is available from the argument parser:
...
+``mujoco_description_formatted.xml`` together with the ``assets`` directory that the
+MJCF references (i.e. the output of a previous ``--save_only`` run). When ``--cache-dir``
+points at such a directory, the expensive mesh conversion / ``obj2mjcf`` step is skipped:
+the cached MJCF is published on ``--publish_topic`` and/or copied into ``--output`` when
```
eholum-nasa, line 78: That and I think I was thrown off by the wording - 

```suggestion
MJCF references (i.e. the output of a previous ``--save_only`` run). When ``--cache-dir``
points at such a directory, the conversion pipeline is skipped, and the previously generated
MJCF is published on ``--publish_topic`` and/or copied into ``--output`` when
```

mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
```diff
@@ -1222,6 +1225,73 @@ def get_xml_from_file(urdf_file=None):
...
+    the ``<compiler>`` tag's attribute values change) so the rest of the document is
+    preserved verbatim -- in particular the absence of an ``<?xml ?>`` declaration,
+    which MuJoCo rejects at the start of the description.
+
+    :param xml_string: the MJCF document as a string.
+    :param base_dir: directory the relative asset dirs are anchored to (typically the
+        cache / output directory).
+    :returns: the MJCF string with relative compiler asset dirs made absolute.
+    """
+    compiler_match = re.search(r"<compiler\b[^>]*>", xml_string)
```
eholum-nasa, line 1278: Will this catch things like (which some xml formatters do to us):

```xml
<compiler 
  meshdir="assets"
  angle="radian"
  autolimits="true"
/>
```

According to [AI this would do it](https://docs.python.org/3/library/re.html#re.DOTALL),

```
compiler_match = re.search(r"<compiler\b[^>]*>", xml_string, re.DOTALL)
```

mujoco_ros2_control/docs/tools.rst
eholum-nasa: Ah, I get it. And I assume you want it to be automatic so that you don't have to manually specify from a launch file (which is what we do)?

FWIW We basically save away our scenes with manual modifications, and then have a launch arg / xacro argument that configures the HW interface to load a scene file from disk, rather than publishing. Could something similar here if that would cover your use case? 

I just worry that this will be a little confusing, but maybe that's down to the name of `--cache-dir` and `--asset_dir` (which we should standardize those args underscore vs dash)...

`--asset_dir` uses regenerated assets if they are available to skip obj decomposition

`--cache-dir` basically says here's where everything lives just use that? If I'm reading this correctly? 

I think the concern, from a users perspective, is that it doesn't really check if the URDF has changed or if things should be regenerated (ie, if the cache is stale), but I'm not sure if there's a simple way to programmatically check that. 

All that to say maybe calling it something more specific like `--pregenerated_mjcf` might be more descriptive?

mujoco_ros2_control/docs/tools.rst
```diff
@@ -66,6 +66,34 @@ A complete list of options is available from the argument parser:
      -a ASSET_DIR, --asset_dir ASSET_DIR
                            Optionally pass an existing folder with pre-generated OBJ meshes. (default: None)
      --scene SCENE         Optionally pass an existing xml for the scene (default: None)
+     --cache-dir CACHE_DIR
```
eholum-nasa, line 69: I recognize this is weird but every other arg is doing it...

```suggestion
     --cache_dir CACHE_DIR
```

eholum-nasa (commented):
Yeah it makes sense! I think I was thrown off by the `--cache-dir` and the docs, but IMO renaming something that has less baggage that comes with the term would clarify it, does that make sense?

mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
saikishor: According to AI, this would work hahahha

mujoco_ros2_control/docs/tools.rst
saikishor: I understand that forwarding the args is better, but it is making our URDF complicated time to time, that's why I think it is better to add this arg for the timebeing until we ad support and then maybe we can remove it?

I tried to use the robot_state_publisher, but then it was creating other issues

saikishor:
I think it is fair to handle it at the user side instead of adding a new arg and confusing them. For now, I'll handle this on my end.

Issue #120, opened 2026-02-25
Current limitations of URDF -> MJCF converter


saikishor:
I'm creating this issue to document and collect the opinions of others who are having trouble generating the final MJCF from the URDF. The idea is to use it to improve the current converter

pang-yann:
Hi there, I am working to integrate the mujoco_ros2_control library in my work but am encountering some limitations with the URDF-to-MJCF converter in handling some of the points discussed in this issue. 

Are there any plans to extend the behaviour of the current converter to support multiple <mujoco_inputs> tags or selectively choosing visual or collision meshes?

saikishor:
> Hi there, I am working to integrate the mujoco_ros2_control library in my work but am encountering some limitations with the URDF-to-MJCF converter in handling some of the points discussed in this issue.
> 
> Are there any plans to extend the behaviour of the current converter to support multiple <mujoco_inputs> tags or selectively choosing visual or collision meshes?

Yes, there is some necessity or the need for this, We want to take a look at it. We should see how to approach this to fit all usecases.

ndunkelb-nasa:
@saikishor, do you have time to revisit this w.r.t. your [previous PR](https://github.com/ros-controls/mujoco_ros2_control/pull/209)? I want to make sure we can find a solution that meets what everyone is looking for. Specifically, I am worried about the issue that I called out here. 

>For example, the UR collision objects are giant and would make all of our collisions unrealistic in mujoco. I am fine defaulting to collision objects for collision because that is intuitive, but maybe we have another component in processed_inputs that can call out specific links to use the visual object instead?

Before going too far down an implementation rabbit hole for something a bit complicated, I wanted to see if we could come to some agreement of what the user-facing interface looks like, then jump into the change after that.

I am down to put in some time work on this as well, but wanted to make sure there was a plan first.

saikishor:
Sure thing. I will try to take a look and write to you a plan on how I'm planning to handle this

saikishor:
This is the idea I've for addressing the collision meshes

* If the URDF only contains the `visual` meshes, but not the `collision` meshes in the collision tag, then the visual meshes are used for MuJoCo
* If the visual and collision meshes are defined, the converter now uses meshes/information defined in the `collision` tag of the URDF
     1. Supports primitive types such as Box, Sphere, cylinder etc
     2. Supports the STL, DAE and OBJ files
[...]

PR #209, opened 2026-05-30
Use collision tags for the collision geometries for the URDF->MJCF conversion

saikishor:
## Description

Recently, when creating different mujoco environments on the fly, I came across the same issue that the converter is using MuJoCo and I have some places the convex hulls are defined in the collision and I believe it is better that it resuses it and not to add different logic based on the URDF tag etc. The following cases are covered:

* If the visual and collision meshes are defined, the converter now uses meshes/information defined in the `collision` tag of the URDF
     1. Supports primitive types such as Box, Sphere, cylinder etc
     2. Supports the STL, DAE and OBJ files
     3.  Visual meshes are used for the visual part in MuJoCo and the Collision meshes are used for the collision detection.
     4. Visual meshes are set to group 2
     5. Collision meshes are set to group 3
* If the URDF only contains the `visual` meshes, but not the `collision` meshes in the collision tag, then the visual meshes are used for MuJoCo

Fixes some concerns listed in #120

With changes in main:

kangaroo_visual_collision_mesh_groups_main

With proposed changes:

kangaroo_visual_collision_mesh_groups

Maybe yes

I did use it for certain parts to improve and also reiterate on ideas

### DIFF
--- a/mujoco_ros2_control/docs/tools.rst
+++ b/mujoco_ros2_control/docs/tools.rst
@@ -129,19 +129,25 @@ Rough outline of the automated conversion process
      For instance, if running multiple types of UR robots in one sim, there will be multiple ``shoulder.dae`` files.
 
 - Reads absolute filepaths of all meshes (from both ``<visual>`` and ``<collision>`` tags) and
-  converts either ``.dae`` or ``.stl`` to ``.obj`` using trimesh.
-
-  - Puts all meshes into an ``assets/`` folder under ``mjcf_data/`` relative to current working dir.
-  - Modifies filepaths again in URDF to point to the ``assets/`` folder.
-  - Decomposes large meshes into multiple components to ensure convex hulls.
+  materializes them into an ``assets/`` folder under ``mjcf_data/`` relative to the current
+  working dir, rewriting the URDF mesh paths to point there.
 
 - Handles visual and collision geometry independently:
 
-  - The ``<visual>`` geometry is used purely for rendering (the ``visual`` class).
-  - The ``<collision>`` geometry drives physics (the ``collision`` class). When a link defines no
-    ``<collision>``, a collision is synthesized from its ``<visual>`` so the visual mesh is reused
-    as the collision shape (the previous behaviour). This synthesis happens at the URDF level, so it
-    stays correct per link even when MuJoCo fuses fixed-jointed bodies together.
+  - **Visual** geometry is used purely for rendering (the ``visual`` class). Visual meshes are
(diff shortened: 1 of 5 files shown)


saikishor:
What do you guys think about this?

ndunkelb-nasa:
> What do you guys think about this?

I think this is definitely something we need to think about! However, I think that there is kind of a problem that the underlying reason for collision meshes for typical ROS things (like motion planning) encourages relatively low-fidelity models (even just like cylinders at times) for motion plans because you don't want to spend time collision checking. However, in MuJoCo, we care a lot about accurate contact dynamics, so that is why I made the decision previously to use the visual tags instead of the collision ones.

However, I definitely acknowledge that this should be at a minimum an option (we want this for other reasons right now), I am just not sure if I want to make it the default. For example, the UR collision objects are giant and would make all of our collisions unrealistic in mujoco. I am fine defaulting to collision objects for collision because that is intuitive, but maybe we have another component in processed_inputs that can call out specific links to use the visual object instead? 

Idk, I really hate this in general, and could see even a way that you might want an entirely different object for mujoco collisions different from the typical collision and visual meshes you might want for motion planning.

By the way, I haven't actually read through the implementation, but will do that tomorrow!

ndunkelb-nasa:
Alright, @saikishor , I have tried to make my way through this like 3 or 4 times now, but honestly, this PR is complex enough that I can't keep track of everything in a readthrough. Would it be possible to break this into a couple of different chunks? 

Unfortunately this tool has so many trees that it goes down, that the logic has become very very convoluted, and so many of the functions tie into the others as well. I feel like the right move would be to technically start this again from scratch and try to make it readable, but I don't think anyone wants to or has time to do that at this point...

saikishor:
@ndunkelb-nasa i have made is quite modular. I can try to split the 18 commits into different PRs. I hope they are self contained. 

Will that approach work for you?

ndunkelb-nasa:
> @ndunkelb-nasa i have made is quite modular. I can try to split the 18 commits into different PRs. I hope they are self contained.
> 
> Will that approach work for you?

Yeah, lets try that! I think there is just way too much going on here.

saikishor:
Sure. Thanks. I'm in middle of a deliverable these days i should be free in 7-10 days. I'll do that at that time or when I find sometime.

Thanks

scastro-nasa:
@saikishor this is the old mega-PR before it was split, right? I think it can be closed too

saikishor:
Yes you are right. Thanks

PR #210, opened 2026-06-01
Refactor camera as plugin

msavchen-nasa:
Separating the cameras into their own ROS plugins.


eholum-nasa:
will reopen from the nasa fork

PR #207, opened 2026-05-29
Add Plugin for the 3D Lidar Extension

eholum-nasa:
Finishes off #187 by adding a `mujoco_ros2_control_plugin` that wraps the extension from #205.

Since the underlying mujoco extension does all of the work, the plugin just runs each update, checks if there's new data available, and copies and publishes it if so. 

I'm loathe to add the plugin to the actual examples until we figure out a better solution for mujoco extensions, so for now I'm opting to assume these are "advanced" use cases with a bit of buyer beware. If you want to see it in action we are testing this with our ridgeback in: https://github.com/NASA-JSC-Robotics/phoebe_bridgeback_ws/pull/11.

Can also test the standalone mujoco extension (depending on your setup) by ensuring the plugin lib is copied to the vendor location (this depends on how you build...):

```
# if using mujoco_vendor apt install, ensure the library is installed next to simulate
$ sudo cp install/mujoco_3d_lidar/lib/mujoco_3d_lidar/libmujoco_3d_lidar.so /opt/ros/${ROS_DISTRO}/opt/mujoco_vendor/bin/mujoco_plugin/
$ ros2 run mujoco_vendor simulate mujoco_extensions/mujoco_3d_lidar/example/lidar.xml

# Or with pixi it should just work...
$ mujoco-simulate mujoco_extensions/mujoco_3d_lidar/example/lidar.xml 
```

For testing the actual `mujoco_ros2_control_plugin`, recommend either using Phoebe ridgeback, your own lidar, or trusting that the unit test works.

PR #211, opened 2026-06-03
Refactor RGB-D Cameras to a Plugin

msavchen-nasa:
Refactoring cameras as a plugin to move them out of the core hw interface.

The advantages,

* Enables users to use different kinds of cameras without creating the publishers (like us)
* Removes the extra lock on the sim mutex and consolidates camera data copying to the control loop

This of course comes at the cost of users needing to update their configs to load the camera plugin, and move the configuration out of the ros2_control xacro.

None of the actual rendering code as changed. The primary difference is that the update rate is just checked in the `update()` callback based on the current sim time. The update loop sleeps until it is triggered to process and fire off a camera update. This stlll happens in a background thread, so the only time spent blocking the control loop is in the copying of data.


mujoco_ros2_control_plugins/src/camera_plugin.cpp
```diff
@@ -332,12 +388,22 @@ void MujocoCameras::update_loop()
   mjr_resizeOffscreen(max_width, max_height, &mjr_con_);
   RCLCPP_INFO(node_->get_logger(), "Resized offscreen buffer to %d x %d", max_width, max_height);
 
+  is_initialized_ = true;
```
msavchen-nasa: @eholum-nasa 
I added this flag to prevent a race condition between the main thread and `update_loop` thread. 
I was noticing that behavior more often during testing.
Let me know what you think!

mujoco_ros2_control_demos/launch/02_mjcf_generation.launch.py
```diff
@@ -131,6 +132,7 @@ def launch_setup(context, *args, **kwargs):
             parameters=[
                 {"use_sim_time": True},
                 ParameterFile(parameters_file),
+                ParameterFile(mujoco_plugins_file),
```
msavchen-nasa: The mujoco camera camera functionality is paired with the ability to publish ROS data.
That is why I think we should, by default include the plugin config files during testing.
If you would like we can decouple these two elements, but I feel like that is out of scope of this PR.

eholum-nasa: Yeah I think this is a nice change, I didn't like only having it in one place, anyway.

mujoco_ros2_control_plugins/src/camera_plugin.cpp
eholum-nasa: What if we just moved `publish_images_` down from the end of `init` here? I think that would give us the same outcome?

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
```diff
@@ -68,37 +78,59 @@ struct CameraData
...
+ * Plugin Parameters Structure (and default topics)
+ * ---------------------------
+ *
+ * mujoco_camera_plugin:
+ *   type: "mujoco_ros2_control_plugins/CameraPlugin"
+ *   camera_publish_rate: 6.0
+ *   <camera_name>:
+ *     frame_name: ""
+ *     info_topic: <camera_name>/camera_info
+ *     image_topic: <camera_name>/colo
```
eholum-nasa:
```suggestion
 *     image_topic: <camera_name>/color
```

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
```diff
@@ -68,37 +78,59 @@ struct CameraData
...
+ *     depth_topic: <camera_name>/depth
+ *
+ * Implementation notes
+ * --------------------
+ * If the camera name, in the parameters, does not match any of the mujoco cameras
+ * the data will not be published to ROS topics.
+ * If no cameras parameters are given (but mujoco_camera_plugin and it's type are declared)
+ * cameras topics will be namespace and given an index to avoid name collision.
```
eholum-nasa: Bringing this in line with what's above

```suggestion
 * The camera name in the parameters must match the name of the mujoco camera in the MJCF.
 * If no camera parameters are provided but the plugin is active, the topics will default to the above. 
```

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
```diff
@@ -68,37 +78,59 @@ struct CameraData
...
+ * If any of the parameters are not specified default topic names will be assigned.
+ * A user can provide topic names for multiple cameras as long as cameras are not named the same.
+ * Name collision is resolved by last name in the yaml file.
+ *
+ * Plugin Parameters Structure (and default topics)
+ * ---------------------------
+ *
+ * mujoco_camera_plugin:
+ *   type: "mujoco_ros2_control_plugins/CameraPlugin"
+ *   camera_publish_rate: 6.0
```
eholum-nasa, line 95:
```suggestion
 *   # All cameras will publish data at the same rate   
 *   camera_publish_rate: 6.0
```

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
```diff
@@ -68,37 +78,59 @@ struct CameraData
...
 
 /**
- * @brief Wraps MuJoCo cameras for publishing ROS 2 RGB-D streams.
+ * @brief Plugin that publishes the mujoco camera data to ROS 2 topics.
+ *
+ * Camera topics can be named in the ROS 2 controls plugins YAML config file
+ * and loaded as node parameters.
+ * If any of the parameters are not specified default topic names will be assigned.
+ * A user can provide topic names for multiple cameras as long as cameras are not named the same.
+ * Name collision is resolved by last name in the yaml file.
```
eholum-nasa, line 87: I think you probably cover this sufficiently in the params example and implementation notes.

eholum-nasa (commented):
Some minor notes but otherwise implementation looks good!

One other thing, can we move the camera docs from the [hw interface ](https://github.com/ros-controls/mujoco_ros2_control/blob/main/mujoco_ros2_control/docs/hardware_interface.rst#cameras) to the [plugins docs](https://github.com/ros-controls/mujoco_ros2_control/blob/main/mujoco_ros2_control_plugins/doc/plugins.rst)?

msavchen-nasa:
> One other thing, can we move the camera docs from the [hw interface ](https://github.com/ros-controls/mujoco_ros2_control/blob/main/mujoco_ros2_control/docs/hardware_interface.rst#cameras) to the [plugins docs](https://github.com/ros-controls/mujoco_ros2_control/blob/main/mujoco_ros2_control_plugins/doc/plugins.rst)?

I knew i forgot something, good catch!

mujoco_ros2_control_plugins/src/camera_plugin.cpp
msavchen-nasa: I dont think so...
I that the update method is called before we are able to initialize the `mj_camera_data_`, and `mjv_copyData` is already trying to write to it.
`publish_images_` flag is not being used by any of the functions in either thread up to those points. 
(I did move it, and saw the problem pop up during the testing)

eholum-nasa: It was being set at the end of initialization, which is probably just wrong now since those containers are constructed by the update loop now. I _think_ moving it there solved that problem for me, but please test and let me know.

eholum-nasa (dismissed):
Added some minor doc tweaks and I swear the updated camera loop now works. Please let me know if that is not the case for you!

This is a rather significant breaking change so will let other folks know before merging.

eholum-nasa (dismissed):
@saikishor @ndunkelb-nasa IMO this is good to go, but it's a breaking change that I also contributed to so I feel icky merging it without one of you taking a look to understand the change.

Long story short, if you're using cameras the config is moving from the ros2_control xacro to a plugin...

saikishor:
> @saikishor @ndunkelb-nasa IMO this is good to go, but it's a breaking change that I also contributed to so I feel icky merging it without one of you taking a look to understand the change.
> 
> Long story short, if you're using cameras the config is moving from the ros2_control xacro to a plugin...

Hello @eholum-nasa,

I'm on my way back home from ICRA, I'll try to take a look during this weekend hopefully. We'll most likely have to branch out to not break for LTS distros. However, we can also make an announcement to let users know. 

So, is it a regular break or is there a strong reason like for better performance etc?

eholum-nasa:
Thanks, @saikishor! There's no rush, ICRA sounded busy so please spend some time recovering!

> So, is it a regular break or is there a strong reason like for better performance etc?

Mostly a "regular break". That said, this _should_ slightly improve stability and the performance of the underlying physics simulation by removing a bunch of the random locks we had on the sim mutex that we passed around everywhere. The bigger win for us is that we can implement different kinds of plugins for the camera that we plan on using.

> We'll most likely have to branch out to not break for LTS distros. 

This, along with supporting [mujoco extensions](https://github.com/ros-controls/mujoco_ros2_control/pull/205), and pulling the [lidar sensor out](https://github.com/ros-controls/mujoco_ros2_control/pull/207), are the big interface changes we had on our bucket list for #168. 

Yeah we will bring this all into our fork to test everything together, too, now that it's a little more stable.

We don't have any other big ticket items planned, just wanted to get these in place before ROSCON / some internal projects we are kicking off.

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -322,9 +317,6 @@ MujocoSystemInterface::on_init(const hardware_interface::HardwareComponentInterf
   const double sim_speed_factor =
       std::stod(get_hardware_parameter(get_hardware_info(), "sim_speed_factor").value_or("-1"));
 
-  // Pull the camera publish rate out of the info, if present, otherwise default to 5 hz.
-  const auto camera_publish_rate =
-      std::stod(get_hardware_parameter(get_hardware_info(), "camera_publish_rate").value_or("5.0"));
```
saikishor, line 327: @msavchen-nasa @eholum-nasa what about adding the plugin automatically if the information is present in the URDF?, this way we can ensure backward compatibility right?

When the user adds info via the hardware interface, we can print a warning that this functionality will be removed soon etc etc

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
```diff
@@ -68,37 +78,59 @@ struct CameraData
...
+ *     frame_name: ""
+ *     info_topic: <camera_name>/camera_info
+ *     image_topic: <camera_name>/color
+ *     depth_topic: <camera_name>/depth
+ *
+ * Implementation notes
+ * --------------------
+ * If the camera name, in the parameters, does not match any of the mujoco cameras
+ * the data will not be published to ROS topics.
+ * If no cameras parameters are given (but mujoco_camera_plugin and it's type are declared)
```
saikishor:
```suggestion
 * If no cameras parameters are given (but mujoco_camera_plugin and its type are declared)
 * the default topics and frame are used automatically
```

saikishor (changes requested):
In general, looks great

Just 2 minor nitpicks

mujoco_ros2_control/src/mujoco_system_interface.cpp
eholum-nasa: That's definitely doable. Pretty sure we could do the same thing with lidar!

saikishor: Perfect. Let's do it in this direction for now :) and in the docs we only show the new way.

eholum-nasa: Ok, done and done, added a handful of helper functions that should also support lidar depreciation (assuming we want that, too). I also reverted the changes on demos 02 - 04 so we still see the warning when we launch those and we can explicitly test those cases in CI. Let me know if you'd prefer I just change those, too, though.

E.g.:

```bash
ros2 launch mujoco_ros2_control_demos 02_mjcf_generation.launch.py
...

[ros2_control_node-3] [WARN] [1780722821.987065248] [MujocoSystemInterface]: 
[ros2_control_node-3] Camera publishing is being auto-configured because camera were found in the MuJoCo model butno CameraPlugin was explicitly configured!
[ros2_control_node-3] This automatic behavior is deprecated and will be removed in a future release.
[ros2_control_node-3] To silence this warning, either:
[ros2_control_node-3]   1. Add a CameraPlugin entry to your mujoco_plugins YAML config, or
[ros2_control_node-3]   2. Set the hardware parameter 'auto_register_cameras' to 'false' to disable auto-registration.
[ros2_control_node-3] For more information refer to the CameraPlugin documentation in mujoco_ros2_control_plugins.
[ros2_control_node-3] 
```

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -2125,13 +2108,138 @@ void MujocoSystemInterface::load_mujoco_plugins()
...
     RCLCPP_ERROR(get_logger(), "Failed to create plugin loader: %s", ex.what());
   }
 }
 
+///
+/// Functions added to support migration of cameras and lidars to plugins, and out of the core hardware interface.
+/// TODO: Delete these once camera / lidar configuration through xacro is fully deprecated.
+///
+
+bool MujocoSystemInterface::auto_register_plugin_if_needed(const std::string& plugin_type, const std::string& plugin_ns,
```
eholum-nasa, line 2127: Added this since we're using it for legacy lidar plugin support, too.

https://github.com/NASA-JSC-Robotics/mujoco_ros2_control/pull/7

mujoco_ros2_control_demos/config/mujoco_ros2_control_plugins.yaml
```diff
@@ -4,3 +4,11 @@
       heart_beat_plugin:
         type: "mujoco_ros2_control_plugins/HeartbeatPublisherPlugin"
         update_rate: 1.0
+      mujoco_camera_plugin:
```
eholum-nasa, line 7: Demo 01 (the normal one) is configured to use this, but 2, 3, and 4 are still following the legacy code path. It's a little weird but at least it gives us integration test coverage out of the box for both the new plugin and the legacy plugin loading. Also confirmed it in our workspace, which #justworks on this branch with no changes.

saikishor (approved):
Looks good to me.

If you want you can reduce the legacy test by one and have one more for new plugin method to have a balance ;) haha


Thanks for such a great work guys 🥳🥳

eholum-nasa:
> If you want you can reduce the legacy test by one and have one more for new plugin method to have a balance ;) haha

I think it _might_ be worth breaking out a new demo! I might stand up a very, very, very basic mobile robot for lidar...

PR #214, opened 2026-06-09
Refactor Rangefinder base Lidar as a Plugin

eholum-nasa:
Following on from the work in #211, this refactors the existing rangefinder based lidar implementation to be a plugin, and pulls it out of the core hardware interface. Like the RGB-D cameras, this will leave a fallback path in place with a warning if users are still configuring lidar sensors in the old way. Of note, nothing in the core demo set is changing with this PR, so all existing lidar setups should remain functional until users update their configs to either the plugin or a different lidar implementation.

This will open the door to new, and perhaps improved, lidar sensor implementations (e.g. #205) that do not rely on expensive rangefinders for raycasting.

For lidar, because of the up-and-coming replacement as well as more limited usage (I'm guessing), I did not want to include detailed documentation. I think it would be better to push users towards the replacement, which has significant capabilities and performance improvements over the baseline.


mujoco_ros2_control_demos/config/mujoco_ros2_control_plugins.yaml
```diff
@@ -12,3 +12,16 @@
           info_topic: /camera/color/camera_info
           image_topic: /camera/color/image_raw
           depth_topic: /camera/aligned_depth_to_color/image_raw
+      # Warning: This is included for test purposes only. We do not recommend users
+      # use this plugin.
```
saikishor, line 16: What do you mean?, Are we replacing this by the other plugin you are preparing?

scastro-nasa (approved):
LGTM

saikishor (approved):
LGTM

mujoco_ros2_control_demos/config/mujoco_ros2_control_plugins.yaml
scastro-nasa: having just reviewed https://github.com/ros-controls/mujoco_ros2_control/pull/205, yes the plan is to replace it by that since it's _much_ more performant.

eholum-nasa: > having just reviewed https://github.com/ros-controls/mujoco_ros2_control/pull/205, yes the plan is to replace it by that since it's much more performant.

That's the idea, but just leaving this around in case other people are using the existing lidar sensors (other than us).

PR #215, opened 2026-06-10
fixing plugin export so that downstream plugins don't fail

ndunkelb-nasa:
Trying to write a new mujoco_ros2_control_plugin in a separate package, and including the package that exists here was resulting in this error, which seemed to just be that we weren't properly exporting OpenGL from the original. Adding OpenGL to the list of exports seems to have resolved this error.

```sh
  By not providing "FindOpenGL::EGL.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "OpenGL::EGL", but CMake did not find one.

  Could not find a package configuration file provided by "OpenGL::EGL" with
  any of the following names:

    OpenGL::EGLConfig.cmake
    opengl::egl-config.cmake

  Add the installation prefix of "OpenGL::EGL" to CMAKE_PREFIX_PATH or set
  "OpenGL::EGL_DIR" to a directory containing one of the above files.  If
  "OpenGL::EGL" provides a separate development package or SDK, be sure it
  has been installed.
```


saikishor (dismissed):
LGTM

Thanks for fixing it

mujoco_ros2_control_plugins/CMakeLists.txt
eholum-nasa: We _might_ experiment with something like this? We really should only need to publicly link those libraries that are directly required by the plugin base....

```suggestion
target_link_libraries(mujoco_ros2_control_plugins 
PUBLIC
  ${MUJOCO_LIB}
  pluginlib::pluginlib
  rclcpp::rclcpp
PRIVATE
  realtime_tools::realtime_tools
  ${sensor_msgs_TARGETS}
  ${std_msgs_TARGETS}
  ${visualization_msgs_TARGETS}
  ${mujoco_ros2_control_msgs_TARGETS}
  glfw
  OpenGL::EGL
  OpenGL::OpenGL
)
```

And remove some of the exported dependencies.

This is hypothetical... but if this doesn't work we should probably break plugin base into it's own library so that downstream plugins don't need to import the world.

eholum-nasa (changes requested):
holding this up to try something else quickly...

eholum-nasa (commented):
Sorry for the correspondingly huger change.

Updated so the plugin is a header only library, which makes _way_ more sense than making users link against the world. 

This now exports a `mujoco_ros2_control_plugins` as a standalone library, and rearranges the existing plugins to be their own `mujoco_ros2_control_plugins_impl` compiled binary that gets loaded separately from the base class.

Makes building external plugins WAY less of a headache.

eholum-nasa (dismissed):
@ndunkelb-nasa this ok?

mujoco_ros2_control_tests/src/test_plugin.cpp
```diff
@@ -0,0 +1,24 @@
+#include <pluginlib/class_list_macros.hpp>
```
eholum-nasa, line 1: Through this in to make sure the library was actually setup / linkable correctly.

ndunkelb-nasa:
> @ndunkelb-nasa this ok?

Yeah looks good to me! Did you just move all the `.hpp`s to src just so that they are semantically separated?

ndunkelb-nasa:
Btw, looks like I can't approve bc it was originally my PR, but here is my :+1: :heavy_check_mark:

eholum-nasa:
> Yeah looks good to me! Did you just move all the .hpps to src just so that they are semantically separated?

That, but also that the headers don't get pulled in to the base `mujoco_ros2_control_plugins` interface!

eholum-nasa (approved):

ndunkelb-nasa:
> That, but also that the headers don't get pulled in to the base `mujoco_ros2_control_plugins` interface!

Ah right, because it grabs the whole directory for the export. I guess this has the one downside that if someone wanted to inherit from one of the plugins that is a default, they couldn't get to the headers from here, but that is probably not a big deal at this point. If wanted later, we could rethink.

PR #217, opened 2026-06-19
Support polling, streaming, and disabled cameras in camera plugin

scastro-nasa:
## Description

This PR adds more options to the camera plugin, specifically to allow "polled" cameras that only render and publish when calling a `std_srvs/Trigger` service.

Generally, there is now a new parameter where you can specify that cameras are streaming, polled, or disabled if you don't even want them to render/publish.

I also snuck in a change where you don't need an intermediate buffer of depth data to flip the coordinates; it can all be done in place. This speeds things up about ~33%!

No, the defaults are still that every camera is streaming.

Yes, I used Claude Opus 4.8 to help me iterate, debug, and write tests, but with extensive manual review, testing, and modification/cleanup.

### DIFF
--- a/mujoco_ros2_control_demos/config/mujoco_ros2_control_plugins.yaml
+++ b/mujoco_ros2_control_demos/config/mujoco_ros2_control_plugins.yaml
@@ -8,7 +8,7 @@
         type: "mujoco_ros2_control_plugins/CameraPlugin"
         camera_publish_rate: 6.0
         camera:
-          camera_type: streaming
+          policy: streaming
           frame_name: camera_color_optical_frame
           info_topic: /camera/color/camera_info
           image_topic: /camera/color/image_raw
--- a/mujoco_ros2_control_demos/config/mujoco_ros2_control_plugins.yaml
+++ b/mujoco_ros2_control_demos/config/mujoco_ros2_control_plugins.yaml
@@ -8,6 +8,7 @@
         type: "mujoco_ros2_control_plugins/CameraPlugin"
         camera_publish_rate: 6.0
         camera:
+          camera_type: streaming
           frame_name: camera_color_optical_frame
           info_topic: /camera/color/camera_info
           image_topic: /camera/color/image_raw
--- a/mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
+++ b/mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
@@ -51,7 +51,7 @@
 namespace mujoco_ros2_control_plugins
 {
 
-enum class CameraType
+enum class CameraPolicy
 {
   DISABLED = 0,  // Publishing is disabled.
   STREAMING,     // Publishing is constant at the specified publish rate.
@@ -60,7 +60,7 @@ enum class CameraType
 
 struct CameraData
 {
-  CameraType type = CameraType::STREAMING;
+  CameraPolicy policy = CameraPolicy::STREAMING;
 
   // Set by the trigger service for polled cameras to request a one-shot capture.
   bool poll_requested{ false };
@@ -117,14 +117,14 @@ struct CameraData
  *     info_topic: <camera_name>/camera_info
  *     image_topic: <camera_name>/color
  *     depth_topic: <camera_name>/depth
- *     trigger_service_name: <camera_name>/trigger  # only used for polling camera type
+ *     trigger_service_name: <camera_name>/trigger  # only used for polling camera policy
  *
  * Implementation notes
  * --------------------
  * If the camera name, in the parameters, does not match any of the mujoco cameras
  * the data will not be published to ROS topics.
- * If no cameras parameters are given (but mujoco_camera_plugin and its type are declared)
- * the default type (streaming), topics, and frame are used automatically.
+ * If no camera parameters are given, but the mujoco_camera_plugin and its policy are
+ * declared, the default policy (streaming), topics, and frame are used automatically.
  *
  */
 class CameraPlugin : public MuJoCoROS2ControlPluginBase
@@ -173,9 +173,10 @@ private:
   void close();
 
   /**
-   * @brief Parses camera information from the mujoco model.
+   * @brief Parses camera information from the mujoco model and node parameters.
+   * @return true if registering cameras succeeds, otherwise false.
    */
-  void register_cameras();
+  bool register_cameras();
 
   /**
    * @brief Initializes the rendering context and starts processing.
--- a/mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
+++ b/mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
@@ -112,7 +112,7 @@ struct CameraData
  *   # All cameras will publish data at the same rate
  *   camera_publish_rate: 6.0
  *   <camera_name>:
- *     camera_type: <camera_type>  # "disabled", "polled", or "streaming"
+ *     policy: <policy>  # "disabled", "polled", or "streaming"
  *     frame_name: ""
  *     info_topic: <camera_name>/camera_info
  *     image_topic: <camera_name>/color
--- a/mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
+++ b/mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
@@ -61,7 +61,13 @@ enum class CameraType
 struct CameraData
 {
   CameraType type = CameraType::STREAMING;
-  bool triggered{ false };
+
+  // Set by the trigger service for polled cameras to request a one-shot capture.
+  bool poll_requested{ false };
+
+  // Set when this camera is selected to render on the next pass and is waiting
+  // for the rendering thread to consume it, then clear it when rendering occurs.
+  bool render_pending{ false };
 
   mjvCamera mjv_cam;
   mjrRect viewport;
@@ -137,12 +143,26 @@ public:
   void cleanup() override;
 
   /**
-   * @brief Manually force the cameras to publish their updates.
+   * @brief Synchronously run a single render/publish cycle in the calling thread.
    *
-   * This is included for test purposes only, as the plumbing to manually force a data
-   * sync in WIP.
+   * This mirrors what the rendering thread does on each pass: streaming cameras are
+   * always selected, and polled cameras are selected only if their trigger service has
+   * been called since the last cycle. It is intended for tests, where it lets the camera
+   * selection logic be exercised deterministically without relying on the asynchronous
+   * rendering thread or a live GL context.
    */
   void trigger_update();
+
+  /**
+   * @brief Whether the rendering thread has successfully initialized its GL context.
+   *
+   * Primarily useful for tests that need to wait for the asynchronous rendering thread to
+   * come up before forcing a render.
+   */
+  bool is_rendering_available() const
+  {
+    return publish_images_.load();
+  }
 
 private:
   // ROS interfaces
@@ -164,7 +184,7 @@ private:
   void update_loop();
 
   /**
-   * @brief Updates the camera images and publishes info, images, and depth maps.
+   * @brief Renders and publishes every camera that is flagged to render.
    */
   void update_cameras();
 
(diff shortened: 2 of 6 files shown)


mujoco_ros2_control_plugins/src/camera_plugin.cpp
```diff
@@ -125,6 +176,32 @@ void CameraPlugin::register_cameras()
...
+    {
+      camera.type = CameraType::DISABLED;
+    }
+    else
+    {
+      RCLCPP_ERROR(node_->get_logger(), "Invalid camera type for camera '%s': '%s'. Disabling.", cam_name,
+                   type_str.c_str());
+      camera.type = CameraType::DISABLED;
+      type_str = "disabled";
+    }
```
scastro-nasa, line 205: we could alternatively just throw / shutdown here as we prefer

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
```diff
@@ -44,14 +44,31 @@
...
+#include <std_srvs/srv/trigger.hpp>
 
 #include <pluginlib/class_list_macros.hpp>
 
 namespace mujoco_ros2_control_plugins
 {
 
+enum class CameraType
+{
+  DISABLED = 0,  // Publishing is disabled.
```
saikishor, line 56: Should we have disabled as an option? 
Why is this mode interesting to have?

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
```diff
@@ -94,17 +113,19 @@ struct CameraData
  *   # All cameras will publish data at the same rate
  *   camera_publish_rate: 6.0
  *   <camera_name>:
+ *     camera_type: <camera_type>  # "disabled", "polled", or "streaming"
  *     frame_name: ""
  *     info_topic: <camera_name>/camera_info
  *     image_topic: <camera_name>/color
  *     depth_topic: <camera_name>/depth
+ *     trigger_service_name: <camera_name>/trigger
```
saikishor:
```suggestion
 *     trigger_service_name: <camera_name>/trigger (only used for POLLING camera type)
```

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
```diff
@@ -94,17 +113,19 @@ struct CameraData
  *   # All cameras will publish data at the same rate
  *   camera_publish_rate: 6.0
  *   <camera_name>:
+ *     camera_type: <camera_type>  # "disabled", "polled", or "streaming"
```
saikishor: Just curious, should we call it `camera_type` or just call it `policy` or `steaming_type`?

saikishor (commented):
Leaving a partial review for now

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
scastro-nasa: I considered if you have a MuJoCo setup with multiple cameras but you don't want all of them to publish to ROS? It does take computational resources to render and publish a camera that you may not want to be consumed.

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
scastro-nasa: I'm fine with `camera_type` or `policy` or `mode`, as you prefer.

`streaming_type` I don't like because streaming is one of the various modes of operation.

saikishor: Let's go with `policy` or `mode`? Just because the `camera_mode` might give an impression of `fisheye`, `stereo` etc.

mujoco_ros2_control_plugins/src/camera_plugin.cpp
```diff
@@ -67,21 +67,59 @@ void CameraPlugin::update(const mjModel* model_arg, mjData* data)
...
+  // Streaming cameras render on a fixed-rate clock; polled cameras render as soon as a
+  // trigger has been received. Both are serviced here, on the sim thread, because this is
+  // the only place we can safely snapshot the live mjData without racing the simulation.
   // TODO: Support per-camera publish rates?
-  auto now = node_->get_clock()->now();
-  if ((now - last_publish_time_).seconds() < (1.0 / camera_publish_rate_))
+  const auto now = node_->get_clock()->now();
+  const bool stream_due =
+      has_streaming_cameras_ && (now - last_publish_time_).seconds() >= (1.0 / camera_publish_rate_);
+  const bool poll_due = poll_pending_.load();
```
eholum-nasa: For consuming a signal like this you can grab and set the boolean in a single instruction using [exchange](https://en.cppreference.com/cpp/atomic/atomic/exchange). Which will both set the desired value and return the current value:

```suggestion
  const bool poll_due = poll_pending_.exchange(false);
```

That both removes the need for the `poll_pending_ = false;` below, and removes the (extremely unlikely) scenario that a trigger is called between grabbing this and setting it to false below.

mujoco_ros2_control_plugins/src/camera_plugin.cpp
eholum-nasa: ah yeah, given the complexity that is growing here maybe the more errors with hints the better? rather than starting things up and wondering why it's not working. Could change `register_cameras` to return a boolean for success or failure, then fast fail if registration borks during initialization.

mujoco_ros2_control_plugins/src/camera_plugin.cpp
```diff
@@ -427,65 +537,88 @@ void CameraPlugin::update_loop()
...
-        auto idx = h * camera.width + w;
-        auto idx_flipped = (camera.height - 1 - h) * camera.width + w;
-        camera.depth_buffer[idx] = near / (1.0f - camera.depth_buffer[idx] * (depth_scale));
-        camera.depth_buffer_flipped[idx_flipped] = camera.depth_buffer[idx];
-      }
-    }
-    // Copy flipped data into the depth image message, floats -> unsigned chars
-    std::memcpy(&camera.depth_image.data[0], camera.depth_buffer_flipped.data(), camera.depth_image.data.size());
+    render_and_publish_camera(cameras_[idx], node_->now());
+  }
```
eholum-nasa, line 575: ah, this was wrong before but re-reading it fresh I just noticed. We probably want to pull the simulation time when the data was copied for rendering, rather than sim time, right? Again probably not a significant difference here, but maybe it could be something like this:


```suggestion
  // Use the snapshotted data's timestamp in the ROS header, as that is when the data was actually pulled
  const auto render_time = mj_camera_data_->time;
  int32_t sim_time_sec = static_cast<int32_t>(std::floor(render_time));
  uint32_t sim_time_nanosec = static_cast<uint32_t>((render_time - sim_time_sec) * 1e9);
  rclcpp::Time stamp(sim_time_sec, sim_time_nanosec, RCL_ROS_TIME);
  
  for (const auto idx : render_indices_)
  {
    render_and_publish_camera(cameras_[idx], stamp);
  }
```

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
eholum-nasa: Yeah I think there's value there even if it's unpopular? Useful for testing pipelines or being able to turn things off if necessary. Though I guess there's not _really_ a huge functional difference between a disabled camera and polled camera that never gets called.

I don't have strong feelings and it's not a lot of code... so I would lean towards just leaving it.

eholum-nasa (commented):
Tested and delivers! Noted a few small things in addition to Sai's comments. Thank you for this!

mujoco_ros2_control_plugins/src/camera_plugin.cpp
scastro-nasa: whoa, TIL

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/camera_plugin.hpp
saikishor: Fine by me!

scastro-nasa:
Thanks @saikishor and @eholum-nasa -- implemented your comments. Went with `policy` as the param name.

mujoco_ros2_control_plugins/src/camera_plugin.cpp
scastro-nasa: done but with slight simplification

mujoco_ros2_control_plugins/src/camera_plugin.cpp
```diff
@@ -150,17 +229,38 @@ void CameraPlugin::register_cameras()
...
+    }
+    camera.trigger_service_name = node_->get_parameter(param_ns + "trigger_service_name").as_string();
+
     RCLCPP_INFO(node_->get_logger(), "Adding camera: '%s'", cam_name);
+    RCLCPP_INFO(node_->get_logger(), "    camera_type: '%s'", policy_str.c_str());
     RCLCPP_INFO(node_->get_logger(), "    frame_name: '%s'", camera.frame_name.c_str());
-    RCLCPP_INFO(node_->get_logger(), "    info_topic: '%s'", camera.info_topic.c_str());
-    RCLCPP_INFO(node_->get_logger(), "    image_topic: '%s'", camera.image_topic.c_str());
-    RCLCPP_INFO(node_->get_logger(), "    depth_topic: '%s'", camera.depth_topic.c_str());
+    if (camera.policy != CameraPolicy::DISABLED)
```
eholum-nasa: Should we just `continue` if the camera is disabled? I think we're creating all the things for it otherwise.

mujoco_ros2_control_plugins/src/camera_plugin.cpp
```diff
@@ -150,17 +229,38 @@ void CameraPlugin::register_cameras()
...
     camera.depth_topic = node_->get_parameter(param_ns + "depth_topic").as_string();
 
+    if (!node_->has_parameter(param_ns + "trigger_service_name"))
+    {
+      node_->declare_parameter(param_ns + "trigger_service_name", camera.name + "/trigger");
+    }
+    camera.trigger_service_name = node_->get_parameter(param_ns + "trigger_service_name").as_string();
+
     RCLCPP_INFO(node_->get_logger(), "Adding camera: '%s'", cam_name);
+    RCLCPP_INFO(node_->get_logger(), "    camera_type: '%s'", policy_str.c_str());
```
eholum-nasa:
```suggestion
    RCLCPP_INFO(node_->get_logger(), "    policy: '%s'", policy_str.c_str());
```

mujoco_ros2_control_plugins/src/camera_plugin.cpp
```diff
@@ -427,65 +541,92 @@ void CameraPlugin::update_loop()
...
+
+  camera_near_distance_ = static_cast<float>(mj_model_->vis.map.znear * mj_model_->stat.extent);
   const float far = static_cast<float>(mj_model_->vis.map.zfar * mj_model_->stat.extent);
-  const float depth_scale = 1.0f - near / far;
-  for (auto& camera : cameras_)
+  camera_depth_scale_ = 1.0f - camera_near_distance_ / far;
+
+  // Use the snapshotted data's timestamp in the ROS header, as that is when the data was actually pulled.
+  const rclcpp::Duration duration = rclcpp::Duration::from_seconds(mj_camera_data_->time);
+  rclcpp::Time stamp(duration.nanoseconds(), RCL_ROS_TIME);
```
eholum-nasa, line 570: Oh yeah good call. This doesn't truncate, right?

mujoco_ros2_control_plugins/src/camera_plugin.cpp
scastro-nasa: argh yeah I missed some of the tests too

mujoco_ros2_control_plugins/src/camera_plugin.cpp
scastro-nasa: I don't think so, it's not just the remainder, it's the full time.

mujoco_ros2_control_plugins/src/camera_plugin.cpp
eholum-nasa: oh yeah just noticed that, too. Sorry!

mujoco_ros2_control_demos/config/mujoco_ros2_control_plugins.yaml
```diff
@@ -8,6 +8,7 @@
         type: "mujoco_ros2_control_plugins/CameraPlugin"
         camera_publish_rate: 6.0
         camera:
+          camera_type: streaming
```
saikishor, line 11:
```suggestion
          policy: streaming
```

Should this be renamed now?

eholum-nasa (approved):
:rocket:

saikishor (commented):
LGTM

mujoco_ros2_control_demos/config/mujoco_ros2_control_plugins.yaml
scastro-nasa: I think I did rename it in the last commit! (luckily since the PR was merged)

Issue #216, opened 2026-06-19
test_camera_plugin fails on CI


christophfroehlich:
**Describe the bug**
```
2026-06-18T06:50:08.3216047Z build/mujoco_ros2_control_plugins/test_results/mujoco_ros2_control_plugins/test_camera_plugin.gtest.xml: 1 test, 1 error, 0 failures, 0 skipped
2026-06-18T06:50:08.3217191Z - mujoco_ros2_control_plugins test_camera_plugin.gtest.missing_result
2026-06-18T06:50:08.3217727Z   <<< error message
2026-06-18T06:50:08.3217958Z     The test did not generate a result file:
2026-06-18T06:50:08.3218226Z     
2026-06-18T06:50:08.3218425Z     [==========] Running 2 tests from 1 test suite.
2026-06-18T06:50:08.3218729Z     [----------] Global test environment set-up.
2026-06-18T06:50:08.3219030Z     [----------] 2 tests from CameraPluginTest
2026-06-18T06:50:08.3219379Z     [ RUN      ] CameraPluginTest.InitSucceedsWithNoCameras
2026-06-18T06:50:08.3219875Z     [INFO] [1781764077.576265706] [test_camera_plugin_node]: CameraPlugin initializing cameras...
2026-06-18T06:50:08.3220514Z     [INFO] [1781764077.576390089] [test_camera_plugin_node]: Publishing camera data at rate 5.000000 per second.
2026-06-18T06:50:08.3221082Z     [       OK ] CameraPluginTest.InitSucceedsWithNoCameras (11 ms)
2026-06-18T06:50:08.3221454Z     [ RUN      ] CameraPluginTest.InitAndPublish
2026-06-18T06:50:08.3221898Z     [INFO] [1781764077.583199423] [test_camera_plugin_node]: CameraPlugin initializing cameras...
2026-06-18T06:50:08.3222611Z     [INFO] [1781764077.583269083] [test_camera_plugin_node]: Publishing camera data at rate 5.000000 per second.
2026-06-18T06:50:08.3223194Z     [INFO] [1781764077.583452126] [test_camera_plugin_node]: Adding camera: 'test_cam'
2026-06-18T06:50:08.3223680Z     [INFO] [1781764077.583474287] [test_camera_plugin_node]:     frame_name: ''
2026-06-18T06:50:08.3224199Z     [INFO] [1781764077.583485077] [test_camera_plugin_node]:     info_topic: 'test_cam/camera_info'
[...]

PR #223, opened 2026-07-06
Update Pixi packages with pixi.lock

saikishor:
This PR updates the pixi package information


eholum-nasa:
MujoCo! I'll fix this...

eholum-nasa:
I think I'm there... saw a failure from https://github.com/ros-controls/mujoco_ros2_control/issues/216 which is a separate and more difficult issue.

eholum-nasa (approved):
Approving some of my own work here but if it looks good to you @saikishor it looks good to me!

saikishor (commented):
LGTM

Thanks for fixing it @eholum-nasa

PR #224, opened 2026-07-10
Fix controller manager rate collapse by decoupling control and physics locks

scastro-nasa:
## Description

This PR seeks to address some of the issues introduced in #191 related to observed performance regression.

Basically, we're not going to get quite the same performance as before this PR because prior to it we were actually living in a data race world where multiple things could touch a shared `mjData` object at the same time. So even after this PR, there will still be spikes unless we start running this with RT priority etc., but the improvements are still quite significant.

The major reason for performance regression was that the same mutex was being used for both stepping the simulation (which could choke things up) and actually calling `read()` on the controller manager. Fixing that honestly gave us most of the gains.

The key changes in this PR are:

1. Making sure that the physics loop pushes its copy of data to a snapshot the controller manager consumes so that `read()` doesn't have to content with the same mutex. We're using a double-buffered snapshot so that the `read()` / `write()` operations are as fast as possible.
2. Using a reduced `ControlState` struct for the `read()` and `write()` operations under its own mutex, which prevents from having to copy an entire `mjData` structure and lock when accessing it.
3. For plugins, we still need a full `mjData` snapshot. Notably, if there are no plugins we can exclusively  use `control_state_` instead.

We can simplify the design if we want, since the biggest win is relatively simple compared to the full PR. But I was trying a bunch of things along the way. Looking forward to feedback!

No, this is all under the hood. Granted now the read/write could trail behind the sim by some time steps.

### Data

Took some ~85 second snapshots in plotjuggler. Please note the differences on the Y scale.

Before this PR (main):
image

After this PR:
image


mujoco_ros2_control/tests/CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
 find_package(ament_cmake_gtest REQUIRED)
 find_package(ament_cmake_pytest REQUIRED)
 
-ament_add_gtest(test_mujoco_simluation test_mujoco_simluation.cpp)
-target_link_libraries(test_mujoco_simluation
+ament_add_gtest(test_mujoco_simulation test_mujoco_simulation.cpp)
+target_link_libraries(test_mujoco_simulation
```
scastro-nasa, line 5: sorry, I couldn't help fixing the typo

mujoco_ros2_control/src/mujoco_simulation.cpp
```diff
@@ -1048,6 +1154,16 @@ void MujocoSimulation::physics_loop()
             steps_cv_.notify_all();
           }
 
+          // Refresh the full snapshot on demand, at the START of the iteration rather than
+          // right after stepping. The data is identical (nothing has stepped since the last
+          // batch), but the timing matters: batch end is exactly when the last /clock tick
+          // woke the controller cycle, so an mjData copy there competes with the controller
+          // for memory bandwidth.
```
```suggestion
```

mujoco_ros2_control/tests/test_mujoco_simulation.cpp
```diff
@@ -200,10 +200,17 @@ TEST_F(MujocoSimulationTest, ControlUpdateTests)
   control->ctrl[0] = 0.75;
   control->qfrc_applied[0] = 1.5;
 
-  // Apply to sim and verify
+  // Apply to sim. Inputs are staged and copied into the sim data by the physics loop
+  // immediately before each step, so they must not appear until stepping occurs.
```
saikishor, line 204:
```suggestion
```

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -929,6 +931,18 @@ hardware_interface::return_type MujocoSystemInterface::write(const rclcpp::Time&
 #endif
   };
 
+  // Borrow the latest completed physics snapshot as this cycle's control-data view for the
+  // plugins. Producer-pays copying: the physics loop fills the snapshot buffers on its own
+  // thread, and acquiring one is an O(1) pointer swap — write() never executes a scene-sized
+  // copy, so contact bursts and page faults in mj_copyData can no longer appear as controller
+  // cycle spikes. Without plugins nothing here reads sim state, so the persistent local
+  // scratch container is used instead and no snapshot traffic is generated at all.
```
saikishor:
```suggestion
```

saikishor (changes requested):
The changes look more or less fine, but can you reduce the number of comments in different parts of the code

mujoco_ros2_control/tests/test_mujoco_simulation.cpp
scastro-nasa: I actually think this comment is important because it describes the EXPECTs of zero immediately after.

eholum-nasa (commented):
I really like the double buffer approach here, but I wonder if there's a simpler way to get the same boost using some kind of signal to the physics loop inside the swap function. That would theoretically keep the same API exposed to the ros2 control hw interface but provide the same performance gain from the mutex hold during the longer copies... will play with that if that's ok...

scastro-nasa:
go ahead! But note that `std::atomic<bool> snapshot_refresh_requested_` is already a signal to the physics loop, no?

The key is that the swap operation is held by a `data_exchange_mutex_` whereas the physics is on a separate `sim_mutex_`, so those 2 shouldn't contend with each other? They're just preventing re-entering themselves?

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
```diff
@@ -189,17 +197,65 @@ class MujocoSimulation
...
+
+  /**
+   * @brief Small snapshot of the state the hardware interface needs every control cycle.
+   *
+   * This data structure is much smaller than a full mjData copy, so publishing and copying
+   * it is also far cheaper, regardless of scene complexity. Refreshed by the physics loop
+   * after every step, immediately before that step's /clock tick, so a consumer woken by a
+   * clock tick sees state that is synchronous with (or newer than) that tick's sim time.
+   */
+  struct ControlState
```
eholum-nasa, line 232: Ok, I finally understand this well enough and the question I have is if the savings speed-up in the no-plugins case is worth the complexity of the extra structure, or if we should just do the full `mj_copyData` in the outer physics loop when requested and keep `mjData` as the be-all end-all for swapping data between the simulation and the consumers of the sim.

I put timing loops around the data copy call, as well as the whole outer physics loop just to check timing with and without plugins using CLR,

With the camera plugin enabled:

Image

With all plugins disabled so the copy never happens:

Image

And with Phoebe and it's expensive lidar computation,

With plugins:

Image

Without plugins:

Image

So on my machine we nominally get a 100-200 μs speedup in the physics_loop every control loop hz by having the extra swap for the ControlState, as long as we don't use plugins.

The question is, is the extra complexity worth the gains in the non-plugins case? My inclination is no... but that might just be because all of our robots will always use camera or lidar or something. Maybe the general public would appreciate the zoomy-ness of the physics loop... thoughts?

scastro-nasa: I think there will be a healthy chunk of people that runs the sim for control purposes only, things like blind (proprioception-only) RL locomotion policies, or simply designing actuator controllers in sim.

So I would personally eat the ControlState complexity if the savings are that tangible.

eholum-nasa: Yeah I think you're right. Depending on the model and how complex your stepping is that copy could end up being like 50% of your whole loop rate... It'll never save time for us but I won't be selfish.

Let's run with this... I'll take another read through just to check.

saikishor: What I don't like are the huge peaks. Relatively high, which is very surprising

eholum-nasa: For the mobile robot those large spikes are the lidar sensor spiking the compute time of `mj_step`. If we remove it we still some some spiking, but we don't really have too much control over the convergence rate of the underlying sim since that's mujoco. We also see slow downs in the sim performance when there are lots of contacts, for instance. That said a big part of the benefit of this change is that we're never holding up the physics computation to read data from the physics sim...

Just for completeness I see the same improvements in the control loop stats as @scastro-nasa :

Screenshot 2026-07-14 at 5 40 19 PM

saikishor: Sure. Atleast, let's go ahead with this. It is already quite some improvement

saikishor (dismissed):

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
scastro-nasa: Thanks for the approve! My tyrant boss (aka @eholum-nasa) is reading through this and may have other comments / suggestions, so I will wait for him too

eholum-nasa (dismissed):
I thought I had but I got that sweet sweet 500 error

scastro-nasa:
Were some tests flaky before this PR? I saw a segfault in a previous run and want to know if it's something this could introduce before I merge it.

See https://github.com/ros-controls/mujoco_ros2_control/actions/runs/29441646276/job/87441860698?pr=224

scastro-nasa:
> Were some tests flaky before this PR? I saw a segfault in a previous run and want to know if it's something this could introduce before I merge it.

Was able to reproduce and fixed in https://github.com/ros-controls/mujoco_ros2_control/commit/edac966dc50bffcf421f90a7f183b3451b951e32

Running the CI jobs a few times in here to verify https://github.com/ros-controls/mujoco_ros2_control/actions/runs/29445683396

saikishor:
> > Were some tests flaky before this PR? I saw a segfault in a previous run and want to know if it's something this could introduce before I merge it.
> 
> Was able to reproduce and fixed in https://github.com/ros-controls/mujoco_ros2_control/commit/edac966dc50bffcf421f90a7f183b3451b951e32
> 
> Running the CI jobs a few times in here to verify https://github.com/ros-controls/mujoco_ros2_control/actions/runs/29445683396

Awesome. Thanks a lot

saikishor (approved):

PR #226, opened 2026-07-14
fix dependency declarations for bloom

christianrauch:
## Description

When blooming `mujoco_ros2_control`, I found that some dependencies are not properly defined. This is especially visible when building the packages in isolation.

This PR fixes the dependency declarations and adds a new workflow that uses [`jspricke/ros-deb-builder-action`](https://github.com/jspricke/ros-deb-builder-action) to bloom packages in isolation. The action can also publish the bloomed Debian packages on a branch, but this is deactivated here. The action is only used to check that the Debian packages can be built.

Fixes # (issue)

No.

### Additional Information

No.

## TODOs

- [x] Fork the repository.

### DIFF
--- a/mujoco_ros2_control_plugins/package.xml
+++ b/mujoco_ros2_control_plugins/package.xml
@@ -29,7 +29,7 @@
   <depend>std_srvs</depend>
   <depend>geometry_msgs</depend>
   <depend>visualization_msgs</depend>
-  <depend version_gte="0.0.3">mujoco_ros2_control_msgs</depend>
+  <depend>mujoco_ros2_control_msgs</depend>
   <depend>libglfw3-dev</depend>
   <depend>opengl</depend>
 
--- a/mujoco_ros2_control_plugins/package.xml
+++ b/mujoco_ros2_control_plugins/package.xml
@@ -29,7 +29,7 @@
   <depend>std_srvs</depend>
   <depend>geometry_msgs</depend>
   <depend>visualization_msgs</depend>
-  <depend>mujoco_ros2_control_msgs</depend>
+  <depend version_gte="0.0.3">mujoco_ros2_control_msgs</depend>
   <depend>libglfw3-dev</depend>
   <depend>opengl</depend>
 
--- a/mujoco_ros2_control_plugins/package.xml
+++ b/mujoco_ros2_control_plugins/package.xml
@@ -31,6 +31,7 @@
   <depend>visualization_msgs</depend>
   <depend>mujoco_ros2_control_msgs</depend>
   <depend>libglfw3-dev</depend>
+  <depend>opengl</depend>
 
   <test_depend>mujoco_3d_lidar</test_depend>
   <test_depend>ament_cmake_gtest</test_depend>
--- a/mujoco_ros2_control_plugins/package.xml
+++ b/mujoco_ros2_control_plugins/package.xml
@@ -30,6 +30,7 @@
   <depend>geometry_msgs</depend>
   <depend>visualization_msgs</depend>
   <depend>mujoco_ros2_control_msgs</depend>
+  <depend>libglfw3-dev</depend>
 
   <test_depend>mujoco_3d_lidar</test_depend>
   <test_depend>ament_cmake_gtest</test_depend>
--- a/.github/workflows/ros-deb-builder.yaml
+++ b/.github/workflows/ros-deb-builder.yaml
@@ -1,0 +1,29 @@
+name: ROS Debian Package Builder
+
+on:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - "README.md"
+  pull_request:
+    branches:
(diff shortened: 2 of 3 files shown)


saikishor (commented):
I don't know why you had issues, I was able to bloom the previous one properly

christianrauch:
> I don't know why you had issues, I was able to bloom the previous one properly

Did you bloom the packages in isolation? If dependencies are installed before the package is bloomed, or dependencies are installed as a transient dependency of another package, you won't see these issues unless you build in isolation.

If you compare the packages used in the `CMakeLists.txt` and the ones defined in the `package.xml`, you will see some discrepancies. E.g. `mujoco_ros2_control_plugins` uses `find_package(glfw3 REQUIRED)`, but there is no dependency declaration for `libglfw3-dev` in the `package.xml`.

saikishor:
I haven't ran it in the isolation yet. I usually run it on our internal CI which runs in isolation and then I bloom it

christianrauch:
> I haven't ran it in the isolation yet. I usually run it on our internal CI which runs in isolation and then I bloom it

You won't see these issues if you build the repo in isolation. For example, GLFW3 is declared as a dependency of `mujoco_ros2_control`, but not for `mujoco_ros2_control_plugins`. But both packages use `find_package(glfw3 REQUIRED)`. If you run rosdep on the repo, glfw3 will be installed because of `mujoco_ros2_control` and thus `mujoco_ros2_control_plugins` will build too.

Building properly in isolation, e.g. via the GitHub action in this PR, discovers missing build dependencies.

saikishor:
@christianrauch I've mentioned it couple of times till now, not to do force pushes on to the branches

mujoco_ros2_control_plugins/package.xml
```diff
@@ -29,7 +29,9 @@
   <depend>std_srvs</depend>
   <depend>geometry_msgs</depend>
   <depend>visualization_msgs</depend>
-  <depend>mujoco_ros2_control_msgs</depend>
+  <depend version_gte="0.0.3">mujoco_ros2_control_msgs</depend>
```
saikishor:
```suggestion
  <depend>mujoco_ros2_control_msgs</depend>
```

christianrauch:
> @christianrauch I've mentioned it couple of times till now, not to do force pushes on to the branches

You said this is for comments not to get lost. But there are no review comments on this PR yet.

mujoco_ros2_control_plugins/package.xml
christianrauch: This is actually required because the API changes between the message definition versions.

saikishor:
> > @christianrauch I've mentioned it couple of times till now, not to do force pushes on to the branches
> 
> You said this is for comments not to get lost. But there are no review comments on this PR yet.

But I did push some changes earlier and you simply force-pushed your local changes

christianrauch:
@saikishor Why are you removing the version check? The build will fail with an older version of the message definitions due to the changes in the message definition. You need the newer version to build the package.

saikishor:
> @saikishor Why are you removing the version check? The build will fail with an older version of the message definitions due to the changes in the message definition. You need the newer version to build the package.

Because we package them together and if the mujoco_ros2_control_msgs exist on a different repo, it would be definitely needed. As we package and release them together, I removed it

christianrauch:
> > @saikishor Why are you removing the version check? The build will fail with an older version of the message definitions due to the changes in the message definition. You need the newer version to build the package.
> 
> Because we package them together and if the mujoco_ros2_control_msgs exist on a different repo, it would be definitely needed. As we package and release them together, I removed it

It's the same repo, but if you build in isolation, the packages in the repo will be built independently, and this will cause an issue, e.g. if version 0.0.2 is in the repo, but you actually need version 0.0.3. This is akin to all the other dependency fixes in this PR. 

As far as I understand the isolated build, if you only specify to rely on any package version, it will use the binary bloomed version with the old API. If you specify the version requirement, it will build and use the newer version.

christianrauch:
... but if the bloom CI is happy, this may not be the case here.

But generally, I would say that dependencies should be specified as tightly as possible to avoid potential issues such as API incompatibility.

saikishor:
> As far as I understand the isolated build, if you only specify to rely on any package version, it will use the binary bloomed version with the old API. If you specify the version requirement, it will build and use the newer version.

The isolation builds worked with the current version, so I think it is good enough.

christianrauch:
> > > @christianrauch I've mentioned it couple of times till now, not to do force pushes on to the branches
> > 
> > 
> > You said this is for comments not to get lost. But there are no review comments on this PR yet.
> 
> But I did push some changes earlier and you simply force-pushed your local changes

I rebased, squashed one of your commits to make reading the history easier, and I removed your commit that removes the version requirement, as I think requiring the correct version is the right thing to do.

PR #231, opened 2026-07-16
Fix state passed into pose sensors

scastro-nasa:
## Description

This PR addresses a bug that got through with https://github.com/ros-controls/mujoco_ros2_control/pull/224, in that the pose sensors were getting data from the wrong state data structure.

Also included a regression test since this was not caught in the former PR.

Fixes https://github.com/ros-controls/mujoco_ros2_control/issues/230

### Is this user-facing behavior change?

No

### Did you use Generative AI?

Yes, Claude Fable 5 to help generate a test.

If there are more idiomatic uses of ros2_control syntax for the test, I'm open to suggestions.

PR #232, opened 2026-07-16
Add pose sensor test to launch tests

christianrauch:
## Description

Add a test that the pose published by the `pose_broadcaster` is as expected.

Fixes # (issue)

No.

Yes.

## TODOs

- [x] Fork the repository.


scastro-nasa:
Thank you! I just merged https://github.com/ros-controls/mujoco_ros2_control/pull/231 so it should pass after a merge/rebase, though please note that I changed the site location so you might have to modify the expected value to match.

christianrauch:
> Thank you! I just merged #231 so it should pass after a merge/rebase, though please note that I changed the site location so you might have to modify the expected value to match.

Can you adjust the test with what the new expected pose according to your new sensor placement is?

scastro-nasa:
Updated! I have no idea what CI is doing, but those failures look quite unnatural :sweat_smile: 

(I think something is wrong with GitHub, so will retry the jobs later)

scastro-nasa:
The pose sensor position is quite flaky for this dynamic frame (or maybe the robot doesn't make it within the 5 second limit?), so I had to increase the padding value for the checks by quite a lot and it still doesn't seem to pass reliably.

scastro-nasa:
ok I think the issue was that the launch tests weren't isolating domain IDs.

The ROS tests seem to pass now, but Pixi is somehow reporting a significantly different position now? I wonder if this is a mujoco 3.4 vs. 3.5 (or 3.8 now?) mismatch coming into play now :upside_down_face: 

From my Claude automated run:

>   Version matrix (your 3.4 hunch was right, the other side is 3.8 not 3.5):
>  - Local kilted build: vendored MuJoCo 3.4.0 — the env your expected pose values came from
 > - Pixi/CI build: conda libmujoco 3.8.1 (the lock's mujoco 3.10.0 pypi wheel is only the Python bindings used by the MJCF-generation scripts, not the sim)
>
 > The CI numbers say "different physics", not "flake": CI consistently measured x=1.87526, z=−0.48850, qy=0.00244 vs expected x=1.8678, z=−0.5162, qy=0.0099. The z difference (0.0277) is just over your 0.025 tolerance — the body settles ~3 cm higher and nearly flat under 3.8.1. If my 5 pixi runs land on those same values with tiny spread (like the 17 identical kilted runs), that confirms it's a deterministic per-version resting pose, and no tolerance you'd want to keep would honestly cover both.

I could modify the tests to have two sets of expected poses and either will work? But still not the most satisfying... but it works!

mujoco_ros2_control_tests/CMakeLists.txt
```diff
@@ -31,33 +31,41 @@ if(BUILD_TESTING)
...
     TIMEOUT 75
   )
 
-  add_launch_test(test/robot_launch_transmission_pid_test.py
+  add_ros_isolated_launch_test(test/robot_launch_transmission_pid_test.py
     TIMEOUT 75
   )
 
-  add_launch_test(test/site_velocity_launch_test.py
+  add_ros_isolated_launch_test(test/site_velocity_launch_test.py
```
saikishor, line 68: @scastro-nasa should we change to ros isolated launch tests in this same PR? or we should do it in a separate one?

scastro-nasa: as you prefer -- it's easy enough to extract out : https://github.com/ros-controls/mujoco_ros2_control/pull/234

saikishor: Thanks

saikishor (approved):

christianrauch:
I deleted my forked branch, but there seems to be a [`pose_test` upstream branch](https://github.com/ros-controls/mujoco_ros2_control/tree/pose_test) left. Can you remove that?

Also, can you enable in the settings that merged branches are deleted automatically? The repo currently has 17 branches, of which probably a couple are left over from PRs and could be cleaned up.

scastro-nasa:
I think this is enabled, I accidentally pushed a `pose_test` branch to the main upstream instead of your fork yesterday and it's hard to remember to clean those up. Done!

PR #233, opened 2026-07-16
Add service to set the free joint body state

saikishor:
## Description

Fixes # (issue)

Yes, Claude for the final refinement and adding support for multi objects

## TODOs

- [ ] Fork the repository.


mujoco_ros2_control/docs/hardware_interface.rst
```diff
@@ -306,6 +308,38 @@ Services
       # Step the simulation forward by 100 physics steps
       ros2 service call /ros2_control_node/step_simulation mujoco_ros2_control_msgs/srv/StepSimulation "{steps: 100}"
 
+``~/set_free_joint_state`` (``mujoco_ros2_control_msgs/srv/SetFreeJointState``)
+   Directly sets the pose and velocity of a MuJoCo free-joint object (e.g. a manipulable prop),
+   identified by the name of the body its free joint drives. Useful for teleporting or resetting
+   an object's pose, similar to Gazebo's ``set_entity_state``.
```
scastro-nasa: nit: I'd omit the "similar to Gazebo" part, just adds text.

mujoco_ros2_control/docs/hardware_interface.rst
```diff
@@ -306,6 +308,38 @@ Services
...
+   - ``name`` (``string``): name of the MuJoCo body driven by the target free joint.
+   - ``pose`` (``geometry_msgs/Pose``): desired pose. An unset ``orientation`` defaults to
+     identity. Interpreted in the world frame unless ``reference_frame`` is set.
+   - ``twist`` (``geometry_msgs/Twist``): desired world-frame velocity. Left at its default
+     (all-zero), the object comes to rest at the new pose. Always interpreted in the world
+     frame, regardless of ``reference_frame``.
+   - ``reference_frame`` (``string``): optional name of another MuJoCo body. If empty
+     (default), ``pose`` is interpreted directly in the world frame. If set, ``pose`` is
+     composed onto that body's current world pose, letting you place an object relative to a
+     link instead of computing its world pose yourself.
```
scastro-nasa: I was initially thinking "why not a `PoseStamped` so you don't need the `reference_frame` field?"

But then the intention would be the same reference frame applies to both pose and twists right? Which is against what the doc says.

So I would recommend:

1. `PoseStamped` + `Twist` with the current behavior
2. `Pose` + `Twist` + `string` if the twist can also be expressed in the other reference frame

caveat: if we want to later support non-world frame twists, then we could also decide to stick with 2. so we don't have to modify the interfaces.

scastro-nasa (commented):
Nice, this will be useful!

One high-level question though: If someone wants to reset multiple bodies, it would be kind of annoying to have to call the same service multiple times, and more importantly have to claim the sim mutex and run mj_forward multiple times.

So is it worth making the service interface and the underlying function actually accept a list of body names / poses / twists for this use case? You can pass a single-element list to change 1 body anyhow.

It would additionally help serve a use case we have coming up at NASA, which is that we want to reset the sim with the reset service, but instead of default poses or keyframes, actually using a list of randomized free-body and joint states. Whether we modify the reset service to utilize whatever backend you make here, or we simply decide to call the reset service + set free body state afterwards, that's up to us.

Thoughts?

mujoco_ros2_control/docs/hardware_interface.rst
saikishor: Yeah, I want to support option 2, that's why I didn't go with the first one. Thanks

saikishor:
> So is it worth making the service interface and the underlying function actually accept a list of body names / poses / twists for this use case? You can pass a single-element list to change 1 body anyhow.

You got a point

mujoco_ros2_control/src/mujoco_simulation.cpp
```diff
@@ -1001,6 +1009,110 @@ void MujocoSimulation::step_simulation_callback(
...
+  mj_data_->qvel[qvel_adr + 1] = twist.linear.y;
+  mj_data_->qvel[qvel_adr + 2] = twist.linear.z;
+
+  // Angular velocity
+  mj_data_->qvel[qvel_adr + 3] = twist.angular.x;
+  mj_data_->qvel[qvel_adr + 4] = twist.angular.y;
+  mj_data_->qvel[qvel_adr + 5] = twist.angular.z;
+
+  // Recompute derived quantities (xpos, xmat, sensor data, etc.) from the new state.
+  mj_forward(mj_model_, mj_data_);
```
eholum-nasa: We should probably update the data snapshots, right? Unlikely there's any control that cares, but I'm not sure what the consumers will see depending on timing...

```
refresh_data_snapshot();
publish_control_state();
```

mujoco_ros2_control/src/mujoco_simulation.cpp
```diff
@@ -1001,6 +1009,110 @@ void MujocoSimulation::step_simulation_callback(
...
+      qpos_adr = mj_model_->jnt_qposadr[i];
+      qvel_adr = mj_model_->jnt_dofadr[i];
+      break;
+    }
+  }
+
+  if (qpos_adr == -1)
+  {
+    error_message = "Body '" + body_name + "' is not driven by a free joint.";
+    RCLCPP_WARN(get_logger(), "%s", error_message.c_str());
```
ndunkelb-nasa: It seems like you are relying on the failure state to print these error messages? So this one double prints, when running. So it can just go away.

mujoco_ros2_control/docs/hardware_interface.rst
ndunkelb-nasa: I guess I am a bit confused. I would agree with @scastro-nasa 's comment. Later on adding optional support for making the twist also in the pose frame seems a bit strange. Then would that mean the current behavior would change. Or would you still give the option to users of if they want it to be world frame or not? Obviously a different option would be a `PoseStamped` and a `TwistStamped` and then users could choose for both, and if they leave the header empty, it just defaults to world? Seems like that would accomplish what you are going for and be more explicit. 

Otherwise, I would just support the reference frame being applied to both at this point.

ndunkelb-nasa (commented):
Generally looks good to me, and worked in the cases that I tested! I'm good with this pending resolution of other discussions, especially making the service take a vector. 

After this, I think we might be interested in resetting specific non-free joints to specific states, although I know that will be more complicated with having to manage the control states and such.

mujoco_ros2_control/src/mujoco_simulation.cpp
saikishor: error_message variable is the one for the service client on the user side, and the logger is on the simulation end.

mujoco_ros2_control/docs/hardware_interface.rst
saikishor: I'll see if I can add support to this

ndunkelb-nasa: Open to discussion on this! Just trying to understand the use case where you have it now, and what you are wanting to do with this in the future, and make sure that change stays clear to end users

mujoco_ros2_control/tests/test_mujoco_simulation.cpp
```diff
@@ -369,6 +383,386 @@ TEST_F(MujocoSimulationTest, ResetWorldTest)
       << "Time should advance after unpausing post-reset";
 }
 
+TEST_F(MujocoSimulationTest, SetFreeJointStateSetsPoseAndVelocity)
```
scastro-nasa, line 387: question: do we want to actually call the service here, or do we want to use the `set_free_joint_states()` function directly without middleware in the mix?

mujoco_ros2_control/tests/test_mujoco_simulation.cpp
```diff
@@ -369,6 +383,386 @@ TEST_F(MujocoSimulationTest, ResetWorldTest)
...
+  EXPECT_NEAR(sim_->data()->qpos[qpos_adr + 0], 0.0, 1e-9);
+  EXPECT_NEAR(sim_->data()->qpos[qpos_adr + 1], 0.0, 1e-9);
+  EXPECT_NEAR(sim_->data()->qpos[qpos_adr + 2], 0.0, 1e-9);
+  // world_quat = ref_quat * identity = ref_quat = (cos45, 0, sin45, 0)
+  EXPECT_NEAR(sim_->data()->qpos[qpos_adr + 3], std::sqrt(0.5), 1e-9);  // w
+  EXPECT_NEAR(sim_->data()->qpos[qpos_adr + 4], 0.0, 1e-9);             // x
+  EXPECT_NEAR(sim_->data()->qpos[qpos_adr + 5], std::sqrt(0.5), 1e-9);  // y
+  EXPECT_NEAR(sim_->data()->qpos[qpos_adr + 6], 0.0, 1e-9);             // z
+  // twist stayed in the world frame, unaffected by the pose's reference frame.
+  EXPECT_NEAR(sim_->data()->qvel[qvel_adr + 0], 5.0, 1e-9);
```
scastro-nasa: nit: consider creating a variable for the tolerance
(same in the other tests below this)

scastro-nasa (commented):
A few small questions but otherwise LGTM.

Also obligatory request to see if it's possible to do a verbosity reduction pass to squash Claude's tendencies... I had to do the same for my last PR because it really likes to yap

mujoco_ros2_control/tests/test_mujoco_simulation.cpp
saikishor: As you guys prefer, but the service way is more like the user end functionality

scastro-nasa: I don't have a strong preference; it's more about minimizing flakiness, but if this is reliable enough then it's fine

scastro-nasa:
I think @ndunkelb-nasa has been testing this branch and agreed to give it the final :+1: -- we'll get to you soon!

saikishor:
> I think @ndunkelb-nasa has been testing this branch and agreed to give it the final 👍 -- we'll get to you soon!

No rush at all :)

saikishor:
Once this is merged, I would like to open a PR in the plugins side to publish the information of the available free joint bodies and their respective pose and twist

ndunkelb-nasa (approved):
LGTM! Tested and delivers!

saikishor:
> LGTM! Tested and delivers!

Thank you

PR #239, opened 2026-07-20
Make CI tests more robust

### LINKED ISSUE #236: Test3dLidarSyncAsync fails
**Describe the bug**

```
    [  FAILED  ] Mujoco3dLidarPluginTest.Test3dLidarSyncAsync
    
     1 FAILED TEST
    -- run_test.py: return code 1
    -- run_test.py: inject classname prefix into gtest result file '/home/runner/work/ros2_control_ci/ros2_control_ci/.work/target_ws/build/mujoco_ros2_control_plugins/test_results/mujoco_ros2_control_plugins/test_3d_lidar_plugin.gtest.xml'
    -- run_test.py: verify result file '/home/runner/work/ros2_control_ci/ros2_control_ci/.work/target_ws/build/mujoco_ros2_control_plugins/test_results/mujoco_ros2_control_plugins/test_3d_lidar_plugin.gtest.xml'
  >>>
build/mujoco_ros2_control_plugins/test_results/mujoco_ros2_control_plugins/test_3d_lidar_plugin.gtest.xml: 3 tests, 0 errors, 1 failure, 0 skipped
- mujoco_ros2_control_plugins.Mujoco3dLidarPluginTest Test3dLidarSyncAsync
  <<< failure message
    unknown file
    C++ exception with description "cannot store a negative time point in rclcpp::Time" thrown in the test body.
  >>>

```

**To Reproduce**
https://github.com/ros-controls/ros2_control_ci/issues/861

 - OS: ubuntu
 - Version humble

eholum-nasa:
## Description

Fixes #236 and #216

* Lidar should instantiate the timestamps to a valid, non nan number so that we don't get the negative value if the first update comes before the plugin as finished its compute.
* Rendering contexts in CI are... difficult. For now we ensure that we're using software rendering and llvmpipe in the running image, hopefully that makes mesa more predictable for the camera tests. I also added subscriber connection checks since we've seen errors there, too.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,6 +61,8 @@ jobs:
         run: |
           docker run --rm \
             -e EGL_PLATFORM=surfaceless \
+            -e LIBGL_ALWAYS_SOFTWARE=1 \
+            -e MESA_LOADER_DRIVER_OVERRIDE=llvmpipe \
             ${{ steps.meta.outputs.tags }} \
             bash -c "
               colcon test &&
@@ -105,6 +107,8 @@ jobs:
             -w /workspace \
             -e SCCACHE_DIR=/workspace/.sccache \
             -e EGL_PLATFORM=surfaceless \
+            -e LIBGL_ALWAYS_SOFTWARE=1 \
+            -e MESA_LOADER_DRIVER_OVERRIDE=llvmpipe \
             mujoco_ros2_control_pixi:ci \
             bash -c "
               pixi run setup-colcon && \
--- a/mujoco_ros2_control_plugins/src/mujoco_3d_lidar_plugin.cpp
+++ b/mujoco_ros2_control_plugins/src/mujoco_3d_lidar_plugin.cpp
@@ -305,9 +305,11 @@ void Mujoco3dLidarPlugin::update(const mjModel* /* model */, mjData* data)
 {
   for (Lidar3dConfig& lidar : lidar_sensors_)
   {
-    // Check for new sensor data, and skip if not yet computed or if it is stale
+    // Check for new sensor data, and skip if not yet computed, stale, or otherwise
+    // invalid. This will explicitly skip the 0th timepoint, since that is not a valid
+    // point for a scan, anyway.
     mjtNum compute_time = data->plugin_state[lidar.plugin_stateadr];
-    if (compute_time < 0.0 || compute_time == lidar.last_published_time)
+    if (!(compute_time > lidar.last_published_time))
     {
       continue;
     }
--- a/mujoco_ros2_control_plugins/test/test_camera_plugin.cpp
+++ b/mujoco_ros2_control_plugins/test/test_camera_plugin.cpp
@@ -30,9 +30,14 @@
 
 #include "camera_plugin.hpp"
 
+using namespace std::chrono_literals;
+
 class CameraPluginTest : public ::testing::Test
 {
 protected:
+  static constexpr auto WAIT_TIMEOUT = 2s;
+  static constexpr auto POLL_INTERVAL = 20ms;
+
   static void SetUpTestSuite()
   {
     if (!rclcpp::ok())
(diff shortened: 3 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -61,6 +61,8 @@ jobs:
         run: |
           docker run --rm \
             -e EGL_PLATFORM=surfaceless \
+            -e LIBGL_ALWAYS_SOFTWARE=1 \
+            -e MESA_LOADER_DRIVER_OVERRIDE=llvmpipe \
```
eholum-nasa, line 65: Not sure if different runner images will do different things but this should at least make the mesa env as consistent as possible

eholum-nasa:
Manually ran tests 6 times with no failures. Maybe that means it's fixed forever? Regardless, left the todo in the camera tests in case we ever just want to skip the rendering is borked failure mode.

mujoco_ros2_control_plugins/src/mujoco_3d_lidar_plugin.cpp
```diff
@@ -305,9 +305,11 @@ void Mujoco3dLidarPlugin::update(const mjModel* /* model */, mjData* data)
 {
   for (Lidar3dConfig& lidar : lidar_sensors_)
   {
-    // Check for new sensor data, and skip if not yet computed or if it is stale
+    // Check for new sensor data, and skip if not yet computed, stale, or otherwise
+    // invalid. This will explicitly skip the 0th timepoint, since that is not a valid
+    // point for a scan, anyway.
     mjtNum compute_time = data->plugin_state[lidar.plugin_stateadr];
-    if (compute_time < 0.0 || compute_time == lidar.last_published_time)
+    if (!(compute_time > lidar.last_published_time))
```
scastro-nasa, line 312: nit: use <= to avoid negation?

mujoco_ros2_control_plugins/test/test_camera_plugin.cpp
```diff
@@ -109,6 +109,32 @@ class CameraPluginTest : public ::testing::Test
...
+  template <typename... Subs>
+  void wait_for_subscriber_match(const Subs&... subs)
+  {
+    auto all_matched = [&]() { return ((subs->get_publisher_count() > 0) && ...); };
+    for (int i = 0; i < 50 && !all_matched(); ++i)
+    {
+      std::this_thread::sleep_for(std::chrono::milliseconds(20));
+    }
+    ASSERT_TRUE(all_matched()) << "Subscribers never matched publishers within 2 s";
+  }
```
scastro-nasa: lots of hard-coded times here -- of particular concern is the comment that publishers didn't match within 2s, but 20 ms * 50 iterations = 1 s.

Stuff like this is just prone to go out of sync. I recommend either making the assert say nothing about time, or to automatically calculate the number of iterations from variables for the delay time and total wait time.

mujoco_ros2_control_plugins/test/test_camera_plugin.cpp
```diff
@@ -178,7 +204,9 @@ TEST_F(CameraPluginTest, InitAndPublish)
   mujoco_ros2_control_plugins::CameraPlugin plugin;
   // GLFW is not initialized, and No OpenGL framebuffer will be available so we make the init fall back on to EGL.
   EXPECT_TRUE(plugin.init(plugin_node_, model_, data_, []() { return 0; }));
-  std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
+  // Wait for the rendering thread to come up... If it doesn't we're going to skip this
+  wait_for_rendering(plugin);
```
scastro-nasa: is this comment correct? if `wait_for_rendering()` times out it asserts and fails the test, right?

mujoco_ros2_control_plugins/test/test_camera_plugin.cpp
```diff
@@ -282,12 +312,11 @@ TEST_F(CameraPluginTest, PolledCameraPublishesOncePerTrigger)
...
       "/camera_plugin/poll_cam/color", 10, [&](sensor_msgs::msg::Image::SharedPtr) { ++image_count; });
 
-  // Wait for the rendering thread to come up (so a forced render is safe).
-  for (auto i = 0; i < 50 && !plugin.is_rendering_available(); ++i)
-  {
-    std::this_thread::sleep_for(std::chrono::milliseconds(20));
-  }
-  ASSERT_TRUE(plugin.is_rendering_available()) << "Rendering context unavailable in this environment";
+  // Wait for the rendering thread to come up... If it doesn't we're going to skip this
+  wait_for_rendering(plugin);
```
scastro-nasa: same deal here

mujoco_ros2_control_plugins/src/mujoco_3d_lidar_plugin.cpp
eholum-nasa: yeah... I _think_ what was happening was the time was initialized as a `nan` or some other nonsense, so when the `update` loop was triggered before the sensor had a chance to read this check (`compute_time < 0.0`) was failing. That _shouldn't_ matter now, but the explicit greater than check catches that case as well. Extra confidence for an annoying `!` ?

mujoco_ros2_control_plugins/test/test_camera_plugin.cpp
eholum-nasa: Yeah I should just remove some of the magic numbers here... stand by

mujoco_ros2_control_plugins/test/test_camera_plugin.cpp
eholum-nasa: Oh right I changed it... will adjust.

mujoco_ros2_control_plugins/test/test_camera_plugin.cpp
eholum-nasa: Done and done.

eholum-nasa:
@scastro-nasa I think these are fixed forever. Ran another few times and haven't seen rendering or timing failures from the camera tests... at least.

scastro-nasa (approved):
fixed forever! ™️

PR #241, opened 2026-07-21
Fix transmissions integration working with multiple actuators

saikishor:
## Description

This PR allows us to control the full complex model of Kangaroo with 64 passive joints and 12 active joints on the MuJoCo end and then the ROS Side using a simplified model controlling the complex model with custom in-house transmissions. This would represent the same way how we opeate with the robot.

Nope, not at all.

No

### DIFF
--- a/mujoco_ros2_control/src/mujoco_system_interface.cpp
+++ b/mujoco_ros2_control/src/mujoco_system_interface.cpp
@@ -143,28 +143,27 @@ ActuatorType getActuatorType(const mjModel* mj_model, int mujoco_actuator_id)
  */
 int get_actuator_id(const std::string& actuator_name, const mjModel* mj_model)
 {
-  int mujoco_actuator_id = mj_name2id(mj_model, mjtObj::mjOBJ_JOINT, actuator_name.c_str());
-  if (mujoco_actuator_id == -1)
+  // First interpret the name as a JOINT and look for an actuator that drives it.
+  const int joint_id = mj_name2id(mj_model, mjtObj::mjOBJ_JOINT, actuator_name.c_str());
+  if (joint_id != -1)
   {
-    RCLCPP_DEBUG(rclcpp::get_logger("MujocoSystemInterface"), "Failed to find the actuator : '%s' in the MuJoCo model",
-                 actuator_name.c_str());
-  }
-
-  // Try to locate the matching actuator id for the joint, if available
-  for (int i = 0; i < mj_model->nu; ++i)
-  {
-    // If it is the correct type and matches the joint id, we're done
-    if (mj_model->actuator_trntype[i] == mjTRN_JOINT && mj_model->actuator_trnid[2 * i] == mujoco_actuator_id)
+    for (int i = 0; i < mj_model->nu; ++i)
     {
-      mujoco_actuator_id = i;
-      break;
+      if (mj_model->actuator_trntype[i] == mjTRN_JOINT && mj_model->actuator_trnid[2 * i] == joint_id)
+      {
+        return i;  // Found the actuator attached to that joint.
+      }
     }
+    // A MuJoCo joint of this name exists but no actuator drives it (e.g. a passive closed-loop joint whose
+    // name coincides with a ros2_control joint). Do NOT return the joint id as if it were an actuator id --
+    // fall through and try to resolve the name as an actuator instead.
   }
 
-  // Is no mapping was found, try fallback to looking for an actuator with the same name as the joint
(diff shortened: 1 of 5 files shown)


saikishor:
There are minor logic changes but more formatting changes in this PR

danielcostanzi18:
Follow some screenshots showing the mujoco simulation running while the robot is being controlled in position at he joint level, with the transmissions properly loaded and running, highlighting the equality constraints. At last you can see a screenshot of the visualization in rviz.
Screenshot from 2026-07-21 17-56-28
Screenshot from 2026-07-21 17-56-12
Screenshot from 2026-07-21 17-55-57

saikishor:
Thanks @danielcostanzi18

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -143,73 +143,72 @@ ActuatorType getActuatorType(const mjModel* mj_model, int mujoco_actuator_id)
...
     }
   }
 
-  // Is no mapping was found, try fallback to looking for an actuator with the same name as the joint
-  mujoco_actuator_id = mujoco_actuator_id == -1 ? mj_name2id(mj_model, mjtObj::mjOBJ_ACTUATOR, actuator_name.c_str()) :
-                                                  mujoco_actuator_id;
-  return mujoco_actuator_id;
+  // Otherwise interpret the name directly as an actuator name.
+  const int actuator_id = mj_name2id(mj_model, mjtObj::mjOBJ_ACTUATOR, actuator_name.c_str());
+  RCLCPP_DEBUG_EXPRESSION(rclcpp::get_logger("MujocoSystemInterface"), actuator_id == -1,
```
eholum-nasa: Should this be a warning? I think it only happens on initial construction, right so will it be noisy?

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -1462,116 +1483,135 @@ void MujocoSystemInterface::register_urdf_joints(const hardware_interface::Hardw
...
-          // Direct position control enabled for position actuator
-          actuator_it->is_position_control_enabled = true;
-        }
-        else if (actuator_it->actuator_type == ActuatorType::VELOCITY ||
-                 actuator_it->actuator_type == ActuatorType::MOTOR ||
-                 actuator_it->actuator_type == ActuatorType::CUSTOM)
+      // Apply the control mode to every MuJoCo actuator
+      for (auto* actuator_it : joint_actuators)
+      {
+        const std::string actuator_name = mj_id2name(simulation_->model(), mjOBJ_ACTUATOR, actuator_it->mj_actuator_id);
```
eholum-nasa, line 1499: Some of these actuators get labeled as "unnamed" in the registration function. Is it possible for an unnamed actuator to be included in a transmissions interface? This is hypothetical since based on the example it seems like that _shouldn't_ be how these are setup.

That said... should we just store the mujoco actuator name in the ActuatorData structure during registration? Then we would have the list of actuators and the joints they are attached to after registration, right?

eholum-nasa (commented):
Awesome. This basically expands transmissions to the world of "real" hardware, right? We rarely use them so I'm not super familiar with how everything works under the hood. 

I just had the question if there was a way to simplify the setup of the transmission interfaces by saving the actuator data when we do the registration. Otherwise this is looking pretty good.

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: You are right, it needs to be warn. It will be only noisy, if things are not found, so I think it is okay to leave it with WARN_EXPRESSION

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: I agree with you, maybe we should handle it in a different PR?

saikishor:
> I just had the question if there was a way to simplify the setup of the transmission interfaces by saving the actuator data when we do the registration. Otherwise this is looking pretty good.

Some things can be refined for sure. It might changes things under the hood, so maybe we can leave it for a different PR?

mujoco_ros2_control/src/mujoco_system_interface.cpp
eholum-nasa: Yeah there's probably some hw interface construction things that should be cleaned up...

saikishor: Thanks. Do you agree to do it in a different PR? To keep it seperate

eholum-nasa (approved):
Seems to work for me!

PR #244, opened 2026-07-22
[Feature] modify elements of geom attribute

Ortisa:
This pull request introduces the ability to modify geom parameters of the MuJoCo model via the <modify_element> tag, allowing physical/collision parameters such as friction, condim, solref, solimp (and generally any supported attribute) to be specified without having to manually edit the generated XML.

For example:
```
<modify_element type="geom" mesh="fingertip" class="collision"
                 friction="1.0 0.1 0.005" condim="6" solref="0.001 1" solimp="0.95 0.99 0.001"/>
```
How matching works:
Match by mesh + class (geoms only): the combination of mesh and class (visual/collision) is used as the identifying key, since geoms are differentiated precisely through these two attributes.

Handling decomposed meshes:
When a mesh is decomposed into multiple collision sub-meshes (e.g. via convex decomposition), the generated names follow the pattern `<base_name>_collision_<index_number>`. In this case, it is not necessary to add a modify_element tag for every single decomposed mesh: it only needs to be specified once using the base name. This way, the modification is automatically applied to all decomposed meshes derived from the same original mesh.


mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
```diff
@@ -1124,10 +1143,25 @@ def add_modifiers(dom, modify_element_dict):
...
+
+            # match by explicit name
+            name_key = (element.tagName, element.getAttribute("name"))
+            if name_key in modify_element_dict:
+                attr_dict = modify_element_dict[name_key]
+
+            # match by mesh and class only for geoms
+            elif element.tagName == "geom" and element.hasAttribute("mesh") and element.hasAttribute("class"):
+                raw_mesh_name = element.getAttribute("mesh")
+                geom_class = element.getAttribute("class")
```
scastro-nasa: similar comment here though with the xml etree library: is there a way we can avoid 2 separate checks on the attribute and just getting once + checking the result?

EDIT: oops I didn't post the first comment, refer to it below

scastro-nasa (commented):
This looks great and is well-tested. I have some style suggestions to consider which I'd appreciate pulling in.

mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
```diff
@@ -620,20 +620,39 @@ def get_processed_mujoco_inputs(processed_inputs_element):
...
-                raise ValueError("'name' and 'type' must be in the attributes of a 'modify_element' tag!")
 
-            # remove the name and type entries because those will be used as the key in the returned dict
-            element_name = attr_dict["name"]
+            # we must have a type element
+            if "type" not in attr_dict:
+                raise ValueError("'type' must be in the attributes of a 'modify_element' tag!")
+
+            # remove the type entry because this will be used as the key in the returned dict
             element_type = attr_dict["type"]
```
scastro-nasa: Minor style suggestion, and this did predate your changes to this file, so apologies that it fell on this PR...

Doing something like this does 2 separate python dict lookups:

```python
if "type" not in attr_dict:  # lookup 1
    raise ValueError("type not found")
element_type = attr_dict["type"]  # lookup 2
```

I would suggest instead doing:

```python
element_type = attr_dict.get("type")  # just the one lookup!
if element_type is None:
    raise ValueError("type not found")
```

it also saves you from having to pre-assign certain variables to `None` before trying to extract them, so it might additionally cut the code down a little.

Ortisa: yes, it is definitely more efficient!

scastro-nasa (approved):

saikishor (approved):

PR #247, opened 2026-07-23
MuJoCo free joints state publisher plugin

saikishor:
## Description

This PR offers a plugin that can publish the data of the free joints in the simulation scene. Very useful to know the state and also for the testing purposs.

NO

YES, Claude

### DIFF
--- a/mujoco_ros2_control_plugins/src/free_joint_state_publisher_plugin.cpp
+++ b/mujoco_ros2_control_plugins/src/free_joint_state_publisher_plugin.cpp
@@ -1,0 +1,236 @@
+// Copyright 2026 PAL Robotics S.L.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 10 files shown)


mujoco_ros2_control_plugins/src/free_joint_state_publisher_plugin.cpp
```diff
@@ -0,0 +1,236 @@
...
+  if (publish_rate <= 0.0)
+  {
+    RCLCPP_ERROR(logger_, "publish_rate must be > 0, got %f.", publish_rate);
+    return false;
+  }
+
+  if (!collect_free_joint_entries(body_names))
+  {
+    return false;
+  }
```
scastro-nasa, line 118: fine without an error because `collect_free_joint_entries()` already logs some right?

scastro-nasa (approved):
LGTM!

mujoco_ros2_control_plugins/src/free_joint_state_publisher_plugin.cpp
saikishor: Yup, but better publish everything user asked or fail. If you want every object, just don't specify anything and it will work out of the box

PR #254, opened 2026-07-27
Add base twist plugin for commanding the mobile base

saikishor:
## Description

This PR adds a new plugin that can control the base of the robot by commanding directly the floating joint it is connected to directly; it resembles the ROS planar move plugin that Gazebo offers. This is very reliable as the users doesn't have to spend time modeling their wheels and contact friction etc, and this basically helps to fast prototype

NO

Claude

### Additional Information

[TIAGo Pro with Mobile base twist plugin](https://github.com/user-attachments/assets/6acd0ffc-5278-48f0-a8c1-a224ca464926)

### DIFF
--- a/mujoco_ros2_control/src/mujoco_simulation.cpp
+++ b/mujoco_ros2_control/src/mujoco_simulation.cpp
@@ -24,6 +24,7 @@
 #include <unistd.h>
 #include <cerrno>
 #include <chrono>
+#include <cmath>
 #include <cstdint>
 #include <cstdio>
 #include <cstdlib>
@@ -32,6 +33,7 @@
 #include <fstream>
 #include <future>
 #include <iostream>
+#include <limits>
 #include <memory>
 #include <new>
 #include <stdexcept>
@@ -715,6 +717,7 @@ bool MujocoSimulation::initialize(rclcpp::Node::SharedPtr node, const std::strin
     xfrc_plugin_desired_.assign(6 * mj_model_->nbody, 0.0);
     xfrc_viewer_capture_.assign(6 * mj_model_->nbody, 0.0);
     xfrc_last_written_.assign(6 * mj_model_->nbody, 0.0);
+    qvel_override_staged_.assign(mj_model_->nv, std::numeric_limits<mjtNum>::quiet_NaN());
 
     if (mj_data_ && snapshot_write_ && snapshot_read_)
     {
@@ -860,6 +863,7 @@ void MujocoSimulation::reset_world_state(bool fill_initial_state)
     std::fill(ctrl_staged_.begin(), ctrl_staged_.end(), 0.0);
     std::fill(qfrc_applied_staged_.begin(), qfrc_applied_staged_.end(), 0.0);
     std::fill(xfrc_plugin_desired_.begin(), xfrc_plugin_desired_.end(), 0.0);
+    std::fill(qvel_override_staged_.begin(), qvel_override_staged_.end(), std::numeric_limits<mjtNum>::quiet_NaN());
   }
   std::fill(xfrc_viewer_capture_.begin(), xfrc_viewer_capture_.end(), 0.0);
   std::fill(xfrc_last_written_.begin(), xfrc_last_written_.end(), 0.0);
@@ -1243,6 +1247,7 @@ void MujocoSimulation::apply_control_data(mjData* control_data)
   mju_copy(ctrl_staged_.data(), control_data->ctrl, static_cast<int>(mj_model_->nu));
   mju_copy(qfrc_applied_staged_.data(), control_data->qfrc_applied, static_cast<int>(mj_model_->nv));
   mju_copy(xfrc_plugin_desired_.data(), control_data->xfrc_applied, 6 * static_cast<int>(mj_model_->nbody));
+  mju_copy(qvel_override_staged_.data(), control_data->qvel, static_cast<int>(mj_model_->nv));
   control_inputs_staged_ = true;
 }
 
@@ -1297,6 +1302,15 @@ void MujocoSimulation::apply_staged_control_inputs()
   mju_copy(mj_data_->xfrc_applied, xfrc_viewer_capture_.data(), nbody6);
   mju_addTo(mj_data_->xfrc_applied, xfrc_plugin_desired_.data(), nbody6);
   mju_copy(xfrc_last_written_.data(), mj_data_->xfrc_applied, nbody6);
+
+  // Apply plugin-requested velocity overrides as direct qvel writes
+  for (int i = 0; i < static_cast<int>(mj_model_->nv); ++i)
+  {
+    if (std::isfinite(qvel_override_staged_[i]))
+    {
+      mj_data_->qvel[i] = qvel_override_staged_[i];
+    }
+  }
 }
 
 // simulate in background thread (while rendering in main thread)
--- a/mujoco_ros2_control_demos/doc/tutorials.rst
+++ b/mujoco_ros2_control_demos/doc/tutorials.rst
@@ -88,9 +88,17 @@ Tutorial 5: Base Velocity Plugin
 
 Demonstrates driving a mobile/floating-base robot with ``BaseVelocityPlugin``: a free-floating,
 wheeled chassis (MJCF ``<freejoint>``) carrying a 1-DOF arm, driven from a ``cmd_vel`` topic via a
-velocity servo. The wheels and ground have zero friction, so the ground can never exert any
-tangential force on the base -- only the servo can change its velocity. Collisions with the wall
-in the scene are still handled by the physics engine as normal.
+hard velocity override applied directly to the chassis's free-joint ``qvel`` every cycle. The
+wheels and ground have zero friction, which is now largely cosmetic -- propulsion no longer goes
+through wheel-ground contact at all, so it does not depend on friction either way.
+
+.. warning::
+
+   Because the override is kinematic, it outranks the contact solver: the wall in the scene will
+   **not** stop the base -- it will push through or climb it rather than being stopped by contact,
+   since the commanded velocity is reasserted every cycle regardless of collisions. This is the
+   trade-off for exact, disturbance-immune velocity tracking; see the ``BaseVelocityPlugin``
+   documentation for details.
 
 .. code-block:: bash
 
@@ -99,10 +107,11 @@ in the scene are still handled by the physics engine as normal.
    # In another terminal, drive the base:
    ros2 topic pub /cmd_vel geometry_msgs/msg/Twist "{linear: {x: 0.5}, angular: {z: 0.2}}" --rate 10
 
-**Disturbance rejection**: swinging the mounted arm creates a reaction force/torque on the base
+**Disturbance immunity**: swinging the mounted arm creates a reaction force/torque on the base
 through the joint, unrelated to ground friction, since it acts directly between the two bodies.
-``BaseVelocityPlugin`` measures the base's own velocity regardless of what disturbs it, so with no
-``cmd_vel`` active the base should stay close to put while the arm moves, rather than drifting off.
+Because ``BaseVelocityPlugin`` overrides the base's driven DOFs directly rather than servoing
+against them, this reaction has *no* effect on the base's velocity at all -- not approximately,
+exactly none -- so with no ``cmd_vel`` active the base stays exactly in place while the arm moves.
 The arm's joint is range-limited to a back-and-forth sweep rather than a continuous spin:
 
 .. code-block:: bash
@@ -110,9 +119,9 @@ The arm's joint is range-limited to a back-and-forth sweep rather than a continu
    ros2 topic pub /arm_position_controller/commands std_msgs/msg/Float64MultiArray "data: [0.7]"
    ros2 topic pub /arm_position_controller/commands std_msgs/msg/Float64MultiArray "data: [-0.7]"
 
-**Key concepts:** ``BaseVelocityPlugin`` velocity servo, driving a MJCF ``<freejoint>`` body,
-disturbance rejection against reaction forces from a moving mounted joint, floating-base odometry
-(``odom_free_joint_name``).
+**Key concepts:** ``BaseVelocityPlugin`` free-joint velocity override, driving a MJCF
+``<freejoint>`` body, immunity to reaction forces from a moving mounted joint, floating-base
+odometry (``odom_free_joint_name``).
 
 **Resources:** ``demo_resources/scenes/scene_mobile_base.xml``, ``demo_resources/mobile_base/mobile_base.xml``,
 ``demo_resources/mobile_base/mobile_base.urdf``, ``config/mujoco_ros2_control_plugins_base_velocity.yaml``
--- a/mujoco_ros2_control_demos/doc/tutorials.rst
+++ b/mujoco_ros2_control_demos/doc/tutorials.rst
@@ -83,6 +83,43 @@ Demonstrates ``ros2_control`` transmissions with mechanical reduction ratios.
 **Resources:** ``demo_resources/robot/test_robot.urdf`` with ``use_transmissions:=true``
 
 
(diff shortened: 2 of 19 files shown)


saikishor:
I'm not sure of the approach I'm choosing here, right now what I do is the mjData that we pass to the update method of the plugin, I'm setting them to NaNs, and then if the plugin sets them to a value, then that's what I'm using to override the original values in the mjStep

mujoco_ros2_control/src/mujoco_simulation.cpp
```diff
@@ -1297,6 +1302,15 @@ void MujocoSimulation::apply_staged_control_inputs()
   mju_copy(mj_data_->xfrc_applied, xfrc_viewer_capture_.data(), nbody6);
   mju_addTo(mj_data_->xfrc_applied, xfrc_plugin_desired_.data(), nbody6);
   mju_copy(xfrc_last_written_.data(), mj_data_->xfrc_applied, nbody6);
+
+  // Apply plugin-requested velocity overrides as direct qvel writes
```
eholum-nasa, line 1399: This gives me pause, if a plugin needs to do this is it really a plugin? Could this work in a truly standalone manner by handling control data in the plugin itself? If not... is it really a "plugin"?

eholum-nasa (commented):
Funnily enough we've been considering this exact problem for our ridgeback. 

We do whole body motion planning with Phoebe which requires "joints" to attach the base to world (linear x, y, and a rotational yaw), so those joints existed anyway. To make the base into a magic carpet we just make those joints real and exclude contacts between the wheels, base, and floor. From there it's just a forward command controller fed from a mecanum drive controller -> joint command republisher. The full changes are in:

https://github.com/NASA-JSC-Robotics/phoebe_bridgeback/pull/49

Where the magic carpet commands are just handled [here](https://github.com/NASA-JSC-Robotics/phoebe_bridgeback/pull/49/changes#diff-6819d2c24fdbba00fdd8b75288f2208487e634b7296df505a74a3094b5c49cd7).

Would something like that support your use case? We've found it more useful for handling the sim2realness of an actual mobile base 

I feel like it is less invasive than adding another piece of overridable data for the plugin.

saikishor:
> We do whole body motion planning with Phoebe which requires "joints" to attach the base to world (linear x, y, and a rotational yaw), so those joints existed anyway. To make the base into a magic carpet we just make those joints real and exclude contacts between the wheels, base, and floor. From there it's just a forward command controller fed from a mecanum drive controller -> joint command republisher. The full changes are in:
> 
> [NASA-JSC-Robotics/phoebe_bridgeback#49](https://github.com/NASA-JSC-Robotics/phoebe_bridgeback/pull/49)
> 
> Where the magic carpet commands are just handled [here](https://github.com/NASA-JSC-Robotics/phoebe_bridgeback/pull/49/changes#diff-6819d2c24fdbba00fdd8b75288f2208487e634b7296df505a74a3094b5c49cd7).
> 
> Would something like that support your use case? We've found it more useful for handling the sim2realness of an actual mobile base
> 
> I feel like it is less invasive than adding another piece of overridable data for the plugin.

This might work too. However, I think there is quite some difference in the approach, for a beginner user modifying the XML to add new joints to make it work, instead with the proposed approach, adding a plugin to a yaml already gives him the flexibility to start and run without modifying much his simulation setup. I think it is a big win in this scenario. Moreover, Gazebo also offered similar kind of plugin in the past : https://github.com/ros-simulation/gazebo_ros_pkgs/blob/noetic-devel/gazebo_plugins/src/gazebo_ros_planar_move.cpp. I'm a bit aligned with the proposed approach for the simplicity on the user end.

mujoco_ros2_control/src/mujoco_simulation.cpp
saikishor: Yes, adding the control data to the plugin definitely helps, so I won't have to do this weird setup.

scastro-nasa:
I agree with @saikishor that the python republishing thing is pretty complicated for new users, but also with @eholum-nasa  that needing to modify `mujoco_simulation.cpp/.hpp` is kind of a code smell.

Is there a solution where we don't touch the mujoco_ros2_control source and instead create an actual C++ ROS 2 controller that can publish the base in this ideal form? Then for newcomers it's just a matter of activating this controller and publishing `/cmd_vel` to it.

saikishor:
> I agree with @saikishor that the python republishing thing is pretty complicated for new users, but also with @eholum-nasa  that needing to modify `mujoco_simulation.cpp/.hpp` is kind of a code smell.
> 
> Is there a solution where we don't touch the mujoco_ros2_control source and instead create an actual C++ ROS 2 controller that can publish the base in this ideal form? Then for newcomers it's just a matter of activating this controller and publishing `/cmd_vel` to it.

As of now, there is no easier way as the floating base joints are not exposed as controllable interfaces through ROS Controls interfaces. Even though we expose, we don't have a n out-of-the-box controller that does that work for you. I can rethink this part.

I agree that touching the mujoco_simulation file is not ideal, as per every implementation we come across limitations on that end.

eholum-nasa:
I do think our solution to use a "magic carpet controller" is a reasonable one,

https://github.com/NASA-JSC-Robotics/phoebe_bridgeback/pull/49/changes

I don't think those changes were too significant, a lot of them were just to handle that we have a crazy number of deployment modes to support for Phoebe. That said, the controller + the linear rails is a very straightforward way to end up with perfect odometry if that's what you're looking for? And it matches our existing control setup with the mecanum drive so it's still somewhat seamless from nav2's perspective?

saikishor:
> I do think our solution to use a "magic carpet controller" is a reasonable one,
> 
> https://github.com/NASA-JSC-Robotics/phoebe_bridgeback/pull/49/changes
> 
> I don't think those changes were too significant, a lot of them were just to handle that we have a crazy number of deployment modes to support for Phoebe. That said, the controller + the linear rails is a very straightforward way to end up with perfect odometry if that's what you're looking for? And it matches our existing control setup with the mecanum drive so it's still somewhat seamless from nav2's perspective?

I don't know If this is the way. I don't think it is right to impose everyone to set these linear rails just to be able to be able to track their base without slipping. I think it is better to leave it simple like a plugin no?. Even with this plugin, we are publishing to the same topic /cmd_vel, so everything should be working as it is supposed to work with the proposed approach in this PR right?

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
```diff
@@ -54,6 +54,12 @@ class MuJoCoROS2ControlPluginBase
    * affect the next simulation step.
    * @note This method will be called in a real-time thread, so it should avoid blocking operations and should be
    * efficient.
+   * @note data->qvel is NaN-filled before every plugin's update() runs this cycle. Write a finite value into any
+   * entry to request a hard velocity override on that DOF -- it bypasses the normal mass/contact dynamics for that
+   * DOF this step, overwriting mj_data_->qvel directly right before the next mj_step. Leave an entry NaN to keep its
+   * normal physics-driven value untouched. This is the only channel a plugin has to influence qvel: unlike
+   * ctrl/qfrc_applied/xfrc_applied, data->qvel here cannot be read for the body's actual velocity, since it is not
+   * restored to a real value until after every plugin's update() has run this cycle.
```
eholum-nasa, line 62: This feels like a large can of worms to open for the sake of free-flying magic carpets, especially when the engine already supports this through other approaches (like the rails). That said, I think we can all agree that MuJoCo's wheel/contact model makes this genuinely hard, and I don't disagree with providing a mechanism to make it easier.

If we're going to open this can of worms, I think it's worth revisiting an idea we've bumped up against a few times - expanding the plugin interface with a pre-step hook that gives plugins direct access to mj_data_ immediately before mj_step in the physics loop.

Something like:

```
virtual void update_pre_step(const mjModel* model, mjData* data)
{
  (void)model;
  (void)data;
}
```

If we threw that in and put the CB into the PhysicsLoop then plugins really would have full god-tier access to the sim. We could remove both the `qvel_override_staged_` and the  `xfrc_plugin_desired_` staging mechanisms. This would give plugins way more power, but IMO that would be the "right" thing to do rather than to continue digging very deep hooks deep into simulate. I am not the biggest fan of the use-case-specific staging vectors that we have to handle... so maybe that would make things easier for the next time a plugin needs access to a new field?

Thoughts? 

I think I would be down to move forward with this for now if we commit ourselves to a refactor like what I've proposed above rather than add any more one-offs (rule of three).

mujoco_ros2_control_plugins/doc/plugins.rst
```diff
@@ -285,6 +285,130 @@ ExternalWrench Parameters
...
+       since there is no ``qvel`` to override without one.
+   * - ``cmd_vel_topic``
+     - ``string``
+     - ``cmd_vel``
+     - Command topic name.
+   * - ``use_stamped_twist``
+     - ``bool``
+     - ``false``
+     - Subscribe to ``geometry_msgs/TwistStamped`` instead of ``geometry_msgs/Twist`` (e.g. for
+       Nav2, which publishes stamped twists by default in some configurations).
```
ndunkelb-nasa: I would add to this that if you set this to true, the header is thrown away, bc sometimes the assumption, especially with cmd_timeout might be that the stamp needs to be close enough to real time to work.

mujoco_ros2_control_plugins/src/base_velocity_plugin.cpp
```diff
@@ -0,0 +1,191 @@
...
+  std::lock_guard<std::mutex> lock(cmd_mutex_);
+  latest_cmd_ = { vx, vy, wz, node_->get_clock()->now(), true };
+}
+
+void BaseVelocityPlugin::update(const mjModel* /*model_arg*/, mjData* data)
+{
+  if (body_id_ < 0 || qvel_adr_ < 0)
+  {
+    return;
+  }
```
ndunkelb-nasa: I think you took care of these failure cases in `init()`, right?

mujoco_ros2_control_plugins/src/base_velocity_plugin.cpp
```diff
@@ -0,0 +1,191 @@
...
+  if (cmd_mutex_.try_lock())
+  {
+    cached_cmd_ = latest_cmd_;
+    cmd_mutex_.unlock();
+  }
+
+  // Step 2 - a stale (or never-received) command is treated as a zero-velocity
+  // command, i.e. the base is commanded to stop rather than coast on the last override.
+  double vx_cmd = 0.0, vy_cmd = 0.0, wz_cmd = 0.0;
+  if (cached_cmd_.valid)
```
ndunkelb-nasa: Seems like this is never not valid? Just set to true when initialized, then not ever set to false...

mujoco_ros2_control_demos/doc/tutorials.rst
```diff
@@ -83,6 +83,52 @@ Demonstrates ``ros2_control`` transmissions with mechanical reduction ratios.
...
+Tutorial 5: Base Velocity Plugin
+---------------------------------
+
+Demonstrates driving a mobile/floating-base robot with ``BaseVelocityPlugin``: a free-floating,
+wheeled chassis (MJCF ``<freejoint>``) carrying a 1-DOF arm, driven from a ``cmd_vel`` topic via a
+hard velocity override applied directly to the chassis's free-joint ``qvel`` every cycle. The
+wheels and ground have zero friction, which is now largely cosmetic -- propulsion no longer goes
+through wheel-ground contact at all, so it does not depend on friction either way.
+
+.. warning::
```
eholum-nasa, line 95: Nice, this explains the danger well!

mujoco_ros2_control_plugins/src/base_velocity_plugin.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+  if (use_stamped_twist)
+  {
+    twist_stamped_sub_ = node_->create_subscription<geometry_msgs::msg::TwistStamped>(
+        cmd_vel_topic, rclcpp::SystemDefaultsQoS(),
+        [this](const geometry_msgs::msg::TwistStamped& msg) { twistStampedCallback(msg); });
+  }
+  else
+  {
+    twist_sub_ = node_->create_subscription<geometry_msgs::msg::Twist>(cmd_vel_topic, rclcpp::SystemDefaultsQoS(),
```
eholum-nasa, line 100: My ignorance on ros2_controllers, but are there controllers for wheeled bases that ingest twists and not stamped twists?

mujoco_ros2_control_plugins/src/base_velocity_plugin.hpp
```diff
@@ -0,0 +1,121 @@
...
+  double max_yaw_rate_{ std::numeric_limits<double>::infinity() };
+  rclcpp::Duration cmd_timeout_{ 0, 0 };
+
+  struct CommandState
+  {
+    double vx{ 0.0 };
+    double vy{ 0.0 };
+    double wz{ 0.0 };
+    rclcpp::Time time{ 0, 0, RCL_ROS_TIME };
+    bool valid{ false };
```
eholum-nasa: I don't _think_ `valid` is necessary? Could remove it and just let the timeout check either catch it, but everything defaults to 0 anyway.

ndunkelb-nasa (commented):
Just a couple of comments to check some implementation details!

eholum-nasa (commented):
Definitely does what it promises! Side note that I appreciate all the warnings about breaking physics.

https://github.com/user-attachments/assets/aaeb2652-e050-4449-a58f-2472127bfeac

Just a minor nit and the philosophical design question, but otherwise I think this is in a good state.

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
saikishor: I agree with you. I'm not proud of this implementation. Moreover, if any plugin uses qvel for their use case, then they will receive NaNs instead of a valid qvel. This is something I dislike in this approach. I also agree with the limitations and how much this is causing the design to keep editing the primary system interface.

saikishor:
I'll respond to you guys tonight once I reach home. Thank you for the review

mujoco_ros2_control_plugins/src/base_velocity_plugin.cpp
scastro-nasa: Nav2 over the last few ROS distros has attempted to migrate from Twist to TwistStamped:
https://discourse.openrobotics.org/t/nav2-support-for-twiststamped-available-rolling-jazzy/35995

Note the post also mentions Gazebo and ros2_control, so ... probably stamped is the way to go since it's a superset of twist?

saikishor: Yes, it is better to go with TwistStamped. For humble versions, Nav2 was still with Twist IIRC. Having an option to toggle helps. Maybe I can change the default

eholum-nasa (approved):
Assuming CI looks good thanks for addressing my nits! I didn't see anything in the logs so assuming this was random action runner crashing out...

saikishor:
> Assuming CI looks good thanks for addressing my nits! I didn't see anything in the logs so assuming this was random action runner crashing out...

@eholum-nasa Should we merge this and then discuss about #262 later? or wait for #262 to modify things here?

eholum-nasa:
> @eholum-nasa Should we merge this and then discuss about https://github.com/ros-controls/mujoco_ros2_control/pull/262 later? or wait for https://github.com/ros-controls/mujoco_ros2_control/pull/262 to modify things here?

Either way is good with me. It might take me a few days to get to #262, so giving you the option if you want to merge this and change later...

saikishor:
> Either way is good with me. It might take me a few days to get to #262, so giving you the option if you want to merge this and change later...

OK I'll merge it and add the changes of this plugin also into that PR

PR #256, opened 2026-07-27
Support joint overrides in reset world service

### LINKED ISSUE #255: Set non-free joint states via service as well
Tagging @saikishor since I'd like your feedback before I work on something.

In https://github.com/ros-controls/mujoco_ros2_control/pull/233, we (you) added a service to set free joint states.

I was thinking of also adding capabilities to set joint states for single-dof joints, think like opening or closing a door mid-simulation, or resetting the arm to home pose.

Do you think maybe prior to a release we should modify the existing service so the payload has both a `FreeJointStateArray` and a `JointState` for these purposes? Again trying to make sure the user can set the entire state of the sim in one call if possible.

The alternative would be to include this as a separate service / as part of the reset world service, which seems less ideal to me.

scastro-nasa:
## Description

This PR extends the `~/reset_world` service to also apply joint (both single-dof hinge/slide and free joint) overrides.

This stacks on top of keyframes, you can effectively reset world to a known state and then apply extra changes like moving individual free objects or sliding joints like doors/drawers/etc.

A lot of the existing code used by #233 was reusable, so the bulk of the changes in this PR are shuffling things around to be maximally reusable in the world reset context. The diffs are at times a little tough to read because of this.

Tested on our internal NASA simulation and :+1: 

Fixes https://github.com/ros-controls/mujoco_ros2_control/issues/255

Yes, it adds a feature and changes a ROS service + message definition.

Yes, Claudio Fabula 5

... it went pretty extensive on the test coverage, so happy to trim down the tests which add a lot to this PR if desired. Basically half the PR is tests lol.

### DIFF
--- a/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
+++ b/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
@@ -45,6 +45,7 @@
 #include <mujoco_ros2_control_msgs/srv/set_pause.hpp>
 #include <mujoco_ros2_control_msgs/srv/step_simulation.hpp>
 #include <mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp>
+#include <sensor_msgs/msg/joint_state.hpp>
 
 namespace mujoco_ros2_control
 {
@@ -178,10 +179,17 @@ public:
   }
 
   /**
-   * @brief Reset simulation state (qpos/qvel/ctrl/sensors/forces) to the captured initial state.
+   * @brief Reset simulation state (qpos/qvel/ctrl/sensors/forces) to the captured initial state,
+   * optionally applying joint state overrides on top of the restored state.
+   *
+   * Overrides are assumed to have been validated by the caller, so they cannot fail here
+   * (see `resolve_joint_state_writes` / `resolve_free_joint_writes`).
+   *
    * @note Caller must hold the sim mutex.
    */
-  void reset_world_state(bool fill_initial_state);
+  void
+  reset_world_state(bool fill_initial_state, const sensor_msgs::msg::JointState& joint_state_overrides = {},
+                    const std::vector<mujoco_ros2_control_msgs::msg::FreeJointState>& free_joint_state_overrides = {});
 
   /**
    * @brief Sets the pose and velocity of one or more free-joint objects, identified by body name.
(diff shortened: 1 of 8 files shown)


mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
```diff
@@ -178,10 +179,17 @@ class MujocoSimulation
...
+   *
+   * Overrides are assumed to have been validated by the caller, so they cannot fail here
+   * (see `resolve_joint_state_writes` / `resolve_free_joint_writes`).
+   *
    * @note Caller must hold the sim mutex.
    */
-  void reset_world_state(bool fill_initial_state);
+  void
+  reset_world_state(bool fill_initial_state, const sensor_msgs::msg::JointState& joint_state_overrides = {},
+                    const std::vector<mujoco_ros2_control_msgs::msg::FreeJointState>& free_joint_state_overrides = {});
```
saikishor: Do we have to have the default arguments here?

mujoco_ros2_control_msgs/srv/ResetWorld.srv
```diff
@@ -1,4 +1,25 @@
...
+# `velocity` must each be empty or the same length as `name`; an empty array leaves that field
+# at its reset value. `effort` is not supported and must be empty. `header` is not used.
+sensor_msgs/JointState joint_state_overrides
+
+# Optional overrides for free joints, applied on top of the reset state and keyed by the name
+# of the body each free joint drives. See FreeJointState.msg for per-entry fields; any
+# `pose`/`twist` frame_id is resolved against body poses *after* the reset is applied, before
+# any overrides are written.
+FreeJointState[] free_joint_state_overrides
+
```
saikishor: Should we create a new message to be able to call it overrides and then inside we have joint_states and free_joint_states

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
scastro-nasa: there is one call of `reset_world_state(true)`, no extra args, in the code when the a UI reset is detected. We could just as well change that one call site, but I do like keeping the option around to not need to create any override messages. Do you have a strong inclination one way or the other?

mujoco_ros2_control_msgs/srv/ResetWorld.srv
scastro-nasa: I don't think I have a super strong opinion either way, I guess we could make a `mujoco_ros2_control_msgs/SimulationState` message or something?

It's more build time / possible confusion to users, but we do save on a little bit of typing and -- more importantly -- if there ever are more override fields it's easy to extend.

saikishor: Yup, that's the idea. I think it is clear that way no?

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
saikishor: I usually prefer to avoid default args as much as possible. My rule of thumb is if 80% of usecases cover the same parameters or same performance, then add a default arg, if not avoid it. Ofcourse, sometimes not to break API

saikishor: @scastro-nasa I still need to review the parts of code. I've a question, do we wait for this to get in? or we tag the package already?. I'm fine with any option

scastro-nasa: I would prefer we get this in before tagging so we don't have to immediately break ABI / message definitions right after. I can get this done today!

mujoco_ros2_control_msgs/srv/ResetWorld.srv
scastro-nasa: I don't really have a strong opinion, so could go either way for me. Will implement your idea!

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
saikishor: That's very true. Sure, we can wait. No rush. Take your time :)

mujoco_ros2_control_msgs/srv/ResetWorld.srv
saikishor: I'm flexible, I don't have a strong opinion here either, I thought of it as we have overrides in both naming. haha

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
scastro-nasa: Done, take a look?

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
```diff
@@ -360,18 +367,61 @@ class MujocoSimulation
...
+   */
+  struct SingleDofJointWrite
+  {
+    int qpos_adr{ -1 };
+    int qvel_adr{ -1 };
+    bool has_position{ false };
+    bool has_velocity{ false };
+    mjtNum position{ 0.0 };
+    mjtNum velocity{ 0.0 };
+  };
```
saikishor: I'm not sure if we need another struct to do the things around

saikishor (dismissed):
Looks good to me. 

Some parts can be refined. I've made some simplifications locally. I can open a different PR if needed.

scastro-nasa:
> Some parts can be refined. I've made some simplifications locally. I can open a different PR if needed.

If you have some changes, feel free to push them here, as long as I can take a look and give them the OK.

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
scastro-nasa: I agree, and was probably just going off the way it was done for free joints. If we can find a way to get around this and get rid of both the `FreeJointWrite` and `SingleDofJointWrite` structs, that would be even better!

saikishor:
> If you have some changes, feel free to push them here, as long as I can take a look and give them the OK.

Ofcourse, but they are quite some changes. We can merge it and then I can open it tomorrow sometime.

scastro-nasa:
I got inspired by your comment to remove the "write" structs, so pushing something in a bit. Let me know what you think there / whether that's similar to your changes? I'm in no rush to get this change in...

saikishor:
> I got inspired by your comment to remove the "write" structs, so pushing something in a bit. Let me know what you think there / whether that's similar to your changes? I'm in no rush to get this change in...

Suree. I'll take a look and try to commit my changes. 

My changes include the following:
* Doc strings
* Rename the variable in the messages from override to state_override and then free_joint_states to free_joints
* Refine the methods by adding some new + use std::optional instead of has_position and has_velocity
* Reset world state is split into restore reset state and finalize reset

saikishor (approved):
I like the changes now

scastro-nasa:
> I like the changes now

Thank you! Feel free to tag me on any subsequent ones!

PR #257, opened 2026-07-29
Magnetometer

christianrauch:
## Description

This adds support for the MuJoCo sensor [`magnetometer`](https://mujoco.readthedocs.io/en/stable/XMLreference.html#sensor-magnetometer).

Fixes # (issue)

Yes. It will expose new interfaces (`magnetic_field.x`, `magnetic_field.y`, `magnetic_field.z`) when `<magnetometer name="..." site="..."/>` is used in the MuJoCo model.

Yes. Codex.

### Additional Information

I did not add a [`magnetometer_broadcaster`](https://github.com/ros-controls/ros2_controllers/tree/master/magnetometer_broadcaster), as this has not been released yet to the bloom repo.

## TODOs

- [x] Fork the repository.

### DIFF
--- a/mujoco_ros2_control/CMakeLists.txt
+++ b/mujoco_ros2_control/CMakeLists.txt
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 3.16)
 
 project(mujoco_ros2_control)
 
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
 find_package(ros2_control_cmake REQUIRED)
 
 # find dependencies
--- a/mujoco_ros2_control/CMakeLists.txt
+++ b/mujoco_ros2_control/CMakeLists.txt
@@ -2,10 +2,6 @@ cmake_minimum_required(VERSION 3.16)
 
 project(mujoco_ros2_control)
 
-set(CMAKE_CXX_STANDARD 20)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_EXTENSIONS OFF)
-
 find_package(ros2_control_cmake REQUIRED)
 
 # find dependencies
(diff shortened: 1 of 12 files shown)


mujoco_ros2_control/CMakeLists.txt
```diff
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 3.16)
 
 project(mujoco_ros2_control)
 
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
```
saikishor: Any reason to set these?

If we were to set these, ros2_control_cmake should be handling this.

saikishor (commented):
Overall LGTM. Thanks for adding tests. Appreciate it

Just a minor nitpick

mujoco_ros2_control/CMakeLists.txt
christianrauch: > Any reason to set these?

[`contains`](https://en.cppreference.com/cpp/container/unordered_map/contains) requires C++20, and it has to be set explicitly for older Ubuntu versions (22.04 in this case).

> If we were to set these, ros2_control_cmake should be handling this.

Do we have to use the same C++ version as the other parts in the ROS control stack?

Is this to ensure ABI compatibility or something similar?

How do I set this? According to https://github.com/ros-controls/ros2_control_cmake/pull/28, this should already be set. I presume that this wasn't backported to older versions and distros.

saikishor: Can you replace it with count? https://en.cppreference.com/cpp/container/unordered_map/count

We do that to ensure to have same kind of settings over all the packages and for consistency reasons

saikishor (approved):

mujoco_ros2_control/CMakeLists.txt
saikishor: Thank you :)

PR #258, opened 2026-07-30
support merging multiple sets of state interfaces from multiple sensor tags

christianrauch:
## Description

At the moment, the `MujocoSystemInterface` does not support multiple `sensor` tags with the same name.

For example, the `sensor`:
```XML
    <sensor name="base_imu">
      <param name="mujoco_type">imu</param>
      <param name="mujoco_sensor_name">body</param>

      <state_interface name="orientation.x" />
      <state_interface name="orientation.y" />
      <state_interface name="orientation.z" />
      <state_interface name="orientation.w" />
      <state_interface name="angular_velocity.x" />
      <state_interface name="angular_velocity.y" />
      <state_interface name="angular_velocity.z" />
      <state_interface name="linear_acceleration.x" />
      <state_interface name="linear_acceleration.y" />
      <state_interface name="linear_acceleration.z" />
    </sensor>
```
will create interfaces in the `base_imu` namespace, e.g. `base_imu/angular_velocity.x`.

It is not possible to add a second set of state interfaces to the same namespace, and something like:
```XML
    <sensor name="base_imu">
      <param name="mujoco_type">magnetometer</param>
      <param name="mujoco_sensor_name">body_magnetometer</param>

      <state_interface name="magnetic_field.x"/>
      <state_interface name="magnetic_field.y"/>
      <state_interface name="magnetic_field.z"/>
    </sensor>
```
will be ignored.

This PR enables mapping multiple sensors to the same namespace, such that modular sensors can place state interfaces into the same namespace, in the same way this can be done with multiple `<ros2_control type="sensor">` for a modular real hardware system.

Fixes # (issue)

Yes. It allows adding multiple sets of state interfaces to the same namespace.

No.

## TODOs

- [x] Fork the repository.

### DIFF
--- a/mujoco_ros2_control/CMakeLists.txt
+++ b/mujoco_ros2_control/CMakeLists.txt
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 3.16)
 
 project(mujoco_ros2_control)
 
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
 find_package(ros2_control_cmake REQUIRED)
 
 # find dependencies
--- a/mujoco_ros2_control/CMakeLists.txt
+++ b/mujoco_ros2_control/CMakeLists.txt
@@ -2,10 +2,6 @@ cmake_minimum_required(VERSION 3.16)
 
 project(mujoco_ros2_control)
 
-set(CMAKE_CXX_STANDARD 20)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_EXTENSIONS OFF)
-
 find_package(ros2_control_cmake REQUIRED)
 
 # find dependencies
--- a/mujoco_ros2_control/src/mujoco_system_interface.cpp
+++ b/mujoco_ros2_control/src/mujoco_system_interface.cpp
@@ -526,7 +526,7 @@ std::vector<hardware_interface::StateInterface> MujocoSystemInterface::export_st
   // Add state interfaces for fts sensors
   for (auto& sensor : ft_sensor_data_)
   {
-    for (const hardware_interface::ComponentInfo& ci : sensors_hw_info_[sensor.name])
+    for (const auto& ci : sensors_hw_info_[sensor.name])
     {
       for (const auto& state_if : ci.state_interfaces)
       {
@@ -561,7 +561,7 @@ std::vector<hardware_interface::StateInterface> MujocoSystemInterface::export_st
   // Add state interfaces for IMU sensors
   for (auto& sensor : imu_sensor_data_)
   {
-    for (const hardware_interface::ComponentInfo& ci : sensors_hw_info_[sensor.name])
+    for (const auto& ci : sensors_hw_info_[sensor.name])
     {
       for (const auto& state_if : ci.state_interfaces)
       {
@@ -641,7 +641,7 @@ std::vector<hardware_interface::StateInterface> MujocoSystemInterface::export_st
   // Add state interfaces for pose sensors
   for (auto& sensor : pose_sensor_data_)
   {
-    for (const hardware_interface::ComponentInfo& ci : sensors_hw_info_[sensor.name])
+    for (const auto& ci : sensors_hw_info_[sensor.name])
     {
       for (const auto& state_if : ci.state_interfaces)
       {
--- a/mujoco_ros2_control/src/mujoco_system_interface.cpp
+++ b/mujoco_ros2_control/src/mujoco_system_interface.cpp
@@ -526,9 +526,9 @@ std::vector<hardware_interface::StateInterface> MujocoSystemInterface::export_st
   // Add state interfaces for fts sensors
   for (auto& sensor : ft_sensor_data_)
   {
-    if (auto it = sensors_hw_info_.find(sensor.name); it != sensors_hw_info_.end())
+    for (const hardware_interface::ComponentInfo& ci : sensors_hw_info_[sensor.name])
     {
-      for (const auto& state_if : it->second.state_interfaces)
+      for (const auto& state_if : ci.state_interfaces)
       {
         if (state_if.name == "force.x")
         {
@@ -561,9 +561,9 @@ std::vector<hardware_interface::StateInterface> MujocoSystemInterface::export_st
   // Add state interfaces for IMU sensors
   for (auto& sensor : imu_sensor_data_)
   {
-    if (auto it = sensors_hw_info_.find(sensor.name); it != sensors_hw_info_.end())
+    for (const hardware_interface::ComponentInfo& ci : sensors_hw_info_[sensor.name])
     {
-      for (const auto& state_if : it->second.state_interfaces)
+      for (const auto& state_if : ci.state_interfaces)
       {
         if (state_if.name == "orientation.x")
         {
@@ -641,9 +641,9 @@ std::vector<hardware_interface::StateInterface> MujocoSystemInterface::export_st
   // Add state interfaces for pose sensors
   for (auto& sensor : pose_sensor_data_)
   {
-    if (auto it = sensors_hw_info_.find(sensor.name); it != sensors_hw_info_.end())
+    for (const hardware_interface::ComponentInfo& ci : sensors_hw_info_[sensor.name])
     {
-      for (const auto& state_if : it->second.state_interfaces)
+      for (const auto& state_if : ci.state_interfaces)
       {
         if (state_if.name == "position.x")
         {
@@ -1871,7 +1871,7 @@ void MujocoSystemInterface::register_sensors(const hardware_interface::HardwareI
                 sensor_name.c_str(), mujoco_type.c_str(), mujoco_sensor_name.c_str());
 
     // Add to the sensor hw information map
-    sensors_hw_info_.insert(std::make_pair(sensor_name, sensor));
+    sensors_hw_info_[sensor_name].push_back(sensor);
 
     if (mujoco_type == "fts")
     {
--- a/mujoco_ros2_control/src/mujoco_system_interface.cpp
+++ b/mujoco_ros2_control/src/mujoco_system_interface.cpp
@@ -680,9 +680,9 @@ std::vector<hardware_interface::StateInterface> MujocoSystemInterface::export_st
   // Add state interfaces for magnetometer sensors
   for (auto& sensor : magnetometer_sensor_data_)
   {
-    if (sensors_hw_info_.count(sensor.name) > 0)
+    for (const auto& ci : sensors_hw_info_[sensor.name])
     {
-      for (const auto& state_if : sensors_hw_info_.at(sensor.name).state_interfaces)
+      for (const auto& state_if : ci.state_interfaces)
       {
         if (state_if.name == "magnetic_field.x")
         {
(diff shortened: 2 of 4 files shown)


mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -526,9 +526,9 @@ std::vector<hardware_interface::StateInterface> MujocoSystemInterface::export_st
   // Add state interfaces for fts sensors
   for (auto& sensor : ft_sensor_data_)
   {
-    if (auto it = sensors_hw_info_.find(sensor.name); it != sensors_hw_info_.end())
+    for (const hardware_interface::ComponentInfo& ci : sensors_hw_info_[sensor.name])
```
saikishor:
```suggestion
    for (const auto& ci : sensors_hw_info_[sensor.name])
```

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -561,9 +561,9 @@ std::vector<hardware_interface::StateInterface> MujocoSystemInterface::export_st
   // Add state interfaces for IMU sensors
   for (auto& sensor : imu_sensor_data_)
   {
-    if (auto it = sensors_hw_info_.find(sensor.name); it != sensors_hw_info_.end())
+    for (const hardware_interface::ComponentInfo& ci : sensors_hw_info_[sensor.name])
```
saikishor:
```suggestion
    for (const auto& ci : sensors_hw_info_[sensor.name])
```

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: Maybe we should also use the find here, because if not if the keydoesn't exit, we'll have a problem

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -641,9 +641,9 @@ std::vector<hardware_interface::StateInterface> MujocoSystemInterface::export_st
   // Add state interfaces for pose sensors
   for (auto& sensor : pose_sensor_data_)
   {
-    if (auto it = sensors_hw_info_.find(sensor.name); it != sensors_hw_info_.end())
+    for (const hardware_interface::ComponentInfo& ci : sensors_hw_info_[sensor.name])
     {
-      for (const auto& state_if : it->second.state_interfaces)
+      for (const auto& state_if : ci.state_interfaces)
```
saikishor, line 658: Same here

saikishor (changes requested):
The idea looks good, just some minor comments

mujoco_ros2_control/src/mujoco_system_interface.cpp
christianrauch: If I understand the semantics of the `operator[]` correctly, it will construct a value when accessing via the key. Since the default constructor is an empty list, the loop will not iterate when the key does not exist in the `unordered_map` previously. Hence, it will iterate over an empty list on the fly, and we do not have to check this before.

saikishor: It doesn't work that way, you are not initializing it, you are accessing the element

christianrauch: See https://en.cppreference.com/cpp/container/unordered_map/operator_at:
> Inserts a [value_type](https://en.cppreference.com/cpp/container/unordered_map#Member_types) object constructed in-place from std::piecewise_construct, std::forward_as_tuple(key), std::tuple<>() if the key does not exist.

Something like:
```C++
#include <iostream>
#include <string>
#include <vector>
#include <unordered_map>

int main()
{
  std::unordered_map<std::string, std::vector<int>> a;
  std::cout << "len of value at 'abc': " << a["abc"].size() << std::endl;
}
```
Gives you:
```
len of value at 'abc': 0
```
because the `std::vector<int>` is constructed in-place just like the documentation claims.

saikishor: Ok, let me recheck it. 

Thanks

saikishor: Thanks for the example, you are right

christianrauch: Done.

mujoco_ros2_control/CMakeLists.txt
```diff
@@ -2,6 +2,10 @@ cmake_minimum_required(VERSION 3.16)
 
 project(mujoco_ros2_control)
 
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
```
saikishor:
```suggestion
```

saikishor (changes requested):
@christianrauch I've a curious question: let's say you set the same sensor information for the Pose and IMU; both have the orientation interfaces, right?. How are you planning to deal with such scenarios?

christianrauch:
> @christianrauch I've a curious question: let's say you set the same sensor information for the Pose and IMU; both have the orientation interfaces, right?. How are you planning to deal with such scenarios?

That's a good question, and I am not sure how the controller manager will deal with this. But what I am suggesting here can be done in ros2_control already, using different sensor plugins.

saikishor:
> That's a good question, and I am not sure how the controller manager will deal with this. But what I am suggesting here can be done in ros2_control already, using different sensor plugins.

Your use case makes total sense, and yes, it should be able to run with different sensor hardware components. I'll rethink about this to see if there are any other corner cases and will let you know :)

christianrauch:
I did a quick check with two `mock_components/GenericSystem` claiming the same interfaces (`joint1/position` and `joint1/velocity`):
```XML
<?xml version="1.0"?>
<robot xmlns:xacro="http://www.ros.org/wiki/xacro" name="two_mock_robot">
  <link name="world"/>
  <link name="base_link"><visual><geometry><box size="0.2 0.2 0.1"/></geometry></visual></link>
  <link name="link1"><visual><geometry><cylinder radius="0.05" length="0.5"/></geometry><origin xyz="0 0 0.25"/></visual></link>
  <link name="link2"><visual><geometry><cylinder radius="0.04" length="0.4"/></geometry><origin xyz="0 0 0.2"/></visual></link>

  <joint name="world_to_base" type="fixed"><parent link="world"/><child link="base_link"/></joint>
  <joint name="joint1" type="revolute"><parent link="base_link"/><child link="link1"/><origin xyz="0 0 0.05"/><axis xyz="0 0 1"/><limit lower="-3.14" upper="3.14" effort="100" velocity="1.0"/></joint>
  <joint name="joint2" type="revolute"><parent link="link1"/><child link="link2"/><origin xyz="0 0 0.5"/><axis xyz="0 1 0"/><limit lower="-1.57" upper="1.57" effort="100" velocity="1.0"/></joint>

  <ros2_control name="MockHardware1" type="system">
    <hardware><plugin>mock_components/GenericSystem</plugin></hardware>
    <joint name="joint1">
      <command_interface name="position"/>
      <state_interface name="position"><param name="initial_value">0.0</param></state_interface>
      <state_interface name="velocity"/>
    </joint>
  </ros2_control>

  <ros2_control name="MockHardware2" type="system">
    <hardware><plugin>mock_components/GenericSystem</plugin></hardware>
    <joint name="joint1">
      <command_interface name="position"/>
      <state_interface name="position"><param name="initial_value">0.0</param></state_interface>
      <state_interface name="velocity"/>
    </joint>
  </ros2_control>
</robot>
```

This results in:
```
[INFO] [controller_manager]: Loading hardware 'MockHardware1' 
[INFO] [controller_manager]: Loaded hardware 'MockHardware1' from plugin 'mock_components/GenericSystem'
[INFO] [controller_manager]: Initialize hardware 'MockHardware1' 
[INFO] [controller_manager]: Successful initialization of hardware 'MockHardware1'
[INFO] [controller_manager]: Loading hardware 'MockHardware2' 
[INFO] [controller_manager]: Loaded hardware 'MockHardware2' from plugin 'mock_components/GenericSystem'
[INFO] [controller_manager]: Initialize hardware 'MockHardware2' 
[INFO] [controller_manager]: Successful initialization of hardware 'MockHardware2'
[WARN] [controller_manager]: Exception occurred while importing state interfaces: ResourceStorage: Tried to insert StateInterface with already existing key. Insert[joint1/velocity]
[WARN] [controller_manager]: Exception occurred while importing state interfaces: ResourceStorage: Tried to insert StateInterface with already existing key. Insert[joint1/position]
[WARN] [controller_manager]: Importing state interfaces for the hardware 'MockHardware2' returned no state interfaces.
[ERROR] [controller_manager]: Exception of type : St13runtime_error occurred while importing command interfaces for the hardware 'MockHardware2' : ResourceStorage: Tried to insert CommandInterface with already existing key. Insert[joint1/position]
[INFO] [controller_manager]: Activating component 'MockHardware2'.
[INFO] [resource_manager]: 'configure' hardware 'MockHardware2' 
[INFO] [resource_manager]: Successful 'configure' of hardware 'MockHardware2'
[INFO] [resource_manager]: 'activate' hardware 'MockHardware2' 
[INFO] [resource_manager]: Successful 'activate' of hardware 'MockHardware2'
[INFO] [controller_manager]: Activating component 'MockHardware1'.
[INFO] [resource_manager]: 'configure' hardware 'MockHardware1' 
[INFO] [resource_manager]: Successful 'configure' of hardware 'MockHardware1'
[INFO] [resource_manager]: 'activate' hardware 'MockHardware1' 
[INFO] [resource_manager]: Successful 'activate' of hardware 'MockHardware1'
```

Both components are activated, but interfaces are assigned "first come, first serve":
```
$ ros2 control list_hardware_components -v
Hardware Component 1
	name: MockHardware2
	type: system
	plugin name: mock_components/GenericSystem
	state: id=3 label=active
	read/write rate: 100 Hz
	is_async: False
	command interfaces
	state interfaces
Hardware Component 2
	name: MockHardware1
	type: system
	plugin name: mock_components/GenericSystem
	state: id=3 label=active
	read/write rate: 100 Hz
	is_async: False
	command interfaces
		joint1/position [double] [available] [unclaimed]
	state interfaces
		joint1/velocity [double] [available]
		joint1/position [double] [available]
```

I guess it would be better to properly fail here, but that is the responsibility of the resource manager.

saikishor:
@christianrauch are you fine with the following changes?

christianrauch:
Yes, I am fine with the changes.

I am just a bit confused about why some of the commits are duplicated now. E.g. the "revert C++ 20" commit from the magnetometer PR (adfd887f2b6912383ca9d524edbcf50ee9d8beb2) is already on the main branch. Seeing the same change here (5814599e91cd0b44a73fba5663f88a628ae3330f) is confusing and makes it more difficult to see the actual commit changes that are applied to the main branch. I.e. it is not clear to me why these changes have to be worked on multiple times. Is this related to the merging vs. rebasing to update the PR?

saikishor:
> Yes, I am fine with the changes.
> 
> I am just a bit confused about why some of the commits are duplicated now. E.g. the "revert C++ 20" commit from the magnetometer PR (adfd887f2b6912383ca9d524edbcf50ee9d8beb2) is already on the main branch. Seeing the same change here (5814599e91cd0b44a73fba5663f88a628ae3330f) is confusing and makes it more difficult to see the actual commit changes that are applied to the main branch. I.e. it is not clear to me why these changes have to be worked on multiple times. Is this related to the merging vs. rebasing to update the PR?

Yes I believe so, anyway it is not going to be an issue as we squash commits. If you can confirm the changes alone, it would be great

saikishor (approved):

christianrauch:
> Yes I believe so, anyway it is not going to be an issue as we squash commits. If you can confirm the changes alone, it would be great

Well, yes, I am fine with your changes.

There is no way to "officially" approve the changes in the PR since I cannot approve my own PRs.

saikishor:
Your word is fine ;)

PR #262, opened 2026-08-05
Simplify mujoco_ros2_control_plugins integration

saikishor:
## Description

Inspired from the discussions started in https://github.com/ros-controls/mujoco_ros2_control/pull/254#discussion_r3716344976

A bit yes, what mjData is parsed to update of the plugin

Idea is written down, but mostly brainstormed andd implemented with Claude

### Additional Information
I think this approach is interesting as it doesn't need a lot of handshakes of data and this should be pretty easy to expand for plugins, and I moved the plugins update cycle to the physics loop, so we don't need different datas

## TODOs

- [ ] Fork the repository.

### DIFF
--- a/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
+++ b/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
@@ -297,6 +297,16 @@ public:
   void register_plugins(std::vector<std::shared_ptr<mujoco_ros2_control_plugins::MuJoCoROS2ControlPluginBase>> plugins);
 
   /**
+   * @brief Stops driving the registered plugins.
+   *
+   * Once this returns, no plugin `update()` is in flight and none will be started, so the owner can
+   * safely tear the plugins down. Plugins cache the live model and data pointers, so the required
+   * teardown order is: `detach_plugins()`, then each plugin's `cleanup()`, then destroying this
+   * simulation. Idempotent.
+   */
+  void detach_plugins();
+
+  /**
    * @brief Accessor for the mutex which locks access to the data and model.
    */
   std::recursive_mutex& mutex() const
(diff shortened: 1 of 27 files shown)


saikishor:
@scastro-nasa I added you here, in case it interests you

mujoco_ros2_control/include/mujoco_ros2_control/mujoco_simulation.hpp
```diff
@@ -301,28 +327,50 @@ class MujocoSimulation
...
    */
   void apply_staged_control_inputs();
 
+  /// One commandable field: where plugins write it, where it lands, and how many entries it has.
+  struct CommandableField
+  {
+    mjtNum* commanded;
+    mjtNum* live;
+    int count;
+  };
```
scastro-nasa, line 344: might be worth adding brief comments on what these fields mean

scastro-nasa (commented):
Didn't give it a thorough look yet (will let @eholum-nasa do that probably), but certainly the hw interface code was greatly simplified, so that's good! I also like that you put some of the plugin initialization stuff in a separate method.

The one thing we should probably look at is a before/after plotjuggler of the hw interface statistics. While I don't think this is a revert of the behavior before I added the double-buffer to fix the deadlocks (https://github.com/ros-controls/mujoco_ros2_control/pull/224), we should make sure we're not regressing on performance with this change. Or maybe it even got better!

eholum-nasa (commented):
Yeah changing ownership rules really will help simplify some of the timing issues `MujocoSimulation`! That said, I have some design concerns about the significance of the change to the contract in the PluginInterface, and with the plugin ownership moving to the `MujocoSimulation`. In summary I think we should:

* Keep plugin management purely in the hardware interface
  * I’d really like `MujocoSimulation` to be the wrapper around the physics loop as much as possible, and putting deep hooks into it (like for the `q_vel` override) make the overall system less flexible.
* Keep the contract of `update` call in the plugin interface tied to the ros2_control update loop
  * I could _probably_ be convinced we could change this, but it's a pretty big API change for the plugins. I like the option to act on a non-physics sim locked copy of the data easily.
  * It’s easier for some things to have this tied to the control loop rate, rather than the physics loop. Also the physics loop is scary and easy to mess up.

That said I _do_ think we can have our cake and eat it too. Fundamentally my question is how much access do we want to provide to users through plugins? IMO we’re at the point that we should just give plugin interfaces access to the physic loop’s `mjData` immediately before stepping and let it add what it needs.

I think adding something functional like the reset world callback would make sense here for the MujocoSimulation class:

```c
void set_pre_step_callback(std::function<void(mjData*)> callback);
```

Then we would basically keep the same logic in the PhyiscsLoop of applying control functions, step callbacks, and stepping:

```c
apply_staged_control_inputs();
pre_step_callback_( mj_data_); } // That callback could literally do whatever, it's god mode
mj_step(mj_model_, mj_data_);
```

And then adding something like this to the plugin interface, with the caveat that it could explode the sim:

```c
virtual void pre_step(mjModel* /*model*/, mjData* /*data*/) {}
```

Most plugins would not need to change, since only those that actually need to modify physics data would need to use this (like the velocity override and force applicator). Moreover, we’d still be able to remove logic needed to plumb through control inputs from plugins.

Does that make sense? I have a version this I can clean up if you want to see what it looks like.

eholum-nasa:
Added here for consideration: https://github.com/ros-controls/mujoco_ros2_control/pull/282

PR #263, opened 2026-08-05
Refine reset world overrides functionality

saikishor:
## Description
Related to refinements of #256 

Fixes # (issue)

A little, but as it is not released, shouldn't be a problem

To refine things, yes

### DIFF
--- a/mujoco_ros2_control_msgs/msg/SimulationState.msg
+++ b/mujoco_ros2_control_msgs/msg/SimulationState.msg
@@ -1,7 +1,13 @@
-# Expresses a complete MuJoCo simulation state.
+# A sparse set of MuJoCo joint states: only the joints and bodies listed here are specified.
+# This is not a complete simulation state -- it carries no `ctrl`, `act`, `time`, or sensor data.
 
-# List of single-DOF joint states (e.g., hinges/slides).
+# Single-DOF (hinge/slide) joint states, keyed by MuJoCo joint name and given as parallel
+# `name`/`position`/`velocity` arrays. `position` and `velocity` must each be empty or the same
+# length as `name`, so a given field is set for every listed joint or for none of them.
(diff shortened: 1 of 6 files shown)


scastro-nasa (dismissed):
Thanks! These cleanups look good.

mujoco_ros2_control_msgs/msg/SimulationState.msg
```diff
@@ -1,7 +1,13 @@
-# Expresses a complete MuJoCo simulation state.
+# A sparse set of MuJoCo joint states: only the joints and bodies listed here are specified.
+# This is not a complete simulation state -- it carries no `ctrl`, `act`, `time`, or sensor data.
 
-# List of single-DOF joint states (e.g., hinges/slides).
+# Single-DOF (hinge/slide) joint states, keyed by MuJoCo joint name and given as parallel
+# `name`/`position`/`velocity` arrays. `position` and `velocity` must each be empty or the same
+# length as `name`, so a given field is set for every listed joint or for none of them.
+# `effort` is not supported and must be empty; `header` is unused.
```
scastro-nasa: I'm not 100% about this line because these restrictions pertain specifically to the use of the message inside the `/reset_world` service. Technically the message definition itself can do what it wants in other uses, but in the _service_ these extra things may be worth nothing.

saikishor: Agreed, You are right as always ;)

scastro-nasa (approved):

mujoco_ros2_control_msgs/msg/SimulationState.msg
scastro-nasa: if this were true, my life would be far easier lol

saikishor: HAHAHAHHAHA

PR #264, opened 2026-08-06
Use collision-tag meshes/info for collision geoms (primitives, STL/DAE/OBJ)

saikishor:
## Description

  First of a series of PRs splitting up #209 into smaller, reviewable pieces (see that PR for the original context, motivation, and before/after screenshots). Fixes some of the concerns listed in #120.

  This PR makes the URDF→MJCF converter treat `<visual>` and `<collision>` tags independently instead of always deriving collision geometry from whatever mesh was used for rendering:

  - Mesh filepaths are now read from **both** `<visual>` and `<collision>` tags (previously only one source was used), and converted to `.obj` as needed.
  - If a link defines a `<collision>` tag, its mesh/geometry is what drives physics — no longer just a byproduct of the visual mesh.
  - If a link has no `<collision>` tag, a collision is synthesized from its `<visual>` (matching the previous behaviour), so nothing regresses for URDFs that don't define collision geometry.
  - This synthesis happens at the URDF level (before MuJoCo fuses fixed-jointed bodies), so it stays correct per-link even after body fusion.

  Yes — links that define separate collision geometry in their URDF will now get that geometry in the generated MJCF instead of their visual mesh.

### DIFF
--- a/mujoco_ros2_control/docs/tools.rst
+++ b/mujoco_ros2_control/docs/tools.rst
@@ -128,11 +128,20 @@ Rough outline of the automated conversion process
      Duplicate mesh files will have an ``_N`` appended to them to avoid conflicts in the output ``assets`` folder.
      For instance, if running multiple types of UR robots in one sim, there will be multiple ``shoulder.dae`` files.
 
-- Reads absolute filepaths of all meshes and converts either ``.dae`` or ``.stl`` to ``.obj`` using trimesh.
+- Reads absolute filepaths of all meshes (from both ``<visual>`` and ``<collision>`` tags) and
+  converts either ``.dae`` or ``.stl`` to ``.obj`` using trimesh.
 
   - Puts all meshes into an ``assets/`` folder under ``mjcf_data/`` relative to current working dir.
   - Modifies filepaths again in URDF to point to the ``assets/`` folder.
   - Decomposes large meshes into multiple components to ensure convex hulls.
+
+- Handles visual and collision geometry independently:
+
+  - The ``<visual>`` geometry is used purely for rendering (the ``visual`` class).
+  - The ``<collision>`` geometry drives physics (the ``collision`` class). When a link defines no
+    ``<collision>``, a collision is synthesized from its ``<visual>`` so the visual mesh is reused
+    as the collision shape (the previous behaviour). This synthesis happens at the URDF level, so it
+    stays correct per link even when MuJoCo fuses fixed-jointed bodies together.
 
 - Publishes the new formatted robot description XML file that can be used for conversion.
 - Converts the new robot description URDF file.
--- a/mujoco_ros2_control/mujoco_ros2_control/__init__.py
+++ b/mujoco_ros2_control/mujoco_ros2_control/__init__.py
@@ -15,6 +15,7 @@
 from .urdf_to_mujoco_utils import (
     add_mujoco_info,
     remove_tag,
+    add_missing_collisions,
     extract_mesh_info,
     replace_package_names,
     get_images_from_dae,
@@ -47,6 +48,7 @@ from .urdf_to_mujoco_utils import (
 __all__ = [
     "add_mujoco_info",
     "remove_tag",
+    "add_missing_collisions",
     "extract_mesh_info",
     "replace_package_names",
     "get_images_from_dae",
--- a/mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
+++ b/mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
@@ -538,18 +538,27 @@ def update_obj_assets(dom, output_filepath, mesh_info_dict):
 
 def update_non_obj_assets(dom, output_filepath):
     """
-    We want to take the group 1 objects that get created, and turn them into the equivalent
-    but both in group 2 and in group 3. That means taking something like this
+    Classifies the geoms that MuJoCo imported from the URDF into the "visual" and
+    "collision" default classes. Because the source URDF now always carries both a
+    <visual> and a <collision> per renderable link (the latter synthesized from the
+    visual when missing, see add_missing_collisions), MuJoCo emits a separate geom for
+    each, and we simply tag them rather than duplicating a single geom.
+
+    MuJoCo imports a URDF <visual> as a geom that still carries its raw import
+    attributes, e.g.
         <geom type="mesh" contype="0" conaffinity="0" group="1" density="0" rgba="0.2 0.2 0.2 1" mesh="finger_v6"/>
-    and turning it into this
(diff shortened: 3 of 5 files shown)


ndunkelb-nasa:
@saikishor to do some initial testing, I was trying to see if I could get this to run in [our primary workspace](https://github.com/NASA-JSC-Robotics/clr_ws), but I get some pretty quick errors when I try to run any urdf -> mjcf conversion there. I wonder if we need some more tests to handle cases with different types of meshes based on what we find out here.

I am trying to look into what the issue is, but if you are curious as well, I am just opening up our workspace on jazzy-devel, checking out this PR for mujoco_ros2_control and running the following. 
```sh
ros2 launch clr_mujoco_config generate_clr_mjcf.launch.py
```

In the decomposition section, getting this error
```
[make_mjcf_from_robot_description.py-1] Traceback (most recent call last):
[make_mjcf_from_robot_description.py-1]   File "/home/er4-user/ws/install/mujoco_ros2_control/lib/mujoco_ros2_control/make_mjcf_from_robot_description.py", line 531, in <module>
[make_mjcf_from_robot_description.py-1]     main()
[make_mjcf_from_robot_description.py-1]   File "/home/er4-user/ws/install/mujoco_ros2_control/lib/mujoco_ros2_control/make_mjcf_from_robot_description.py", line 508, in main
[make_mjcf_from_robot_description.py-1]     fix_mujoco_description(
[make_mjcf_from_robot_description.py-1]   File "/home/er4-user/ws/install/mujoco_ros2_control/lib/mujoco_ros2_control/make_mjcf_from_robot_description.py", line 323, in fix_mujoco_description
[make_mjcf_from_robot_description.py-1]     dom = mrc.update_obj_assets(dom, output_filepath, mesh_info_dict)
[make_mjcf_from_robot_description.py-1]           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[make_mjcf_from_robot_description.py-1]   File "/home/er4-user/ws/install/mujoco_ros2_control/lib/python3.12/site-packages/mujoco_ros2_control/urdf_to_mujoco_utils.py", line 465, in update_obj_assets
[make_mjcf_from_robot_description.py-1]     sub_dom = minidom.parse(mesh_path)
[make_mjcf_from_robot_description.py-1]               ^^^^^^^^^^^^^^^^^^^^^^^^
[make_mjcf_from_robot_description.py-1]   File "/usr/lib/python3.12/xml/dom/minidom.py", line 1985, in parse
[make_mjcf_from_robot_description.py-1]     return expatbuilder.parse(file)
[make_mjcf_from_robot_description.py-1]            ^^^^^^^^^^^^^^^^^^^^^^^^
[make_mjcf_from_robot_description.py-1]   File "/usr/lib/python3.12/xml/dom/expatbuilder.py", line 907, in parse
[make_mjcf_from_robot_description.py-1]     with open(file, 'rb') as fp:
[make_mjcf_from_robot_description.py-1]          ^^^^^^^^^^^^^^^^
[make_mjcf_from_robot_description.py-1] FileNotFoundError: [Errno 2] No such file or directory: 'mjcf_data/assets/decomposed/finger_v6__1/finger_v6__1/finger_v6__1.xml'
[ERROR] [make_mjcf_from_robot_description.py-1]: process has died [pid 22869, exit code 1, cmd '/home/er4-user/ws/install/mujoco_ros2_control/lib/mujoco_ros2_control/make_mjcf_from_robot_description.py --urdf /tmp/tmpmulds1d3.urdf --convert_stl_to_obj --save_only --ros-args'].
```

Whereas on main, it doesn't try to decompose an `__1` version.

saikishor:
@ndunkelb-nasa can you test it with the last PR #269? Yo see if it happens?. If not, I can take a look and add those cases or verify why it is not working

saikishor:
@ndunkelb-nasa I tried to test it, but I don't have access to `clr_mujoco_config` package

ndunkelb-nasa:
Yeah, that runs all the way through, but still has a bunch of issues after running. In this video, you can see the result of running the generation from main on the left, and the result from running the generation on #269 on the right, then running the output in base mujoco simulate (these both start paused, then I unpause each). Obviously I know that with these changes, I will need to change how the contact disabling and stuff works, but there are a bunch of other problems that you can see in the videos, like `dae`s/`obj`s not keeping their visual meshes throughout the process. 

[new_generation.webm](https://github.com/user-attachments/assets/593e94e4-99cb-497d-8411-8065cef667b5)

Also something seems to have been super crazy with the scaling of this as you can see lol (also starts paused then I unpause).

[collision_scaling.webm](https://github.com/user-attachments/assets/203bf834-73c9-4e90-87d6-c5e35e109c33)

There seems to be quite a few things going on here, which is why I was hoping to test things in smaller chunks and verify functionality, rather than trying to fix 4-5 intertwined things from the final PR.

> @ndunkelb-nasa I tried to test it, but I don't have access to clr_mujoco_config package

Hmm, you should definitely be able to run, the package is [here](https://github.com/NASA-JSC-Robotics/chonkur_l_raile/tree/jazzy-devel/clr_mujoco_config), and gets pulled into the ws with submodules from the instructions on the front page of [clr_ws](https://github.com/NASA-JSC-Robotics/clr_ws), but don't want to necessarily force you to set all that up...Maybe the solution is to make a minimal example that breaks for this new  implementation? Unfortunately there seem to be a bunch of issues.

saikishor:
@ndunkelb-nasa sure, thanks for the info. I'll take a look and get back to you

scastro-nasa:
FYI the collision issue with the massive mesh seems to be on our side: https://github.com/NASA-JSC-Robotics/surface_robotics_mockups_description/pull/21

Also @saikishor I'm starting to look at this set of PRs, so I might be putting up some PRs to your branches for your review.

(EDIT: and I may have accidentally pushed to your branch first, but I force-pushed it back)

saikishor:
> FYI the collision issue with the massive mesh seems to be on our side: https://github.com/NASA-JSC-Robotics/surface_robotics_mockups_description/pull/21
> 
> Also @saikishor I'm starting to look at this set of PRs, so I might be putting up some PRs to your branches for your review.
> 
> (EDIT: and I may have accidentally pushed to your branch first, but I force-pushed it back)

No worries. Thanks a lot for taking a look at this.

scastro-nasa (approved):
I've merged my change into here and this branch works standalone now with our setup.

NOTE, however, that the default option remains to use the visual meshes, and I added a `--use_collision_tags` argument that makes it behave like the original version of this PR. I think this will be preferable as it doesn't break user workflows by default, but rather gives them the option.

Will now move up the stack and review the remaining PRs.

scastro-nasa:
@saikishor I think we may want to start burning down this stack and merging some of the PRs. This one to me is ready to go now.

saikishor:
> @saikishor I think we may want to start burning down this stack and merging some of the PRs. This one to me is ready to go now.

Sure, I can merge this

PR #265, opened 2026-08-06
visual-only mesh fallback + visual/collision group split, refactors make_mjcf_from_robot_description.py to reuse the new logic

saikishor:
## Description

  Second in the series splitting up #209 (see that PR for full context).

  Builds on top of #264  by giving visual and collision geoms their own distinct roles instead of sharing one code path:

  - **Visual** geometry is now purely for rendering (`visual` class). Visual meshes are kept as plain references — `.stl`/`.obj` untouched, only `.dae` converted (MuJoCo can't load DAE) — and are placed
  under `assets/visual/`. They are **never decomposed**, and the URDF material colour is carried over as the geom's `rgba`.
  - **Collision** geometry drives physics (`collision` class). Collision primitives (box/sphere/cylinder/capsule) are used as-is; collision meshes are used as a single whole-mesh collision unless the link
  is explicitly listed in a `decompose_mesh` input, in which case they're convex-decomposed with obj2mjcf under `assets/decomposed/`. Decomposition is therefore opt-in per link rather than automatic.
  - A mesh shared by a visual and a collision tag is converted once and reused for both.
  - `make_mjcf_from_robot_description.py` was refactored to reuse this shared logic instead of duplicating it.


scastro-nasa (commented):
ok it is this branch that then causes some of our visual meshes to disappear... digging into it now.

EDIT: Also, the decomposed meshes do not show up at all in this branch?

Image

scastro-nasa:
@saikishor I put my changes up in https://github.com/ros-controls/mujoco_ros2_control/pull/298

Just to check though, you do have upstack changes that prevent the visual and collision meshes from both being processed redundantly, right? That's #268 ?

saikishor:
> @saikishor I put my changes up in #298
> 
> Just to check though, you do have upstack changes that prevent the visual and collision meshes from both being processed redundantly, right? That's #268 ?

Yess exactly

scastro-nasa (dismissed):

scastro-nasa:
@saikishor let me know if this merge works for your use case so far and then we'll take it from there. When I next sit down with this stack I will keep going with your 3rd PR.

scastro-nasa (approved):

saikishor:
> @saikishor let me know if this merge works for your use case so far and then we'll take it from there. When I next sit down with this stack I will keep going with your 3rd PR.

I think it works for me, I'll recheck in the morning, but I'll merge this, so it is easier for you to go ahead with next one

scastro-nasa:
Cool thanks! I just tested `main` after this merge and so far it looks good on our application with the `--use_collision_tags` toggle.

PR #267, opened 2026-08-06
sets contype/conaffinity=0 on visual-only geoms

saikishor:
## Description

  Third in the series splitting up #209 (see that PR for full context).

  Explicitly marks visual-only geoms as non-colliding and gives visual/collision geoms distinguishing attributes:

  - Visual geoms: `class="visual"`, `contype="0" conaffinity="0" group="2" density="0"` — render-only, guaranteed to never participate in contacts, regardless of any MuJoCo default class inheritance.
  - Collision geoms: `class="collision"`, `group="3" contype="1" conaffinity="1"`. Whole-mesh/primitive collisions are tinted `material="bright_orange"` so they're easy to spot in the viewer by toggling
  group 3 visibility; the material is auto-added to `<asset>` if not already user-defined.
  - Decomposed collision pieces keep obj2mjcf's own per-piece colours instead of being recoloured, so individual convex hulls stay visually distinguishable.

  Yes — visual geoms are now guaranteed not to collide (previously this depended on default class inheritance), and collision geoms get a visible highlight colour in the MuJoCo viewer.


mujoco_ros2_control/docs/tools.rst
```diff
@@ -148,6 +148,18 @@ Rough outline of the automated conversion process
...
+
+    - **Visual** geoms: ``class="visual"`` with ``contype="0" conaffinity="0" group="2"
+      density="0"`` (render-only, never collide) and keep their ``rgba``.
+    - **Collision** geoms: ``class="collision"`` with ``group="3" contype="1" conaffinity="1"``.
+      Whole-mesh/primitive collisions are tinted ``material="bright_orange"`` so they are easy to
+      spot in the viewer (toggle group 3); the material is added to the ``<asset>`` automatically
+      when not already defined (an existing definition, e.g. from ``mujoco_inputs``, is left
+      untouched). **Decomposed** collision pieces are the exception: they get the same group 3 /
+      contype / conaffinity attributes but keep obj2mjcf's own per-decomposition materials/colors
+      instead of being recolored, so the individual convex hulls stay distinguishable.
```
scastro-nasa: should the decomposed meshes bit be placed in a 3rd bullet point?

mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
```diff
@@ -44,6 +44,20 @@
 # Matches obj2mjcf's coacd default.
 DEFAULT_DECOMPOSE_THRESHOLD = "0.05"
 
+# Explicit attributes written onto classified geoms so visual/collision separation does not
+# depend on the user supplying a <default class="..."> block via mujoco_inputs.
+# Visuals are render-only (contype=0 -> never collide) and live in group 2; collisions do the
+# colliding (contype/conaffinity=1), live in group 3, and are tinted COLLISION_MATERIAL_NAME
+# so they can be inspected in the viewer.
```
scastro-nasa: can probably cut down on some of the specifics here, just explain what they are but let the actual values speak for themselves in case it changes?

mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
```diff
@@ -626,8 +640,17 @@ def replace_geoms(mesh_name, sub_geoms, collision_class=None):
...
+                    if collision_class:
+                        sub_geom_local.setAttribute("class", collision_class)
+                    # Give decomposed pieces the same collision-separation attributes as plain
+                    # collisions (group 3, contype/conaffinity 1), but NOT the bright_orange
+                    # material: decomposed meshes keep obj2mjcf's own materials/rgba so the
+                    # individual convex pieces stay distinguishable.
+                    for attribute, value in COLLISION_GEOM_ATTRS.items():
+                        if attribute == "material":
+                            continue
+                        sub_geom_local.setAttribute(attribute, value)
```
scastro-nasa: i can't easily tell why this block of code corresponds to a decomposed piece?

scastro-nasa (changes requested):
Thanks for cleaning this up!

I see a bit of a logic mishap. When I do not set `--use_collision_tags`, it appears only some of the primitives (not meshes) that are visual-used-for-collision are bright orange. but others have remained gray:

Image

When I say `--use_collision_tags`, more of the collision bodies are primitives so you see more orange. But meshes still are gray.

Image

---

Also, this seems to have undone my change where I actually had regular collision meshes in group 3 and decomposed meshes in group 4 for better toggleability. Would you be okay if we kept this behavior, or is there motivation for putting them in the same layer?

BTW I am happy to fix this behavior with our model, but more asking for your opinion on expected behavior before I  maybe push some commits onto here.

sea-bass:
ok I fixed it in the latest commit! (oops this isn't my work account)

image

image

saikishor (commented):
Tested locally and it works for my usecase. @scastro-nasa I need an approval to proceed with this

scastro-nasa (approved):

scastro-nasa:
@ndunkelb-nasa @eholum-nasa because I was the last pusher I can't approve this PR. Could I get a hand? Thanks!

scastro-nasa:
Also @saikishor I assume this means you're okay with plain collision meshes (the orange ones now) and decomposed ones being in separate groups 3 and 4?

saikishor:
> Also @saikishor I assume this means you're okay with plain collision meshes (the orange ones now) and decomposed ones being in separate groups 3 and 4?

For now, it is okay. It is just fine detailing. We can refine it later if needed. I'll go ahead and merge this for now. Should I wait for their approval just in case?

scastro-nasa:
I think you can go ahead if you possess the power!

PR #268, opened 2026-08-06
sibling links reuse already-decomposed collision meshes instead of redecomposing

saikishor:
## Description

  Fourth in the series splitting up #209 (see that PR for full context).

  Fixes decomposition for mirrored/sibling links (e.g. left/right feet). MuJoCo emits a scaled sibling mesh for mirrored geometry (same source mesh, `scale="1 -1 1"`) rather than a separate mesh file.
  Previously, only the original mesh got decomposed, leaving its mirrored sibling with a single whole-mesh collision. This PR decomposes the sibling too, generating scaled copies of the convex pieces, so
  both sides of a mirrored link get proper decomposition.

  Also includes a small pre-commit formatting fix carried over from this point in the original branch history.


scastro-nasa (approved):
LGTM!

PR #269, opened 2026-08-06
only injects default MuJoCo class args when the user hasn't already specified them

saikishor:
## Description

  Last in the series splitting up #209 (see that PR for full context).

  Moves the visual/collision/decomposed-collision physics attributes introduced in earlier PRs of this series off individual geoms and onto MuJoCo `<default class="...">` blocks, so per-geom XML stays
  compact and the attributes are centralised:

  ```xml
  <default>
    <default class="visual">
      <geom contype="0" conaffinity="0" group="2" density="0"/>
    </default>
    <default class="collision">
      <geom group="3" type="mesh" contype="1" conaffinity="1" material="bright_orange"/>
    </default>
    <default class="decomposed_collision">
      <geom group="3" type="mesh" contype="1" conaffinity="1"/>
    </default>
  </default>
  ```

  - These three classes are auto-generated **only if the user hasn't already supplied them** via `mujoco_inputs` — an existing user definition is kept as-is.
  - Visual/collision geoms drop their now-redundant explicit `contype`/`conaffinity`/`group`/`density` attributes and inherit them from the class.
  - Primitive collision geoms still carry their own explicit `type` (e.g. `box`), which overrides the class's `type="mesh"` fallback.
  - Decomposed collision geoms get `type="mesh"` from their class (without it MuJoCo would default to `type="sphere"` and fit a sphere around each convex piece) but no `material` override, so obj2mjcf's
  per-piece colours remain visible.

  Yes, generated MJCF files now rely on default classes for these attributes; users who already define `visual`/`collision`/`decomposed_collision` classes in their own `mujoco_inputs` will have those definitions respected untouched.

### DIFF
--- a/mujoco_ros2_control/docs/tools.rst
+++ b/mujoco_ros2_control/docs/tools.rst
@@ -129,19 +129,25 @@ Rough outline of the automated conversion process
      For instance, if running multiple types of UR robots in one sim, there will be multiple ``shoulder.dae`` files.
 
 - Reads absolute filepaths of all meshes (from both ``<visual>`` and ``<collision>`` tags) and
-  converts either ``.dae`` or ``.stl`` to ``.obj`` using trimesh.
-
-  - Puts all meshes into an ``assets/`` folder under ``mjcf_data/`` relative to current working dir.
-  - Modifies filepaths again in URDF to point to the ``assets/`` folder.
-  - Decomposes large meshes into multiple components to ensure convex hulls.
+  materializes them into an ``assets/`` folder under ``mjcf_data/`` relative to the current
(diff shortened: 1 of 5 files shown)


saikishor:
@scastro-nasa this case is already handled right?

scastro-nasa:
> @scastro-nasa this case is already handled right?

Yes, I think so! That means we can close this and celebrate the end of the stack :pray:

saikishor:
> > @scastro-nasa this case is already handled right?
> 
> Yes, I think so! That means we can close this and celebrate the end of the stack 🙏

Amazing. Thank you

PR #271, opened 2026-08-07
Support MuJoCo intvelocity actuators as velocity interfaces

AbsoluteQ:
## Summary

- recognize MuJoCo `intvelocity` actuators as native velocity-command actuators
- document the `intvelocity` to ros2_control command-interface mapping
- add a regression test that exports a velocity interface and verifies that its command reaches `mjData.ctrl`

## Problem

MuJoCo compiles an `intvelocity` shortcut as an actuator with integrator dynamics, fixed gain, and affine position/velocity feedback. The existing actuator detection only inspects the affine bias parameters, so the nonzero position feedback term causes `intvelocity` to be classified as a position actuator.

That classification rejects a ros2_control velocity command interface even though the actuator's `ctrl` input has velocity semantics: MuJoCo integrates that input internally to produce the position setpoint.

## Change

Detect the compiled `intvelocity` signature before the generic position-actuator pattern and route it through the existing native velocity command path. This changes only the ros2_control interface classification; it does not alter the MJCF actuator or its integrated-velocity dynamics.

The regression test uses a real `<intvelocity>` MJCF actuator, confirms that a velocity command interface is exported, writes a `2.0` velocity command, and verifies that MuJoCo receives `2.0` in the actuator's `ctrl` input.

## Validation

- `pre-commit` passed on all changed files
- ROS 2 Humble build passed
- `colcon test --packages-select mujoco_ros2_control`: 136 tests, 0 errors, 0 failures, 0 skipped

### DIFF
--- a/mujoco_ros2_control/docs/hardware_interface.rst
+++ b/mujoco_ros2_control/docs/hardware_interface.rst
@@ -115,19 +115,27 @@ Maps to the following ``ros2_control`` hardware interface:
    * - Command Interface
      - MuJoCo ``position``
      - MuJoCo ``velocity``
+     - MuJoCo ``intvelocity``
      - MuJoCo ``motor``, ``general``, etc.
    * - **position**
      - Native support
      - Supported using PIDs
      - Supported using PIDs
+     - Supported using PIDs
    * - **velocity**
      - Not supported
+     - Native support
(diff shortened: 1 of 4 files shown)


saikishor (approved):
LGTM

PR #280, opened 2026-08-12
Fix magnetometer unit conversion

### LINKED ISSUE #275: wrong magnetometer units
**Describe the bug**

PR https://github.com/ros-controls/mujoco_ros2_control/pull/257 added support for reading the values of a magnetometer sensor of a MuJoCo model and expsoing the values on `magnetic_field` state interfaces.

The MuJoCo [`sensor/⁠magnetometer`](https://mujoco.readthedocs.io/en/stable/XMLreference.html#sensor-magnetometer) provides the magnetic flux Φ (unit: Weber). However, the state interfaces expect the magnetic field (a.k.a. magnetic flux density) B (unit: Tesla).

The conversion is: $T=\frac{Wb}{m^2}$

Or https://en.wikipedia.org/wiki/Weber_(unit):
> One weber is also the total magnetic flux across a surface of one square meter perpendicular to a magnetic flux density of one tesla

I don't see a way to get the flux density from MuJoCo or infer the area of the flux. The plugin should provide a parameter to either define this area by a users, or to compute this area to scale the flux to a desired magnetic field, e.g. for a selected region in the [world](https://en.wikipedia.org/wiki/Earth%27s_magnetic_field).

**Expected behavior**

The `magnetic_field` sensor interfaces should provide the correct value in Tesla instead of Weber.

**Additional context**
Add any other context about the problem here, especially include any modifications to ros2_control that relate to this issue.

Daniiiil1:
## Summary

- add an optional per-sensor `magnetic_field_scale` parameter
- apply the scale when exposing MuJoCo magnetometer values through ROS `magnetic_field` state interfaces
- reject non-numeric, non-finite, and non-positive scale values
- document tesla-per-model-unit, gauss, and flux/area conversions
- test scaled and unscaled mappings to the same MuJoCo sensor together

## Rationale

MuJoCo does not define physical units for a model's global `magnetic` vector, while ROS magnetic-field interfaces use tesla. Passing the values through unchanged therefore only works when the model already uses tesla. A per-sensor factor lets users express the model's unit convention without changing the MJCF or silently mislabeling the exposed values.

The default remains `1.0`, preserving existing configurations.

## Validation

- `git diff --check`
- `codespell` on all changed files
- formatting checked with `clang-format`
- the Pixi build cannot run locally because the workspace supports Linux only and the current host is macOS ARM64; CI covers the supported Linux targets

Fixes #275

### DIFF
--- a/mujoco_ros2_control/docs/hardware_interface.rst
+++ b/mujoco_ros2_control/docs/hardware_interface.rst
@@ -256,10 +256,18 @@ Map it to the corresponding ``ros2_control`` sensor:
      <param name="mujoco_type">magnetometer</param>
      <!-- mujoco_sensor_name does not need to match the ros2_control sensor name -->
      <param name="mujoco_sensor_name">magnetometer_sensor</param>
+     <!-- Convert MuJoCo model units to tesla; defaults to 1.0 -->
+     <param name="magnetic_field_scale">1.0</param>
      <state_interface name="magnetic_field.x"/>
      <state_interface name="magnetic_field.y"/>
      <state_interface name="magnetic_field.z"/>
    </sensor>
+
+MuJoCo does not define physical units for the model's global ``magnetic`` vector, while the ROS
+``magnetic_field`` interfaces use tesla. Set ``magnetic_field_scale`` to the number of tesla per
+MuJoCo unit. For example, use ``0.0001`` when the model values represent gauss. If the model values
+represent magnetic flux in webers through a known sensor area, use the reciprocal of that area in
+square metres because :math:`1\,\mathrm{T} = 1\,\mathrm{Wb}/\mathrm{m}^2`.
 
 .. warning::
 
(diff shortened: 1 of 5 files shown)


saikishor (approved):
LGTM!

@christianrauch will this work for you?

saikishor:
@christianrauch this works for you?

christianrauch:
Looking at how others using MuJoCo deal with this, it does not look like there is actually a relation between the "magnetic flux" in MuJoCo and the "magnetic field". I haven't found examples where people actually convert between these properties. What I have seen that is commonly done is to set the expected values in the `magnetic` option.

So while using the scalar conversion would provide a mapping, I'm not sure this is a good solution.

It would be useful to have someone with more "MuJoCo magnetometer experience" to weigh in on this and explain how the magnetometer values are supposed to be used.

saikishor:
> Looking at how others using MuJoCo deal with this, it does not look like there is actually a relation between the "magnetic flux" in MuJoCo and the "magnetic field". I haven't found examples where people actually convert between these properties. What I have seen that is commonly done is to set the expected values in the `magnetic` option.
> 
> So while using the scalar conversion would provide a mapping, I'm not sure this is a good solution.
> 
> It would be useful to have someone with more "MuJoCo magnetometer experience" to weigh in on this and explain how the magnetometer values are supposed to be used.

Got it. I'll leave it on standby till then

PR #281, opened 2026-08-13
Fix control mode selection

Ortisa:
## Description

The control mode was previously inferred from the MuJoCo actuator type. However, since the actuator type alone is not enough to determine the control mode, it is now set explicitly when the joints are registered.

Additionally, an error is now thrown directly if a PID-based command interface is defined without the corresponding PIDs. This is done to prevent errors during command mode switching, as previously, interfaces could be claimed when switching to a different control mode even if they were not compatible with it.

## TODOs

- [ ] Fork the repository.

### DIFF
--- a/mujoco_ros2_control/src/mujoco_system_interface.cpp
+++ b/mujoco_ros2_control/src/mujoco_system_interface.cpp
@@ -777,12 +777,25 @@ MujocoSystemInterface::perform_command_mode_switch(const std::vector<std::string
 {
   auto update_joint_interface = [this](const std::string& interface_name, bool enabled) {
     const size_t delimiter_pos = interface_name.rfind('/');
+    if (delimiter_pos == std::string::npos)
+    {
+      RCLCPP_ERROR(get_logger(), "Invalid interface name format: %s", interface_name.c_str());
+      return;
+    }
+
     std::string joint_name = interface_name.substr(0, delimiter_pos);
     std::string interface_type = interface_name.substr(delimiter_pos + 1);
 
     // Find the MuJoCoActuatorData in the vector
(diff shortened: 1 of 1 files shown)


saikishor (approved):
I agree with this change as it is silently failing and letting users to be able to still request interfaces to switch.

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -1300,20 +1300,14 @@ bool MujocoSystemInterface::register_mujoco_actuators()
       return std::isfinite(gains.p_gain_) && std::isfinite(gains.i_gain_) && std::isfinite(gains.d_gain_);
     };
 
-    if (actuator_data.actuator_type == ActuatorType::POSITION)
-    {
-      actuator_data.is_position_control_enabled = true;
```
eholum-nasa, line 1305: Does this get set otherwise? If you do have PID position around velocity, will this still enable it?

eholum-nasa (commented):
This looks right to me! But want to confirm nothing is broken as documented in:

https://control.ros.org/rolling/doc/mujoco_ros2_control/mujoco_ros2_control/docs/hardware_interface.html

mujoco_ros2_control/src/mujoco_system_interface.cpp
saikishor: I'll recheck it locally

saikishor: I verified locally, all changes are properly working and it also fails as expected.

eholum-nasa: Thanks for checking! It's been a while since I've looked at this part of the code so appreciate the help.

eholum-nasa (approved):

PR #282, opened 2026-08-14
Add a pre-step callback to the simulation

eholum-nasa:
This will provide god-like powers to plugins without breaking existing contracts. As it is a new interface.

Basically did what I mentioned in [this comment](https://github.com/ros-controls/mujoco_ros2_control/pull/262#pullrequestreview-4940734863). Added a hook into `MujocoSimulation` and extended the plugin interface with an opportunity to modify `mj_data` before the simulation steps. Plugins can modify data before stepping as needed, and we no longer need special cases for applied forces or velocities, so got to remove that.

Perhaps something to consider as an alternative solution to keep `MujocoSimulation` pure and avoid major changes to the API.

Closes #262

### DIFF
--- a/mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
+++ b/mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
@@ -24,9 +24,21 @@ namespace mujoco_ros2_control_plugins
 /**
  * @brief Base class for MuJoCo ROS 2 control plugins
  *
- * This is an example base class that plugins can inherit from.
- * Plugins can extend the functionality of mujoco_ros2_control
- * by implementing custom behaviors.
+ * Plugins extend mujoco_ros2_control with custom behavior, e.g. publishing extra topics or
+ * applying external forces.
+ *
+ * Two optional hooks are available. Override whichever fits your use case:
+ *
+ * - `update()` runs on the ros2_control control thread, once per `write()` cycle. `data` is a
+ *   recent snapshot, not the live `mj_data_`. Use this for anything that doesn't need to run on
+ *   exactly one physics step, e.g. publishing sensor data or servicing a trigger.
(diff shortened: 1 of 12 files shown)


mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
```diff
@@ -46,22 +58,24 @@ class MuJoCoROS2ControlPluginBase
...
-   * entry to request a hard velocity override on that DOF -- it bypasses the normal mass/contact dynamics for that
-   * DOF this step, overwriting mj_data_->qvel directly right before the next mj_step. Leave an entry NaN to keep its
-   * normal physics-driven value untouched. This is the only channel a plugin has to influence qvel: unlike
-   * ctrl/qfrc_applied/xfrc_applied, data->qvel here cannot be read for the body's actual velocity, since it is not
-   * restored to a real value until after every plugin's update() has run this cycle.
+   * @param data Pointer to a recent snapshot (not the live `mj_data_`)
+   */
+  virtual void update(const mjModel* /*model*/, mjData* /*data*/)
+  {
+  }
```
saikishor, line 67: Should we make this pure virtual? or leave it with default implementation?

saikishor (approved):
I like the approach. LGTM

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
eholum-nasa: The advantage to the default implementation is implementations don't need to explicitly define an empty function if they're not using it? We could alternatively for each plugin to be very deliberate about which it's using, maybe it would help dev be more careful?

saikishor: Sure. I'm fine with it.

Thanks for the clarification

saikishor (approved):

Issue #156, opened 2026-03-28
inconsistent sensor parameters


christianrauch:
**Describe the bug**
The documentation provides the example:
```xml
  <sensor name="imu_sensor">
    <param name="mujoco_type">imu</param>
    <!-- There is no requirement for the mujoco_sensor_name to match the ros2_control sensor name -->
    <param name="mujoco_sensor_name">imu_sensor</param>
    <!-- Default value of orientation_mjcf_suffix is '_quat' -->
    <param name="orientation_mjcf_suffix">_quat</param>
    <!-- Default value of angular_velocity_mjcf_suffix is '_gyro' -->
    <param name="angular_velocity_mjcf_suffix">_gyro</param>
    <!-- Default value of linear_acceleration_mjcf_suffix is '_accel' -->
    <param name="linear_acceleration_mjcf_suffix">_accel</param>
  </sensor>
```

However, `linear_acceleration_mjcf_suffix` (and the other `_mjcf_suffix`) is not a parameter of the sensor, but the `<hardware>` and it is retrieved via `get_hardware_parameter_or`.

IMHO, these plugin-specific parameters should be part of the plugin configuration (i.e. inside `<hardware>` /  `</hardware>`) and not in the interface definition. The interface definition can be reused by multiple hardware implementations, and this way, it clutters the interface and makes reusability more difficult.

**Expected behavior**
Clear separation of interface and implementation.

**Screenshots**

**Environment (please complete the following information):**
 - OS: [e.g. Ubuntu]
 - Version [e.g. Humble]
- Anything that may be unusual about your environment

**Additional context**
Add any other context about the problem here, especially include any modifications to ros2_control that relate to this issue.

siddarth09:
Is this issue still open? Something that I can take up and contribute?

saikishor:
> Is this issue still open? Something that I can take up and contribute?

Hello @siddarth09,

Thank you for the interest. we want to tackle this in the future brainstorming it a bit. For that reason, I would leave this for later

siddarth09:
I just wanted to know if there are any other active issues that I can take up? I am really interested in contributing to this repo

christianrauch:
> I just wanted to know if there are any other active issues that I can take up? I am really interested in contributing to this repo

[...]

PR #289, opened 2026-08-24
Fix 1x1 camera images when the MJCF sets no resolution

### LINKED ISSUE #272: `trackcom` camera registered with 1x1 resolution
**Describe the bug**
With a `trackcom` camera in the MuJoCo model, the images are publisehd with 1x1 resolution.

**To Reproduce**
1. create a model with
    ```XML
    <camera name="track" pos="-1 0 .5" xyaxes="0 -1 0 1 0 2" mode="trackcom"/>
    ```
2. configure the camera plugin with:
    ```YAML    mujoco_ros2_control_node:
      ros__parameters:
        mujoco_plugins:
          mujoco_camera_plugin:
            type: "mujoco_ros2_control_plugins/CameraPlugin"
            track:
              policy: streaming
    ```
3. inspect topic `/mujoco_camera_plugin/track/color`

**Expected behavior**
The image should have the same content to what is rendered when selecting the `track` camera in the GUI.

**Screenshots**
```
$ ros2 topic echo --once /mujoco_camera_plugin/track/color
header:
  stamp:
    sec: 49
    nanosec: 230000000
  frame_id: track_frame
height: 1
width: 1
encoding: rgb8
is_bigendian: 0
step: 3
data:
- 22
- 36
- 51
```

 - OS: Ubuntu 26.04
 - Version [lyrical
- Anything that may be unusual about your environment

**Additional context**
Add any other context about the problem here, especially include any modifications to ros2_control that relate to this issue.

siddarth09:
## Description

 A `<camera>` without a `resolution` attribute gets MuJoCo's `1 1` default, and
  `register_cameras()` fed that straight into the image buffers and the render viewport
  (`camera_plugin.cpp:177-179`), so the camera published 1x1 images instead of failing.

  Two changes:

  - Optional per-camera `width`/`height` plugin parameters, defaulting to the MJCF `resolution`
    values, so the resolution can be set from config without editing the model.
  - A 1x1 resolution is now rejected at registration with an error naming both ways to fix it.

  Worth noting it isn't `trackcom`-specific, which is how the issue was originally reported; a
  plain fixed camera without `resolution` behaves identically. The reason the GUI looks fine is
  that the interactive viewer renders at window size and never reads `cam_resolution`; it only
  matters for offscreen rendering.

  Fixes #272

Yes, and it is breaking for one case: a model with a camera that has no `resolution` attribute
  currently starts and publishes 1x1 images, and will now fail plugin init with an error. That
  seemed like the right trade since 1x1 is never intentional, but I raised it on the issue first
  in case a warn-and-default would be preferred.

  The new `width`/`height` parameters are optional and default to the MJCF value, so existing
  configs are unaffected.

 Yes: Claude (Opus 5). I used it to track down the root cause and draft the implementation,
  which I then typed, built and debugged myself.  

### Additional Information

 Three tests: the rejection path, the params path (publishes at the configured size), and the
  existing tests still passing. The rejection test needs no GL context, since init fails before
  the render thread starts.

  The rejection test asserts `cam_resolution == 1` on load as well as the plugin behaviour, so if
  MuJoCo ever changes that default, the test says why rather than just going red.

  The `width`/`height` parameters overlap with #156, so the naming may want to change depending on
  what is planned there.


saikishor (commented):
To be honest, I'm just reviewing the issue and I personally thing that we shouldn't add yet another parameter to define things. The resolution defined in MuJoCo is enough and I think the user should use it appropriately to set the proper resolution and that's it. If we think we want to enforce it somehow, then make that parameter mandatory etc, but I don't think this approach is the right way to go.

@christianrauch FYI

christianrauch:
I also think that this screen rendering does not need another resolution parameter. How is this even supposed to work? The rendering resolution is already defined by the window. If a user defines another resolution via the parameter, how is this going to work with the window resolution?

The issue is about setting the resolution appropriately instead of the default 1x1. The solution must somehow determine the correct resolution from the rendering.

saikishor:
> The issue is about setting the resolution appropriately instead of the default 1x1. The solution must somehow determine the correct resolution from the rendering.

Yup, may be the user sets a resolution and the requested is less than what is available a warning can be printed to demonstrate the same?

PR #291, opened 2026-08-26
Fix plugin parameter namespaces

scastro-nasa:
## Description

This addresses 2 bugs in the plugin parameter namespaces:

1. Several of the plugins were assuming a hard-coded sub-node namespace that didn't necessarily match with the real plugin name, which is `node_->get_sub_namespace()`
2. Several of the plugins were not applying a parameter namespace. So I found this issue when I realized that for example the free joint state publisher was using fully un-prefixed parameters that clobbered my own plugin!

Yes, in that some incorrect / unexpected behavior got fixed and this may cause slight disruptions to users who had multiple plugins or non-default names.

No, this one was all by hand baby, I still got it!

### DIFF
--- a/mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
+++ b/mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
@@ -31,16 +31,19 @@ bool ExternalWrenchPlugin::init(rclcpp::Node::SharedPtr node, const mjModel* mod
   model_ = model;
 
   // Visualization parameters
-  if (!node_->has_parameter("force_arrow_scale"))
+  const std::string param_prefix = "mujoco_plugins." + node_->get_sub_namespace() + ".";
+  const std::string force_arrow_scale_param = param_prefix + "force_arrow_scale";
+  if (!node_->has_parameter(force_arrow_scale_param))
   {
-    node_->declare_parameter("force_arrow_scale", force_arrow_scale_);
+    node_->declare_parameter(force_arrow_scale_param, force_arrow_scale_);
   }
-  if (!node_->has_parameter("torque_arrow_scale"))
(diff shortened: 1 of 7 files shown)


mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
```diff
@@ -31,16 +31,19 @@ bool ExternalWrenchPlugin::init(rclcpp::Node::SharedPtr node, const mjModel* mod
   model_ = model;
 
   // Visualization parameters
-  if (!node_->has_parameter("force_arrow_scale"))
```
ndunkelb-nasa, line 34: Hmm, did this work correctly before without this fix? Or does not adding a namespace just resolve all params that are like `*.<param_name>`

The external wrench and free joint state publisher plugins seem to be done differently than the others, so wondering if both types of implementations or just one? I can obviously see that assuming the sub-namespace is the same as the plugin name is a problem, but not sure about the other ones.

ndunkelb-nasa (commented):
Looks mainly good, just had a question about the two different flavors of original implementations and if they are all borked, or just one of the flavors.

mujoco_ros2_control_plugins/src/external_wrench_plugin.cpp
scastro-nasa: it worked correctly, just that the parameter was simply named `force_arrow_scale` (no namespace) instead of `mujoco_plugins.<my_plugin_name>.force_arrow_scale`.

So if you loaded 2 of the same type of plugin, or even 2 plugins that both didn't do namespacing but happened to have a common param name, the parameters would 100% clobber each other.

scastro-nasa:
huh... why this makes the tests fail is a mystery. Will fix those momentarily!

ndunkelb-nasa (approved):
LGTM!

PR #294, opened 2026-08-27
Add world_reset plugin hook and restore eq_active on world reset

bgill92:
I am working on a vacuum gripper plugin for mujoco_ros2_control that activates and deactivates weld equality constraints at runtime to simulate a vacuum gripper grabbing things.

Two changes are needed for such runtime-activated constraints to properly reset with the simulation:
- Restore eq_active from eq_active0 (the MJCF-authored defaults) in reset_world_state(), so a runtime-activated constraint returns to its authored state. The existing qpos/qvel/ctrl restore does not touch this field.
- Add a world_reset(mjData*) hook to MuJoCoROS2ControlPluginBase (default no-op, so existing plugins are unaffected) and dispatch it to all plugin instances after the state restore. Plugins cannot detect resets themselves: resets preserve simulation time for ROS clock continuity, so watching data->time go backwards does not work. An explicit notification is the only reliable signal for clearing plugin-owned state (e.g. a latched vacuum) that the reset invalidated.

### DIFF
--- a/mujoco_ros2_control/src/mujoco_simulation.cpp
+++ b/mujoco_ros2_control/src/mujoco_simulation.cpp
@@ -876,9 +876,7 @@ void MujocoSimulation::reset_world_state(bool fill_initial_state,
   std::fill(mj_data_->qfrc_applied, mj_data_->qfrc_applied + mj_model_->nv, 0.0);
   std::fill(mj_data_->xfrc_applied, mj_data_->xfrc_applied + 6 * mj_model_->nbody, 0.0);
 
-  // Restore equality-constraint activations to their MJCF defaults: plugins may activate
-  // constraints at runtime, and a reset must return them to the authored state (this field
-  // is not touched by the qpos/qvel/ctrl restore above).
+  // Restore equality-constraint activations to their MJCF defaults
   std::copy(mj_model_->eq_active0, mj_model_->eq_active0 + mj_model_->neq, mj_data_->eq_active);
 
   {
--- a/mujoco_ros2_control/src/mujoco_simulation.cpp
+++ b/mujoco_ros2_control/src/mujoco_simulation.cpp
@@ -876,6 +876,11 @@ void MujocoSimulation::reset_world_state(bool fill_initial_state,
   std::fill(mj_data_->qfrc_applied, mj_data_->qfrc_applied + mj_model_->nv, 0.0);
   std::fill(mj_data_->xfrc_applied, mj_data_->xfrc_applied + 6 * mj_model_->nbody, 0.0);
 
+  // Restore equality-constraint activations to their MJCF defaults: plugins may activate
+  // constraints at runtime, and a reset must return them to the authored state (this field
+  // is not touched by the qpos/qvel/ctrl restore above).
+  std::copy(mj_model_->eq_active0, mj_model_->eq_active0 + mj_model_->neq, mj_data_->eq_active);
+
   {
     // Clear staged control inputs so stale commands from before the reset are not re-applied
     // on the next step.
--- a/mujoco_ros2_control/src/mujoco_system_interface.cpp
+++ b/mujoco_ros2_control/src/mujoco_system_interface.cpp
@@ -2213,7 +2213,7 @@ void MujocoSystemInterface::reset_simulation_state(bool /*fill_initial_state*/)
   // eq_active has already been restored to MJCF defaults.
   for (auto& plugin : plugin_instances_)
   {
-    plugin->world_reset(simulation_->data());
+    plugin->on_reset_world(simulation_->data());
   }
 }
 
--- a/mujoco_ros2_control/src/mujoco_system_interface.cpp
+++ b/mujoco_ros2_control/src/mujoco_system_interface.cpp
@@ -2213,7 +2213,7 @@ void MujocoSystemInterface::reset_simulation_state(bool /*fill_initial_state*/)
   // eq_active has already been restored to MJCF defaults.
   for (auto& plugin : plugin_instances_)
   {
-    plugin->on_reset_world(simulation_->data());
+    plugin->on_reset(simulation_->data());
   }
 }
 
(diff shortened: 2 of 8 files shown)


mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
```diff
@@ -37,8 +37,13 @@ namespace mujoco_ros2_control_plugins
  *   e.g. a hard kinematic override. There's no command buffer: an untouched entry keeps its
  *   last value, so releasing a command (e.g. an expired force) needs an explicit write. This
  *   should be used with caution!
+ * - `world_reset()` runs on the control thread immediately after a world reset (ResetWorld
+ *   service or a detected UI reset), with the live `mj_data_` in its post-reset state. Use
+ *   this to clear plugin bookkeeping that refers to the pre-reset world (latched states,
+ *   cached ids that are still valid, ...). Note that the reset preserves simulation time
+ *   (ROS clock continuity), so detect resets via this hook, not by watching `data->time`.
```
scastro-nasa: can you make this a bit less verbose?

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
```diff
@@ -77,6 +82,14 @@ class MuJoCoROS2ControlPluginBase
   {
   }
 
+  /**
+   * @brief Called after a world reset, on the control thread. Default does nothing.
+   * @param data Pointer to the live MuJoCo data in its post-reset state.
+   */
+  virtual void world_reset(mjData* /*data*/)
```
scastro-nasa: if this is called automatically by the system interface, I might recommend calling it `on_world_reset()` or something?

mujoco_ros2_control_tests/test/test_world_reset_dispatch.cpp
```diff
@@ -0,0 +1,190 @@
+// Copyright 2026 PAL Robotics S.L.
```
scastro-nasa: technically this can be copyright you / your work

scastro-nasa (commented):
Great to see you on here @bgill92!

Generally looks good, but will defer to the pros to look at this.

mujoco_ros2_control/src/mujoco_simulation.cpp
```diff
@@ -876,6 +876,11 @@ void MujocoSimulation::reset_world_state(bool fill_initial_state,
   std::fill(mj_data_->qfrc_applied, mj_data_->qfrc_applied + mj_model_->nv, 0.0);
   std::fill(mj_data_->xfrc_applied, mj_data_->xfrc_applied + 6 * mj_model_->nbody, 0.0);
 
+  // Restore equality-constraint activations to their MJCF defaults: plugins may activate
+  // constraints at runtime, and a reset must return them to the authored state (this field
+  // is not touched by the qpos/qvel/ctrl restore above).
```
```suggestion
```

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
saikishor: Why not generic one like reset or on_reset?, do we need to specificy it to be world reset?

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
saikishor: agreed

mujoco_ros2_control/src/mujoco_system_interface.cpp
```diff
@@ -2208,6 +2208,13 @@ void MujocoSystemInterface::reset_simulation_state(bool /*fill_initial_state*/)
     joint.velocity_interface.command_ = 0.0;
     joint.effort_interface.command_ = 0.0;
   }
+
+  // Notify plugins: they own runtime state that the world reset has invalidated.
+  // eq_active has already been restored to MJCF defaults.
+  for (auto& plugin : plugin_instances_)
+  {
+    plugin->world_reset(simulation_->data());
+  }
```
saikishor, line 2217: I need to check the code once again to make sure, that there is no race condition between the update method and this new method

saikishor (dismissed):
LGTM

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
bgill92: Done.

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
bgill92: The service to reset the world is called `reset_world`, so I'm going to go ahead and rename the function to `on_reset_world`. Let me know if you want to change it.

mujoco_ros2_control_tests/test/test_world_reset_dispatch.cpp
bgill92: Done.

mujoco_ros2_control_plugins/include/mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp
saikishor: But maybe it has nothing to do with the reset of the world. I think a generic reset is better as you are trying to reset the settings of the plugins at this point. This can be called on world reset or only reset of free bodies or also the reset of the joints right?

bgill92: That's a fair point. I've renamed it to `on_reset()`, since you're trying to just reset the state of the plugin.

saikishor (dismissed):
LGTM

eholum-nasa (approved):
Thanks @bgill92! Updated the rst plugin docs, which could be tied to the c++ header but...

Could you just take a quick read and give me a 👍 if it looks ok to you? Then can ship this.

bgill92:
@eholum-nasa The doc LGTM!

eholum-nasa (approved):

saikishor (approved):

PR #295, opened 2026-08-28
Fix collision geometries usage and add command-line option to toggle it

scastro-nasa:
## Description

Makes some fixes to add to https://github.com/ros-controls/mujoco_ros2_control/pull/264 based on testing on NASA's CLR setup. Specifically:

1. Fixed some bugs that caused the startup to fail for our model
2. Added an `argparse` argument `--use_collision_tags` so you can toggle the legacy and new behaviors (we can keep building on this).

So with that toggle I can get the old behavior, plus the new one!

[mjros2ctrl-collision-meshes.webm](https://github.com/user-attachments/assets/d6ba7562-5211-43a9-9eea-323cf85e003d)

There are still issues with the decomposed meshes sharing layers with other meshes, so they're not easy to isolate. But is a different grouping for these handled upstack?

Yes

Yes, Claude Fable 5

### DIFF
--- a/mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
+++ b/mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
@@ -523,14 +523,6 @@ def update_obj_assets(dom, output_filepath, mesh_info_dict):
                     parent = geom_element.parentNode
                     parent.removeChild(geom_element)
                     for sub_geom in sub_geoms:
-                        # obj2mjcf's file classifies its own geoms: the render geoms (whole
-                        # mesh, with materials) are class "visual" and the collidable ones
-                        # (the whole mesh for composed, the convex pieces for decomposed) are
-                        # class "collision". Match them to the role of the URDF geom being
-                        # replaced - cloning everything into both would render the collision
-                        # pieces and, worse, collide the whole concave mesh as a convex hull.
-                        if (sub_geom.getAttribute("class") == "visual") != is_visual:
-                            continue
                         sub_geom_local = sub_geom.cloneNode(False)
                         if pos:
                             sub_geom_local.setAttribute("pos", pos)
--- a/mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
+++ b/mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
@@ -582,19 +582,18 @@ def update_non_obj_assets(dom, output_filepath):
         if geom.hasAttribute("class"):
             continue
 
+        # Ensure a type: the saved model omits type="sphere" (MuJoCo's default).
+        if not geom.hasAttribute("type"):
+            geom.setAttribute("type", "sphere")
+
         if geom.hasAttribute("contype"):
             # visual geom: keep rgba, strip the raw import attributes.
-            # Ensure a type: the saved model omits type="sphere" (MuJoCo's default).
-            if not geom.hasAttribute("type"):
-                geom.setAttribute("type", "sphere")
             geom.setAttribute("class", "visual")
             for attribute in remove_attributes:
                 if geom.hasAttribute(attribute):
                     geom.removeAttribute(attribute)
         else:
-            # collision geom: ensure a type, drop rgba (not rendered)
-            if not geom.hasAttribute("type"):
-                geom.setAttribute("type", "sphere")
+            # collision geom: drop rgba (not rendered)
             geom.setAttribute("class", "collision")
             if geom.hasAttribute("rgba"):
                 geom.removeAttribute("rgba")
--- a/mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
+++ b/mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
@@ -450,7 +450,11 @@ def update_obj_assets(dom, output_filepath, mesh_info_dict):
     for mesh in meshes:
         mesh_name = mesh.getAttribute("name")
 
-        # This should definitely be there, otherwise something is horribly wrong
+        # MuJoCo emits an auto-renamed sibling asset (e.g., "my_mesh1") when the same mesh
+        # file is referenced with a different scale. Leave these referencing the whole mesh.
+        if mesh_name not in mesh_info_dict:
+            continue
+
         scale = mesh_info_dict[mesh_name]["scale"]
 
         mesh_path = ""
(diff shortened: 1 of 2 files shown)


mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
```diff
@@ -577,7 +583,10 @@ def update_non_obj_assets(dom, output_filepath):
             continue
 
         if geom.hasAttribute("contype"):
-            # visual geom: keep rgba, strip the raw import attributes
+            # visual geom: keep rgba, strip the raw import attributes.
+            # Ensure a type: the saved model omits type="sphere" (MuJoCo's default).
+            if not geom.hasAttribute("type"):
+                geom.setAttribute("type", "sphere")
```
ndunkelb-nasa: Is this necessary? Or just nice to have?

mujoco_ros2_control/scripts/make_mjcf_from_robot_description.py
```diff
@@ -266,22 +266,27 @@ def run_obj2mjcf(output_filepath, decompose_dict, mesh_info_dict):
...
-                threshold,
-            ]
-            subprocess.run(cmd)
-
-            thresholds_data[mesh_name] = float(threshold)
+    # run obj2mjcf to generate folders of processed objs with decompose option for decomposed
+    # components. Make sure a collision mesh sharing its stem with a visual mesh is decomposed too.
+    for mesh_name, mesh_item in mesh_info_dict.items():
+        filename_no_ext = os.path.splitext(os.path.basename(mesh_item["filename"]))[0]
+        if mesh_name not in decompose_dict and filename_no_ext not in decompose_dict:
```
ndunkelb-nasa, line 273: Trying to follow here, why are we checking for both mesh name and filename_no_ext? Do those not end up being the same? Or is this from the repeated mesh issue before? I think this might mess up scenarios like having multiple types of URs in the same model, where they are both named base.dae, but are different meshes?

mujoco_ros2_control/scripts/make_mjcf_from_robot_description.py
```diff
@@ -484,10 +496,17 @@ def main(args=None):
...
 
-    # Keep authored collision geometry so it drives the MuJoCo collision geoms, and
-    # only synthesize a collision from the visual for links that don't define one.
-    # This way visual meshes render the robot while the (often simpler) collision
-    # geometry is used for physics, with the visual mesh as the fallback.
+    # Unless --use_collision_tags is given, drop the URDF's authored collision geometry
+    # first so every link falls back to a collision synthesized from its visuals (the
+    # legacy behavior, and the default).
+    if not parsed_args.use_collision_tags:
+        xml_data = mrc.remove_tag(xml_data, "collision")
```
ndunkelb-nasa, line 503: Ah that is a clever way to do this 0.o

mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
scastro-nasa: I do get errors without it, something about `mesh geom '' (id=471) must have valid meshid`

mujoco_ros2_control/scripts/make_mjcf_from_robot_description.py
scastro-nasa: Exactly, it's to avoid decomposing duplicates of the same mesh that is potentially copied across visual/collision. But there is also a mangle added in the original PR, like `__1`, `__2`, etc. that should avoid the issue you are talking about?

mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
ndunkelb-nasa: Hmm, super weird, this sounds like the issue we saw with the axes for CLR where Mujoco 3.12's native urdf converter prefers to not put in  a value if it is the same as the default.

However, I am a bit confused by this, because it certainly seems like it should default to sphere...
https://mujoco.readthedocs.io/en/stable/XMLreference.html#body-geom

mujoco_ros2_control/scripts/make_mjcf_from_robot_description.py
ndunkelb-nasa: Ah right, I forgot in Sai's new version, that gets added pre this step, right? I think previously it might have been done later down the pipeline (or maybe I am just misremembering)

mujoco_ros2_control/scripts/make_mjcf_from_robot_description.py
scastro-nasa: it also means we can similarly do clever substitutions in this eventual future of deciding what to do on a per-body level, like "if link has mj_collision tag, replace its collision tag with this"

ndunkelb-nasa: Yeah, I think the only thing really stopping us from doing this is if other URDF parsers remain happy. I know somewhere in the converter we end up with angry python URDF parsers if we have extra tags at some point

ndunkelb-nasa (approved):

mujoco_ros2_control/scripts/make_mjcf_from_robot_description.py
```diff
@@ -484,10 +496,17 @@ def main(args=None):
...
+
+    # With --use_collision_tags, authored collision geometry is kept so it drives the
+    # MuJoCo collision geoms, and a collision is only synthesized from the visual for
+    # links that don't define one. This way visual meshes render the robot while the
+    # (often simpler) collision geometry is used for physics, with the visual mesh as
+    # the fallback.
```
saikishor:
```suggestion
```

Do we need this?

mujoco_ros2_control/scripts/make_mjcf_from_robot_description.py
saikishor: I like this

saikishor (approved):
LGTM

mujoco_ros2_control/mujoco_ros2_control/urdf_to_mujoco_utils.py
saikishor: Maybe, it is better to mention this in the comments?. I like to leave evidence here and there, gives good reasoning IMO

scastro-nasa: so this was already present in the collision geoms below, just that it comes up again in the visuals now

PR #299, opened 2026-09-02
Do not overwrite a camera snapshot that is still being rendered

MaximilienNaveau:
## Description

`CameraPlugin::update_cameras()` deliberately runs *without* holding `data_mutex_`, because
rendering is slow and must not block the control loop, so it reads `mj_camera_data_` and the
per-camera `render_pending` flags unlocked. Nothing stopped `update()`, on the sim thread,
from calling `mjv_copyData` into that same snapshot mid-render. The result is an image
blended from two different simulation times: corruption that still looks like a perfectly
valid frame, and which becomes more likely the larger the cameras or the faster the publish
rate.

A `render_in_flight_` flag, set and cleared under the lock so it can never be observed as
`false` while `new_data_` is already `true`, makes `update()` skip the slot instead:

* A skipped **streaming** slot is counted and reported through a throttled warning naming
  the things that fix it.
* A **polled** trigger is re-armed rather than consumed, so a service request is deferred to
  the next pass rather than lost.
* The slot is consumed on a skip. Without that, every control cycle for the rest of the
  render would re-enter the guard and the counter would report *attempts* -- thousands per
  second at a 2 kHz control rate -- rather than frames actually lost.

`dropped_frames()` exposes the counter, which is also what makes the behaviour testable.

Fixes # (issue)

Yes, though only in what is reported. Frames that were previously published as silently
corrupted images are now skipped, and a throttled warning appears when that happens:

```text
Camera rendering cannot keep up: 12 streaming frame(s) dropped so far.
```

No new parameters. Polled camera behaviour is unchanged: triggers are deferred, never
dropped.

Yes -- Claude Code (Claude Opus). The synchronisation guard, the dropped-frame accounting,
the two tests and the documentation note were produced with it. The problem was identified
and the result reviewed by hand.

### Additional Information

Both tests use a 1280x720 camera whose render pass cannot keep up with a 1 kHz slot clock.
The second is the interesting one: it drives `update()` roughly 2500 times over 500 ms and
requires the drop count to stay bounded by the number of *slots* elapsed rather than
tracking the number of calls, which is exactly the distinction the slot-consuming logic
exists to make.

Verified on Humble against a Kangaroo biped with six 848x480 cameras at 30 Hz, where the
corruption was originally observed.

## TODOs

- [x] Fork the repository.

### DIFF
--- a/mujoco_ros2_control_plugins/test/test_camera_plugin.cpp
+++ b/mujoco_ros2_control_plugins/test/test_camera_plugin.cpp
@@ -361,6 +361,92 @@ TEST_F(CameraPluginTest, PolledCameraPublishesOncePerTrigger)
   plugin.cleanup();
 }
 
+// ---------------------------------------------------------------------------------------
+// Dropped-frame accounting
+//
+// update() runs on the sim thread and must not overwrite a snapshot the rendering thread is
+// still working on. When it has to skip, it consumes the slot and counts one lost frame.
+// A high publish rate against a camera large enough that a pass outlasts the interval puts
+// the guard on the hot path without needing to control the rendering thread directly.
+// ---------------------------------------------------------------------------------------
+
+namespace
+{
+// 1280x720 is comfortably slower to render and read back than a 1 ms publish interval.
+constexpr const char* SLOW_CAMERA_MODEL = R"(<?xml version="1.0"?>
+<mujoco model="slow_camera">
+  <visual>
+    <global offwidth="1280" offheight="720"/>
+  </visual>
+  <worldbody>
+    <light pos="0 0 3"/>
+    <body name="box" pos="0 0 0.1">
+      <freejoint/>
+      <inertial pos="0 0 0" mass="1.0" diaginertia="0.01 0.01 0.01"/>
+      <geom type="box" size="0.05 0.05 0.05"/>
+    </body>
+    <camera name="cam" pos="0 -1 0.2" mode="fixed" resolution="1280 720"/>
+  </worldbody>
+</mujoco>
+)";
+}  // namespace
+
+TEST_F(CameraPluginTest, NoFramesAreDroppedWhenTheRendererKeepsUp)
+{
+  load_model(SLOW_CAMERA_MODEL);
+  // One frame every 100 s, so no slot ever comes due after the first.
+  plugin_node_->declare_parameter("mujoco_plugins.camera_plugin.camera_publish_rate", 0.01);
+
+  mujoco_ros2_control_plugins::CameraPlugin plugin;
+  ASSERT_TRUE(plugin.init(plugin_node_, model_, data_));
+  wait_for_rendering(plugin);
+
+  for (int i = 0; i < 2000; ++i)
+  {
+    plugin.update(model_, data_);
+  }
+  EXPECT_EQ(plugin.dropped_frames(), 0u);
+  plugin.cleanup();
+}
+
+// The regression this guards: without the in-flight check, update() would copy into the
+// snapshot the rendering thread is reading, producing a frame blended from two sim times --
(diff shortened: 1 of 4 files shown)


mujoco_ros2_control_plugins/test/test_camera_plugin.cpp
```diff
@@ -361,6 +361,92 @@ TEST_F(CameraPluginTest, PolledCameraPublishesOncePerTrigger)
...
+
+  mujoco_ros2_control_plugins::CameraPlugin plugin;
+  ASSERT_TRUE(plugin.init(plugin_node_, model_, data_));
+  wait_for_rendering(plugin);
+
+  // Drive update() the way a 2 kHz control loop would, for long enough that several render
+  // passes are outrun.
+  std::size_t calls = 0;
+  const auto deadline = std::chrono::steady_clock::now() + 500ms;
+  while (std::chrono::steady_clock::now() < deadline)
```
eholum-nasa, line 430: Kind of a similar note to the other tests, just thinking about how to simplify here. Could these tests be easier if we added say, a FRIEND_CLASS function to the camera header that directly supported setting `render_in_flight_`? Then instead of having to contrive success / failure scenarios we could basically have one test that interacted with update directly.

mujoco_ros2_control_plugins/test/test_camera_plugin.cpp
```diff
@@ -361,6 +361,92 @@ TEST_F(CameraPluginTest, PolledCameraPublishesOncePerTrigger)
...
+  // One frame every 100 s, so no slot ever comes due after the first.
+  plugin_node_->declare_parameter("mujoco_plugins.camera_plugin.camera_publish_rate", 0.01);
+
+  mujoco_ros2_control_plugins::CameraPlugin plugin;
+  ASSERT_TRUE(plugin.init(plugin_node_, model_, data_));
+  wait_for_rendering(plugin);
+
+  for (int i = 0; i < 2000; ++i)
+  {
+    plugin.update(model_, data_);
```
eholum-nasa, line 406: I'm not sure that this will actually exercise the "keep up" logic here? We'll get one tick and then we _probably_ won't get to the next trigger time at 100s, unless something is horribly wrong, so dropped frames will _always_ be 0.

eholum-nasa (commented):
Thanks @MaximilienNaveau! Not a race that we have run into since we use polled or slow rendering cameras!

The fix itself _looks_ good to me! I do have questions about the unit tests, especially since we've had stability issues with tests before.

Tangentially, I saw the other PRs (#300, #302, #303, and #304, specifically) all kind of touch the camera plugin as well. Was there an ordering that you had in mind for those? For instance #300 references changes from #302 w.r.t. to the rendering backend, so I feel like it should come first?

MaximilienNaveau:
> Tangentially, I saw the other PRs (#300, #302, #303, and #304, specifically) all kind of touch the camera plugin as well. Was there an ordering that you had in mind for those? For instance #300 references changes from #302 w.r.t. to the rendering backend, so I feel like it should come first?

Hi,

thank you for the review!
For the order of the PR it should be:
- fixes first: #299 #300 #301 .
- then the #302
- and finally the rest.

I will take a second look at the unit-tests. Thank for pointing that out.

mujoco_ros2_control_plugins/test/test_camera_plugin.cpp
MaximilienNaveau: This is the point of this unit-tests. We should never have a drop, we should always be able to generate 2000 images in a 100s or less.

I agree that having another test actually firing the "keep up" message would be good as well.

Let me know if we should keep this test as validation of: "if we keep up nothing happens ever".

PR #296, opened 2026-08-29
Migrate to the new Command/StateInterface API

saikishor:
## Summary

This PR addresses the changes to make sure the mujoco_ros2_control works with the new API of the Command and State Interfaces as the old API will be removed on rolling soon. 

# AI usages

Claude Code Opus 5 is used

PR #306, opened 2026-09-09
Migrate to the modern Handle export API on hardware_interface >= 5.6

destogl:
## Summary

ros2_control is removing the deprecated raw-pointer `Handle` constructor and the by-value `export_state_interfaces()`/`export_command_interfaces()` overrides (see ros-controls/ros2_control#3610). `mujoco_ros2_control` still uses both throughout its joint/FT-sensor/IMU/pose/magnetometer registration, which breaks once that change lands.

- Interface enumeration logic (which joint/sensor interfaces exist) is unchanged in substance, factored into shared helpers (`enumerate_state_interfaces()`/`enumerate_command_interfaces()`) so it isn't duplicated between the old and new export paths.
- The existing internal `state_`/`command_` doubles stay the source of truth; `read()`/`write()` gained a small `set_value()`/`get_value()` sync step against the exported handles, replacing pointer aliasing.
- Guarded behind `HARDWARE_INTERFACE_VERSION_GTE(5, 6, 0)` — the version that introduced `on_export_state_interfaces()` — so humble/jazzy/kilted keep using the existing path unchanged.
- Also fixes two test files (`test_headless_init.cpp`, `test_mujoco_system_interface.cpp`) that called the deprecated export methods directly; on `hardware_interface` >= 5.6 the base class no longer routes those to the overridden `on_export_*()` methods, so exported interfaces came back empty without this fix.

## Test plan

Built and tested against both the current `ros2_control` `master` and the `remove/deprecated_methods` branch:

- `colcon build --packages-select mujoco_ros2_control` — clean on both.
- `colcon test --packages-select mujoco_ros2_control` — same result on both: all C++ gtests pass; the one remaining failure (`test_urdf_to_mujoco_utils`, a Python test) is a pre-existing, unrelated `ModuleNotFoundError: No module named 'urdf_parser_py'` dependency gap, reproducible on unmodified `main` too.

### DIFF
--- a/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_system_interface.hpp
+++ b/mujoco_ros2_control/include/mujoco_ros2_control/mujoco_system_interface.hpp
@@ -19,9 +19,11 @@
 
 #pragma once
 
+#include <functional>
 #include <memory>
 #include <string>
 #include <thread>
+#include <utility>
 #include <vector>
 
 #include <hardware_interface/version.h>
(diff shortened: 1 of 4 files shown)


saikishor:
We have https://github.com/ros-controls/mujoco_ros2_control/pull/296 that does the same

destogl:
@saikishor then closing this.

PR #307, opened 2026-09-11
Add support for simulating sensor noise

saikishor:
## Description

This PR adds support for sensor noise so more realistic sensors can be simulated in the overall framework

NO

Yes, Claude

### DIFF
--- a/mujoco_ros2_control/include/mujoco_ros2_control/utils.hpp
+++ b/mujoco_ros2_control/include/mujoco_ros2_control/utils.hpp
@@ -22,11 +22,14 @@
 #include <Eigen/Core>
 #include <Eigen/Geometry>
 #include <hardware_interface/hardware_info.hpp>
+#include <rclcpp/rclcpp.hpp>
 
 #include "mujoco_ros2_control/data.hpp"
 
+#include <cassert>
 #include <cmath>
 #include <random>
+#include <string>
 
 namespace mujoco_ros2_control
 {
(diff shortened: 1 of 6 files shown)


mujoco_ros2_control/include/mujoco_ros2_control/utils.hpp
```diff
@@ -41,4 +48,118 @@ get_sensor_from_info(const hardware_interface::HardwareInfo& hardware_info, cons
...
+  return default_value;
+}
+
+/**
+ * @brief Reads the sensor's `noise_distribution` parameter ("gaussian" (default) or "uniform").
+ * Anything other than exactly "uniform" (including an absent parameter) is treated as Gaussian.
+ */
+inline NoiseDistribution get_noise_distribution(const hardware_interface::ComponentInfo& sensor)
+{
+  return get_sensor_parameter_or(sensor, "noise_distribution", "gaussian") == "uniform" ? NoiseDistribution::kUniform :
```
ndunkelb-nasa: It would be nice to have a warning message here if there is a typo. Something like "Only noise distrubutions of 'guassian' or 'uniform' are allowed, but you selected 'unifrm'. Defaulting to 'guassian'."

mujoco_ros2_control/include/mujoco_ros2_control/utils.hpp
```diff
@@ -41,4 +48,118 @@ get_sensor_from_info(const hardware_interface::HardwareInfo& hardware_info, cons
...
+                                                                                          NoiseDistribution::kGaussian;
+}
+
+/**
+ * @brief Sets a flat, row-major 3x3 covariance's diagonal to `stddev * stddev`, leaving off-diagonal terms
+ * untouched (0 for a freshly-`resize()`d vector, matching the "independent per-axis noise" assumption).
+ */
+inline void set_diagonal_covariance(std::vector<double>& covariance, double stddev)
+{
+  covariance[0] = covariance[4] = covariance[8] = stddev * stddev;
```
ndunkelb-nasa, line 93: May be worth adding a check here that the incoming vector is size 9? Or changing the name of the function to something like set_3_by_3_covariance() or something. Just so that someone doesn't assume this should just work for other sizes as the name kind of suggests

ndunkelb-nasa (commented):
Wow, I can't believe mujoco actually removed the support to do this 0.o But glad to have it back available!

Just a couple of minor comments, but otherwise, looks pretty good!

mujoco_ros2_control/include/mujoco_ros2_control/utils.hpp
saikishor: Sure thing

mujoco_ros2_control/include/mujoco_ros2_control/utils.hpp
saikishor: Got it, i can add a check for it

saikishor:
@ndunkelb-nasa done

ndunkelb-nasa (approved):
LGTM!

PR #308, opened 2026-09-11
Add FTS Gravity Compensation Plugin

ndunkelb-nasa:
## Description

Performs gravity compensation for a force torque sensor based on provided parameters.

This is intended to simulate hardware configurations that have a way to perform gravity compensation on the force torque sensor data based on the weight and center of gravity of your end effector.
One example of this is with the [Universal Robot Set Payload](https://www.universal-robots.com/manuals/EN/HTML/SW5_22/Content/prod-usr-man/software/PolyScope/content/BasicProgNodes/commandtab_set_payload_en.htm) feature.

See [plugin documentation](https://github.com/NASA-JSC-Robotics/mujoco_ros2_control/blob/ndunkelb/add-fts-grav-comp/mujoco_ros2_control_plugins/doc/plugins.rst#ftsgravcompplugin) for more details!

No

Yes, Claude


mujoco_ros2_control_plugins/doc/plugins.rst
```diff
@@ -493,6 +493,76 @@ FreeJointStatePublisher Parameters
 
    ros2 topic echo /mujoco_ros2_control_node/free_joint_state_publisher/free_joint_states
 
+FtsGravCompPlugin
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Performs gravity compensation for a force torque sensor based on provided parameters.
+
+This is intended to mock hardware configurations that have a way to perform gravity compensation on the force torque sensor data based on the weight and center of gravity of your end effector.
+One example of this is with the `Universal Robot Set Payload <https://www.universal-robots.com/manuals/EN/HTML/SW5_22/Content/prod-usr-man/software/PolyScope/content/BasicProgNodes/commandtab_set_payload_en.htm>`_ command.
```
scastro-nasa: giga nitpick

```suggestion
One example of this is with the `Universal Robots Set Payload <https://www.universal-robots.com/manuals/EN/HTML/SW5_22/Content/prod-usr-man/software/PolyScope/content/BasicProgNodes/commandtab_set_payload_en.htm>`_ command.

File: mujoco_ros2_control_plugins/test/test_fts_grav_comp_plugin.cpp [hunk 2]
```diff
@@ -0,0 +1,367 @@
+// Copyright 2026 PAL Robotics S.L.
```
scastro-nasa: NASA header?

mujoco_ros2_control_plugins/doc/plugins.rst
```diff
@@ -493,6 +493,76 @@ FreeJointStatePublisher Parameters
...
+   :widths: 15 15 70
+   :header-rows: 1
+
+   * - Parameter
+     - Type
+     - Description
+   * - ``frame_id``
+     - ``string``
+     - Name of the mujoco site that the center of mass is represented in.
+       This should probably correspond to a link in the URDF.
```
scastro-nasa: I guess if you don't have a URDF this point is moot, so is it easiest to just remove this sentence?

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.cpp
```diff
@@ -0,0 +1,269 @@
...
+    // everything must be negated to handle the way force and torque sensors are set up
+    fts_[0].wrench.force.x = -sensordata_force[0];
+    fts_[0].wrench.force.y = -sensordata_force[1];
+    fts_[0].wrench.force.z = -sensordata_force[2];
+    fts_[0].wrench.torque.x = -sensordata_torque[0];
+    fts_[0].wrench.torque.y = -sensordata_torque[1];
+    fts_[0].wrench.torque.z = -sensordata_torque[2];
+
+    fts_wrench_publisher_->publish(fts_[0].wrench);
+  }
```
scastro-nasa: marking TODO for removal

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.cpp
```diff
@@ -0,0 +1,269 @@
...
+    fts_[0].wrench.force.y = -sensordata_force[1];
+    fts_[0].wrench.force.z = -sensordata_force[2];
+    fts_[0].wrench.torque.x = -sensordata_torque[0];
+    fts_[0].wrench.torque.y = -sensordata_torque[1];
+    fts_[0].wrench.torque.z = -sensordata_torque[2];
+
+    fts_comped_wrench_publisher_->publish(fts_[0].wrench);
+
+    last_publish_time_ = current_time;
+  }
```
scastro-nasa: again marking this

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.hpp
```diff
@@ -0,0 +1,106 @@
...
+#include "mujoco_ros2_control_plugins/mujoco_ros2_control_plugins_base.hpp"
+
+namespace mujoco_ros2_control_plugins
+{
+
+/**
+ * @brief Per-FTS bookkeeping -
+ *
+ * One instance exists per camera registered by CameraPlugin::register_cameras(). Buffers and
+ * messages are reused across renders to avoid reallocating on every publish.
```
scastro-nasa: copy-pasta camera comments

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.hpp
```diff
@@ -0,0 +1,106 @@
...
+struct FtsData
+{
+  std::string sensor_name;
+  int sensor_adr_force{ -1 };
+  int sensor_adr_torque{ -1 };
+  int cog_site_id;
+  mjtNum cog_pos[3];
+  mjtNum cog_mass;
+  geometry_msgs::msg::Wrench wrench;
+  int fts_site_id{ -1 };
```
scastro-nasa, line 49: worth documenting these?

mujoco_ros2_control_plugins/test/test_fts_grav_comp_plugin.cpp
```diff
@@ -0,0 +1,367 @@
...
+  {
+    node_ = std::make_shared<rclcpp::Node>("fts_grav_comp_test_node");
+    plugin_node_ = node_->create_sub_node("fts_grav_comp_plugin");
+
+    // Use two executor threads so a blocking service callback (sleeping for
+    // its wrench duration) does not prevent other callbacks (e.g. subscription
+    // delivery) from being dispatched.
+    executor_ = std::make_unique<rclcpp::executors::MultiThreadedExecutor>(rclcpp::ExecutorOptions{}, 2);
+    executor_->add_node(node_);
+    spin_thread_ = std::thread([this]() { executor_->spin(); });
```
scastro-nasa, line 107: multi-threaded executor + separate manual threads seems heavy-handed. If you just put topics and services in separate callback groups, a single-threaded executor should be just fine?

I get wanting to dispatch the spin thread though; this is fine.

scastro-nasa (commented):
Looks good -- just have small comments.

I am generally worried about continuing to add more core plugins to `mujoco_ros2_control` under a single package, since the build times will keep going up and up. I suspect at some point it's going to be a `ros2_controllers` situation where there is a separate monorepo of mujoco plugins, each in their own package, so the binaries distribution (or selective source builds) can save the day.

mujoco_ros2_control_plugins/doc/plugins.rst
ndunkelb-nasa: Yeah, I guess that is fair. When you generate your MJCF, it creates sites out of frame ids, so I was trying to make it more ros-y to understand rather than mujoco-y, but you are right, this doesn't make a ton of sense as is.

mujoco_ros2_control_plugins/test/test_fts_grav_comp_plugin.cpp
ndunkelb-nasa: Fixed!

mujoco_ros2_control_plugins/doc/plugins.rst
ndunkelb-nasa: Corrected

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.cpp
ndunkelb-nasa: Cant believe I accidentally left these in! Took these out, and was able to remove all of the extra infra from this as well.

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.hpp
ndunkelb-nasa: Documented!

mujoco_ros2_control_plugins/test/test_fts_grav_comp_plugin.cpp
ndunkelb-nasa: Yeah, I think I just ended up bringing over the infra from a different test case, and left it as is, but updated now!

ndunkelb-nasa:
> Looks good -- just have small comments.
> 
> I am generally worried about continuing to add more core plugins to `mujoco_ros2_control` under a single package, since the build times will keep going up and up. I suspect at some point it's going to be a `ros2_controllers` situation where there is a separate monorepo of mujoco plugins, each in their own package, so the binaries distribution (or selective source builds) can save the day.

Yeah, that is true. Would definitely make sense to bring these elsewhere, although it would get a little muddled because we have demos/documentation intertwined I think...

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.cpp
```diff
@@ -0,0 +1,220 @@
...
+      size_t dot_pos = relative_path.find('.');
+      if (dot_pos != std::string::npos)
+      {
+        std::string sensor_name = relative_path.substr(0, dot_pos);
+
+        // Skip "type" parameter
+        if (sensor_name != "type")
+        {
+          sensor_names_set.insert(sensor_name);
+        }
```
saikishor, line 57: This logic here is very sensitive, but fine

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.cpp
```diff
@@ -0,0 +1,220 @@
...
+    // grab the rest of the parameters that we need
+    const std::vector<double> cog_pos_param = node_->get_parameter(sensor_param_prefix + "CoG.pos").as_double_array();
+    const double cog_mass_param = node_->get_parameter(sensor_param_prefix + "CoG.mass").as_double();
+    const std::string frame_id_param = node_->get_parameter(sensor_param_prefix + "frame_id").as_string();
+
+    // make sure that the force and torque sensors are of the right dimension
+    const int sensor_dim_force = model->sensor_dim[sensor_id_force];
+    const int sensor_dim_torque = model->sensor_dim[sensor_id_torque];
+
+    if ((sensor_dim_force != 3) || (sensor_dim_torque != 3))
```
saikishor, line 111:
```suggestion
    if (cog_pos_param.size() != 3)
    {
        RCLCPP_ERROR(logger_, "CoG.pos should be of exactly size 3 representing x, y, z coordinates");
    }

    if ((sensor_dim_force != 3) || (sensor_dim_torque != 3))
```

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.cpp
```diff
@@ -0,0 +1,220 @@
...
+    mjtNum* sensordata_torque = data->sensordata + fts.sensor_adr_torque;
+    // pre-processed force and torque
+    mjtNum raw_force_in_fts_frame[3];
+    mjtNum raw_torque_in_fts_frame[3];
+    mju_copy3(raw_force_in_fts_frame, sensordata_force);
+    mju_copy3(raw_torque_in_fts_frame, sensordata_torque);
+    // gravity compensated force and torque
+    mjtNum comped_force_in_fts_frame[3];
+    mjtNum comped_torque_in_fts_frame[3];
+
```
saikishor, line 183: If I understand well, this modifies the data within the ros2_control interfaces right?

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+namespace mujoco_ros2_control_plugins
+{
+
+/**
+ * @brief Per-FTS bookkeeping -
+ *
+ * One instance exists per FTS registered by FtsGravCompPlugin::register_fts().
+ */
+struct FtsData
```
saikishor:
```suggestion
struct FTsData
```

Can we change everywhere frmo Ft to FT?

saikishor (changes requested):
I didn't take a look at the tests clearly, but I trust you ;)

Rest looks good to me, just minor nitpicks :)

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.cpp
ndunkelb-nasa: Indirectly, yes! It is modifying the data for the sensors in mujoco directly, so when ros2 control picks up the info out of there, it is good to go!

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.hpp
ndunkelb-nasa: Yeah, I wasn't positive about capitalization for this and feel like it can be confusing no matter what you do... I modified `FtsData` to be `FTsData`, but do you think that it should be changed everywhere? Like 
`FtsGravCompPlugin` -> `FTsGravCompPlugin`? I didn't want to necessarily do `FTSGravCompPlugin` to represent `ForceTorqueSensor` because I feel like all the caps in a row is weird.

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.cpp
ndunkelb-nasa: Yeah, that is fair. I could change it to be something like looking for checking for a required parameter underneath the sensor name. Like seeing if `<sensor_name>.frame_id` exists, which would be more robust if more top-level parameters get added. What do you think about that?

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.cpp
```diff
@@ -0,0 +1,225 @@
...
+    const double cog_mass_param = node_->get_parameter(sensor_param_prefix + "CoG.mass").as_double();
+    const std::string frame_id_param = node_->get_parameter(sensor_param_prefix + "frame_id").as_string();
+
+    // make sure that the force and torque sensors are of the right dimension
+    const int sensor_dim_force = model->sensor_dim[sensor_id_force];
+    const int sensor_dim_torque = model->sensor_dim[sensor_id_torque];
+
+    if (cog_pos_param.size() != 3)
+    {
+      RCLCPP_ERROR(logger_, "CoG.pos should be of exactly size 3 representing x, y, z coordinates");
```
eholum-nasa, line 108: Should this be a failure case?

mujoco_ros2_control_plugins/doc/plugins.rst
```diff
@@ -493,6 +493,75 @@ FreeJointStatePublisher Parameters
...
+           force_torque_sensor_2:
+             # mujoco site the CoG is represented in
+             frame_id: fts_frame_2
+             # specifies the center of gravity w.r.t the 'frame_id' parameter
+             CoG:
+               pos:
+                 - 0.025 # x in m
+                 - 0.3 # y in m
+                 - 0.01 # z in m
+               mass: 5.0 # mass in kg
```
eholum-nasa, line 563: Do we need both sensors in the docs? Would also recommend making the comment match the sensor name where we've got it in tests and examples,

```suggestion
   /**:
     ros__parameters:
       mujoco_plugins:
         fts_grav_comp_plugin:
           type: "mujoco_ros2_control_plugins/FtsGravCompPlugin"
           # name of the sensor to modify. Note that mujoco sensors will look
           # like 'fts_sensor_force' and 'fts_sensor_torque'
           fts_sensor:
             # mujoco site the CoG is represented in
             frame_id: fts_frame_1
             # specifies the center of gravity w.r.t the 'frame_id' parameter
             CoG:
               pos:
                 - 0.0 # x in m
                 - 0.0 # y in m
                 - 0.05 # z in m
               mass: 2.0 # mass in kg
```

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.hpp
```diff
@@ -0,0 +1,105 @@
...
+ */
+struct FTsData
+{
+  // sensor name of FTS. Mujoco should have sensors <sensor_name>_force and <sensor_name>_torque
+  std::string sensor_name;
+  // mujoco sensor address for the force and torque sensors
+  int sensor_adr_force{ -1 };
+  int sensor_adr_torque{ -1 };
+  // site id of the site used for center of gravity reference
+  int cog_site_id;
```
eholum-nasa, line 43: Maybe? Everything else has it!

```suggestion
  int cog_site_id{ -1 };
```

eholum-nasa (commented):
Just curious! Things _look_ good but had some minor questions.

Out of curiosity could it be possible to get the CoM programmatically from the mujoco model rather than having to specify it on the fly? Is that available in the mujoco model? No need to change anything just genuinely curious.

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.cpp
saikishor: Well, true, yes it is better to do it that way, so we can handle it if any new parameters are added on top level. I'll create an issue, we can handle it later

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.hpp
saikishor: FtsGravCompPlugin -> FTsGravCompPlugin

I think it is better to rename it too, because after all, T means Torque

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.cpp
saikishor: Yes, you are right. It should be a failure case

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.hpp
scastro-nasa: `FtsGravCompPlugin` is pretty standard Google style casing, I'm of the opposite opinion where mixing things up looks like a typo.

saikishor: Hahaha let's ask for a tiebreaker from Erik!

saikishor: I would say s smaller, only FT capital like `FTsGravCompPlugin` instead of FTSGravCompPlugin

eholum-nasa: Well... I usually like google's style guide which would say to treat acronyms like works, so `FtsGravCompPlugin`... that said we haven't done a particularly good job of living up to those standards (e.g. https://github.com/ros-controls/mujoco_ros2_control/blob/main/mujoco_ros2_control/include/mujoco_ros2_control/data.hpp#L230).

So if we want to match existing types I would say `FTSensorData` and `FTSensorGravCompPlugin`...  But if we want to follow google `FtsData` and `FtsGravCompPlugin` are fine.

mujoco_ros2_control_plugins/doc/plugins.rst
ndunkelb-nasa: I included 2 just to show that you didn't need a different plugin for each, but dropped it to just one that matches the test params for simplicity!

ndunkelb-nasa:
> Just curious! Things _look_ good but had some minor questions.
> 
> Out of curiosity could it be possible to get the CoM programmatically from the mujoco model rather than having to specify it on the fly? Is that available in the mujoco model? No need to change anything just genuinely curious.

Yeah, I imagine you probably could, but would probably involve some math to get it to resolve to the position you want. But I think that this implementation gives the most flexibility. I think maybe the other option would be to have an optional flag that could resolve just those params on the fly, but if someone wants that, we can maybe do it in a future revision.

mujoco_ros2_control_plugins/src/fts_grav_comp_plugin.hpp
ndunkelb-nasa: Reverted to the previous way based on consensus :+1:

saikishor (approved):

PR #312, opened 2026-09-18
Add check-docs workflow

christophfroehlich:
## Description

Checks the sphinx integration on control.ros.org


eholum-nasa (commented):
This basically hooks the rolling docs into the larger ros-controls documentation build, right? Does this produce docs artifacts that we can download and look at? Or is there a place to see the rendered docs from the job?

The other[ pixi  docs build job](https://github.com/ros-controls/mujoco_ros2_control/blob/main/.github/workflows/docs.yaml) does this in isolation, and maybe that check would be worth dropping if this is the only health check we need!

christophfroehlich:
Exactly, it basically flags anything which would produce a warning [here](https://github.com/ros-controls/control.ros.org/actions/workflows/sphinx-check-warnings.yml).

The reusable workflow does not yet upload artifacts, but could be easily added.

